PR #13, opened 2018-05-16
Config readonly

pookmish:
# READY FOR REVIEW

# Summary
- Replace the config readonly event service so that we can explicitly make module configs read only.

# Needed By (Date)
- asap

# Urgency
- Medium

# Steps to Test

1. Checkout this branch
2. use an existing site/fresh install
3. enable `hs_config_readonly` module
4. attempt to change any configuration provided from a module.
5. validate you cant change content type settings and fields. but you can change display settings

# See Also

### DIFF
--- a/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
+++ b/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
@@ -2,16 +2,116 @@
 
 namespace Drupal\hs_config_readonly\EventSubscriber;
 
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Drupal\config_readonly\ConfigReadonlyWhitelistTrait;
 use Drupal\config_readonly\ReadOnlyFormEvent;
+use Drupal\Core\Config\ExtensionInstallStorage;
+use Drupal\Core\Extension\Extension;
+use Drupal\Core\Extension\ModuleHandlerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Drupal\Core\Form\ConfigFormBase;
+use Drupal\Core\Entity\EntityFormInterface;
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
 
 /**
- * Check if the given form should be read-only.
+ * Class ConfigReadOnlyEventSubscriber.
+ *
+ * This event listener replaces
+ * \Drupal\config_readonly\EventSubscriber\ReadOnlyFormSubscriber because it
+ * gives a "whitelist" capability but we want it to add a blacklist to lock
+ * configurations provided by modules.
+ *
+ * @package Drupal\hs_config_readonly\EventSubscriber
  */
 class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
 
-  public function onFormAlter(ReadOnlyFormEvent $event) {
+  use ConfigReadonlyWhitelistTrait;
 
+  /**
+   * @var \Drupal\Core\Config\ExtensionInstallStorage
+   */
+  protected $extensionConfigStorage;
+
+  /**
+   * @var \Drupal\Core\Config\ExtensionInstallStorage
+   */
+  protected $extensionOptionalConfigStorage;
+
+  /**
+   * @var array
+   */
+  protected $excludedModules = ['system'];
+
+  /**
+   * @var string
+   */
+  protected $drupalRoot;
+
+  /**
+   * Form ids to mark as read only.
+   */
+  protected $readOnlyFormIds = [
+    'config_single_import_form',
+    'system_modules',
+    'system_modules_uninstall',
+    'user_admin_permissions',
+  ];
+
+  /**
+   * Form ids to skip marking as read only.
+   */
+  protected $formIdExceptions = [
+    'search_form',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct($drupal_root, ModuleHandlerInterface $module_handler, ExtensionInstallStorage $extension_config_storage, ExtensionInstallStorage $extension_optional_config_storage) {
+    $this->drupalRoot = $drupal_root;
+    $this->moduleHandler = $module_handler;
+    $this->extensionConfigStorage = $extension_config_storage;
+    $this->extensionOptionalConfigStorage = $extension_optional_config_storage;
+  }
+
+  public function onFormAlter(ReadOnlyFormEvent $event) {
+    // Check if the form is a ConfigFormBase or a ConfigEntityListBuilder.
+    $form_object = $event->getFormState()->getFormObject();
+    $mark_form_read_only = $form_object instanceof ConfigFormBase;
+
+    if (!$mark_form_read_only) {
+      $mark_form_read_only = in_array($form_object->getFormId(), $this->readOnlyFormIds);
+    }
+
+    // Check if the form is an EntityFormInterface and entity is a config
+    // entity.
+    if (!$mark_form_read_only && $form_object instanceof EntityFormInterface) {
+      $entity = $form_object->getEntity();
+      $mark_form_read_only = $entity instanceof ConfigEntityInterface;
+    }
+
+    // Don't block particular patterns.
+    if ($mark_form_read_only && $form_object instanceof EntityFormInterface) {
+      $entity = $form_object->getEntity();
+      $name = $entity->getConfigDependencyName();
+
+      // Block config from a module.
+      $mark_form_read_only = $this->configIsEditable([$name]);
+
+      // Unless its defined as a whiteist pattern.
+      if ($this->matchesWhitelistPattern($name)) {
+        $mark_form_read_only = FALSE;
+      }
+    }
+
+    // Config forms.
+    if ($mark_form_read_only && $form_object instanceof ConfigFormBase) {
+      $editable_config = $this->getEditableConfigNames($form_object);
+      $mark_form_read_only = $this->configIsEditable($editable_config);
+    }
+
+    if ($mark_form_read_only) {
+      $event->markFormReadOnly();
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+   * {@inheritdoc}
+   */
+  public function __construct($drupal_root, ModuleHandlerInterface $module_handler, ExtensionInstallStorage $extension_config_storage, ExtensionInstallStorage $extension_optional_config_storage) {
+    $this->drupalRoot = $drupal_root;
+    $this->moduleHandler = $module_handler;
+    $this->extensionConfigStorage = $extension_config_storage;
+    $this->extensionOptionalConfigStorage = $extension_optional_config_storage;
+  }
+
+  public function onFormAlter(ReadOnlyFormEvent $event) {
```
sherakama: Missing doc block.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    // Config forms.
+    if ($mark_form_read_only && $form_object instanceof ConfigFormBase) {
+      $editable_config = $this->getEditableConfigNames($form_object);
+      $mark_form_read_only = $this->configIsEditable($editable_config);
+    }
+
+    if ($mark_form_read_only) {
+      $event->markFormReadOnly();
```
sherakama: This is a handy little method.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return TRUE;
+  }
+
+  /**
+   * @param $name
+   *
+   * @return bool
+   */
+  protected function isEditableConfig($name) {
```
sherakama: Why do we have two checks for editable config? Yes, I know they are different but it is confusing to have both of these functions esp without a docblock comment outlining why.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: yeah i wish it had a `$event->markFormEditable()` or something. cuz then i wouldn't need to replace the service. I could just create another event subscriber

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+   *   The configuration form.
+   *
+   * @return array
+   *   An array of configuration object names that are editable if called in
+   *   conjunction with the trait's config() method.
+   *
+   * @throws \ReflectionException
+   */
+  protected function getEditableConfigNames(ConfigFormBase $form) {
+    // Use reflection to work around getEditableConfigNames() as protected.
```
sherakama: Thanks for this note.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+   */
+  protected function getLockedConfigs() {
+    $module_configs = [];
+    foreach ($this->moduleHandler->getModuleList() as $name => $extension) {
+      if ($extension->getType() == 'profile' || in_array($name, $this->excludedModules)) {
+        continue;
+      }
+
+      $install_list = $this->listProvidedItems($name);
+      $optional_list = $this->listProvidedItems($name, TRUE);
```
sherakama: 🤔 Are you sure you want to lock the optional ones as well?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: Stole it from the original event subscriber lol

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Get all configs provided by modules.
+   *
+   * @return array
+   *   Config names.
+   */
+  protected function getLockedConfigs() {
+    $module_configs = [];
+    foreach ($this->moduleHandler->getModuleList() as $name => $extension) {
+      if ($extension->getType() == 'profile' || in_array($name, $this->excludedModules)) {
```
sherakama: Why not profiles?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: I figured i'd treat them the same as install ones. since the optional ones could consist of fields that rely on other content types. like a entity reference field.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: I thought maybe profiles would be a starting point. i kinda went both directions with that.

pookmish:
Oh forgot to have this in the steps... add `$settings['config_readonly'] = TRUE;` to your settings.php or local.settings.php file

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+   * @param string $name
+   *   Machine name of extension.
+   * @param bool $do_optional
+   *   FALSE (default) to list config/install items, TRUE to list
+   *   config/optional items.
+   *
+   * @return string[]
+   *   List of config items provided by this extension.
+   */
+  protected function listProvidedItems($name, $do_optional = FALSE) {
```
sherakama: I would expect this function by the name to always return the config/install items and return both config/install and config/optional if $do_optional == true. Could you refactor this name to be more clear?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
sherakama: This seems like something that should be configurable. The peel-and-stick approach to this module works but what if this was based off of distribution?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
sherakama: Would it be possible that you could get in to a scenario that would prevent the user from configuring a new module that was turned on through the UI because it's optional config is now available?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
sherakama: Take note. 😆

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: its a take over of the over-ridden service. i'm happy to rename and put a comment referencing the method is comes from.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: well right now modules can only be turned on via drush. and yes it would prevent that. maybe we limit the modules/profiles that are "blacklisted"?

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: I made it configurable. `hs_config_readonly.yml` can define which modules to ignore. by default if the config doesn't exist, it ignores `system` module.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: i renamed the variable, but it was noted in the comments above the method ;)

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
pookmish: Should we make a UI for this?

pookmish: or it can be easily changed via Drush on a per-site basis but be hidden from the user.

pookmish:
ready for another review @sherakama / @boznik

docroot/modules/humsci/hs_config_readonly/hs_config_readonly.module
```diff
@@ -0,0 +1,15 @@
...
+ * @file
+ * hs_config_readonly.module
+ */
+
+/**
+ * Implements hook_config_readonly_whitelist_patterns().
+ */
+function hs_config_readonly_config_readonly_whitelist_patterns() {
+  // Lets whitelist everything because in our event subscriber we have the
+  // ability to change hose the forms are locked.
```
sherakama: change which hose?

pookmish: No idea. lol

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
sherakama: Note that I want to hammer this point home, I just could note help myself from posting one more note.

docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
sherakama: Yeah, I think this will massage itself out over time. Keep this scenario in mind when you get a user that is blocked from new config screens. I am sure we would want to define a scenario or have a list of items to whitelist.

PR #17, opened 2018-05-25
ARCH-71 Set up publications display and masonry pattern

pookmish:
# READY FOR REVIEW

# Summary
- Set the publications node display using layout builder
- Added the publications view using masonry pattern.

# Needed By (Date)
- next week

# Urgency
- Low

# Steps to Test

1. Checkout this branch
2. install a clean version
2a. suggested disable `acquia_connector`, `simplesamlphp_auth`, & `config_readonly`
3. enable hs_publications
4. create some publications via the UI or devel_generate
5. ensure a publication page looks somewhat normal
6. create an empty "Publications" basic page.
7. click on "layout" to change the basic page contents
8. add the publications masonry list view and the exposed filter block
9. ensure masonry looks correct and the view functions as expected. [mockup](https://projects.invisionapp.com/d/main#/console/14029316/292719604/preview)

# See Also


docroot/modules/humsci/hs_field_helpers/templates/views-view-pattern.html.twig
```diff
@@ -1,3 +1,3 @@
-<div{{ attributes }}>
+<div{{ attributes.addClass('views-view-pattern') }}>
```
sherakama: I have found with earth that this file had to be available in a module as well.

docroot/modules/humsci/hs_layouts/hs_layouts.libraries.yml
```diff
@@ -1,3 +1,9 @@
+admin:
```
sherakama: I went the way of putting layouts in to a module for earth and found that because grids can shift and styles can change it is better to have all of these in the theme. Let's chat about this sometime.

docroot/modules/humsci/hs_field_helpers/templates/views-view-pattern.html.twig
pookmish: I recently found that the "pattern" style only correctly works with a pattern row plugin. If i try to use just fields, it doesn't wrap each row in a div. which broke when i tried to do a masonry pattern.
I'll have to take a look at that more later to step trough for that scenario

pookmish:
Poke https://media1.tenor.com/images/61e979527421fff4ab8b5f07f55875ab/tenor.gif?itemid=3682545

PR #23, opened 2018-06-13
ARCH-72 event card template and styles, new mixins, latest decanter and bourbon

josephgknox:
# READY FOR REVIEW

# Summary
- Horizontal Card, Vertical Card, and Date Stacked Vertical Card Twig, SCSS (and CSS), and UI Components files
- New card (with custom display for horizontal layout) and accent dash mixins
- Updated package.json to include latest versions of Decanter and Bourbon
- Custom Archaeology styles and updates to the Gruntfile.js as needed

# Needed By (Date)
- End of sprint (6/21)

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Ensure that the /patterns page on local shows a horizontal, vertical, and date stacked card that can be used for patterns on the mockups at https://invis.io/MHLC2CFGQYV#/299456614_Homepage and https://projects.invisionapp.com/d/main#/console/14029316/292457736/preview

- Archaeology
- suhumsci

- https://stanfordits.atlassian.net/browse/ARCH-72
- https://stanfordits.atlassian.net/browse/ARCH-83
- https://stanfordits.atlassian.net/browse/ARCH-81

# See Also


josephgknox:
FYI @bixgomez

docroot/themes/humsci/su_humsci_theme/package.json
```diff
@@ -6,9 +6,9 @@
   "license": "MIT",
   "repository": "https://github.com/SU-HSDO/suhumsci",
   "devDependencies": {
-    "bourbon": "^4.2.7",
+    "bourbon": "^5.0.0",
     "bourbon-neat": "^1.8.0",
-    "decanter": "^2.0.0",
+    "decanter": "^3.0.0",
```
pookmish: tumblr_moqmjwe2sa1rh4lfno2_1280-1

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.html.twig
```diff
@@ -0,0 +1,38 @@
+{% set attributes = attributes.addClass('vertical-card') %}
+
+<div{{ attributes }}>
+
+  <div class="featured-card__container">
+    {% if image %}
```
sherakama: This if statement does not wrap the closing </div> and could leaving trailing </div>s where it shouldnt. Wrap the whole statement or add another if statement around the closing div.

josephgknox: Thanks, @sherakama. Fixed.

pookmish: Shouldn't we have some styles for the date when the image is not present.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.ui_patterns.yml
```diff
@@ -0,0 +1,36 @@
+featured_card:
```
pookmish: Maybe we can rename this to something like "dated_vertical_card"

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.html.twig
```diff
@@ -0,0 +1,40 @@
+{% set attributes = attributes.addClass('vertical-card') %}
```
pookmish: This class should be the same as the pattern id

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/scss/featured-card.scss
```diff
@@ -0,0 +1,35 @@
...
+// Decanter
+@import
+  'decanter-no-markup';
+
+.featured-card__container {
+  .featured-card__img {
+    position: relative;
+
+    img {
+      width: 100%;
```
pookmish: I don't like this. it could cause small images to stretch.. without this, it would allow the user to choose which image size to provide using image styles.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
```diff
@@ -0,0 +1,24 @@
...
+//
+
+// Decanter
+@import 'decanter-no-markup';
+
+// SU HumSci
+@import
+  '../utilities/mixins/index';
+
+.horizontal-card {
```
pookmish: Why is this & `.vertical-card` getting added to global styles? these styles should be in the pattern.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.html.twig
josephgknox: For this pattern, the date is nested within the image <div>. So, if there is no image, there will be no date. Also, the more global vertical (or even horizontal depending on use case) should be used for when there is no date stacking over an image.

pookmish: Ok. just wanted to make sure thats the desired outcome. use case: an event without an image but has a date.... 🤔

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.ui_patterns.yml
josephgknox: Yeah, more specific here is probably good. That is, what separates this from a vertical card? I am going with date_stacked_vertical_card.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.html.twig
josephgknox: Word. Good catch.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
josephgknox: If I can't think of a logical defense, then it should be changed. :)

josephgknox: Updated.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/scss/featured-card.scss
josephgknox: Still thinking about this. The height will grow automatically (we will not hard code that) depending on the width. I don't like the potential pixelation, either. Maybe we allow for larger image sizes only that then scale when used with a decanter grid class, so that the pixelation is never bad. I don't want to hardcode a width on the content container either, in case the image style changes. This felt like the most global option, as I also don't like how the card looks with an image that only spans its natural width when the content container stretches 100%.

pookmish: screen shot 2018-06-19 at 1 34 07 pm
I removed the image width 100% and applied `width: fit-content;` to the wrapper with the box shadow.

josephgknox: Nice! The image needs to be edge-to-edge, though. Wondering if we can use your solution, but remove the padding from the image.

josephgknox: No luck so far. Also, not only do we lose the edge-to-edge image, but the fit content for a container will find the largest but of content. See what happens when I update the title of the card:
screen shot 2018-06-19 at 2 10 34 pm

pookmish: well thats better IMO. we will have something in place to make them have 2 or 3 side by side. but at least this way the image doesnt stretch. a more ideal thing for the users i think. but its your front end too sooo.... dont let me dictate it :)

PR #26, opened 2018-06-15
Arch 83 feat hp views

josephgknox:


docroot/themes/humsci/su_humsci_theme/patterns/molecules/featured-card/featured-card.html.twig
```diff
@@ -1,20 +1,38 @@
...
 <div{{ attributes }}>
 
-<div class="featured-card__container">
-  <div class="featured-card__img">
-  {{ image }}
-  <div class="date-stacked">
-    <div class="date-month">{{ month }}</div>
-    <div class="date-day">{{ day }}</div>
+  <div class="featured-card__container">
+    {% if image %}
```
sherakama: This if statement could cause problems. It is wrapping the opening div tag for an image but there is no if statement for the closing tag. This could leave hanging </div>s

sherakama: @josephgknox This needs to be fixed.

josephgknox: Thanks, @sherakama! I'm combining some work from a few tickets, so will be updating this on the ARCH-72 branch (which you also commented on).

PR #31, opened 2018-06-21
Saml entitlement

pookmish:
# READY FOR REVIEW

# Summary
- Added workgroup mapping
- modified blt deploy to exclude keys and db.sql files.

# Needed By (Date)
- whenevers convienent

# Urgency
- Low

# Steps to Test

1. Manual review is about necessary.

# See Also


docroot/sites/archaeology/settings.php
```diff
@@ -820,23 +820,20 @@
...
   'register_users' => TRUE,
   'autoenablesaml' => TRUE,
-  'debug' => TRUE,
+  'debug' => FALSE,
   'secure' => FALSE,
   'httponly' => FALSE,
   'role' => [
-    'population' => 'administrator:suAffiliation,=,hsdo:web|administrator:suAffiliation,=,itservices:webservices',
-    'eval_every_time' => FALSE,
+    'population' => 'administrator:eduPersonEntitlement,=,hsdo:web|administrator:eduPersonEntitlement,=,itservices:webservices',
```
sherakama: Are you sure you want to hard wire these settings? Isn't there a UI to input these?

pookmish: There is a UI, i just took the recommendation on https://www.chapterthree.com/blog/how-to-configure-simplesamlphp-for-drupal-8-on-acquia
And these are the only 2 workgroups released right now. if we switch to use your API call thingy, we can probably move this to normal config.

sherakama: hmm.. Okay. That is for future us to worry about then. Thanks for the explanation.

PR #35, opened 2018-07-03
Hsd8 147 config split

pookmish:
# READY FOR REVIEW (Maybe?)

# Summary
- Set the config directories for archaeology and default
- added environment config splits
- added common config split for shared configs
- Modified hs_config_readonly to look at the staging storage excluding config_ignore settings. This locks down the content type and fields, but opens up the form and display configs.

# Things to note
- Doing symlinks like `config/archaeology/hs_events` to point to `docroot/modules/humsci/hs_events/config/install` kinda works. but requires collection manipulation. Meaning that it deletes every hs_events config and re-imports it as the new collection. Maybe we can do that later since it would involve DB manipulation. Also how would that work on new sites? idk
- Config split for features kinda worked too. but caused conflicts with dependencies.
- UUIDs on the config entities might be an issue in site #2. Unless we simply take this site, clone it and use it as site #2 starter. Might need more discussion with H&S on how they forsee future sites

# Needed By (Date)
- asap

# Urgency
- mild

# Steps to Test
1. Checkout branch.
1. `blt drupal:sync --site=archaeology`
1. review the logs and make sure config gets imported

# See Also

I've attached a git diff from develop that ignores all the config files. not sure if its more helpful.

```diff --git a/docroot/modules/humsci/hs_basic_page/config/install/.htaccess b/docroot/modules/humsci/hs_basic_page/config/install/.htaccess
new file mode 100644
index 0000000..0e4a69a
--- /dev/null
+++ b/docroot/modules/humsci/hs_basic_page/config/install/.htaccess
@@ -0,0 +1,24 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php5.c>
+  php_flag engine off
+</IfModule>
\ No newline at end of file
diff --git a/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.info.yml b/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.info.yml
index 6df9b89..b33e810 100644
--- a/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.info.yml
+++ b/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.info.yml
@@ -6,3 +6,4 @@ version: 8.x
 package: 'Humanities & Sciences'
 dependencies:
   - config_readonly
+  - config_filter
diff --git a/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.services.yml b/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.services.yml
index dbb3e8b..a8c38fe 100644
--- a/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.services.yml
+++ b/docroot/modules/humsci/hs_config_readonly/hs_config_readonly.services.yml
@@ -1,14 +1,6 @@
 services:
-  hs_config_readonly.extension_storage:
-    class: Drupal\Core\Config\ExtensionInstallStorage
-    arguments: ['@config.storage']
-
-  hs_config_readonly.extension_optional_storage:
-    class: Drupal\Core\Config\ExtensionInstallStorage
-    arguments: ['@config.storage', 'config/optional']
-
   hs_config_readonly.form_subscribe:
     class: Drupal\hs_config_readonly\EventSubscriber\ConfigReadOnlyEventSubscriber
     decorates: config_readonly_form_subscriber
     public: false
-    arguments: ['@app.root', '@module_handler', '@config.factory', '@hs_config_readonly.extension_storage', '@hs_config_readonly.extension_optional_storage']
+    arguments: ['@module_handler', '@config.factory', '@config.storage.staging', '@plugin.manager.config_filter']
diff --git a/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php b/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
index 9a0d44a..0e31d47 100644
--- a/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
+++ b/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
@@ -2,11 +2,11 @@
 
 namespace Drupal\hs_config_readonly\EventSubscriber;
 
+use Drupal\config_filter\Config\FilteredStorageInterface;
+use Drupal\config_filter\Plugin\ConfigFilterPluginManager;
 use Drupal\config_readonly\ConfigReadonlyWhitelistTrait;
 use Drupal\config_readonly\ReadOnlyFormEvent;
 use Drupal\Core\Config\ConfigFactoryInterface;
-use Drupal\Core\Config\ExtensionInstallStorage;
-use Drupal\Core\Extension\Extension;
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Drupal\Core\Form\ConfigFormBase;
@@ -28,18 +28,14 @@ class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
   use ConfigReadonlyWhitelistTrait;
 
   /**
-   * Installed configuration service.
-   *
-   * @var \Drupal\Core\Config\ExtensionInstallStorage
+   * @var \Drupal\config_filter\Config\FilteredStorageInterface
    */
-  protected $extensionConfigStorage;
+  protected $configStorage;
 
   /**
-   * Optional configuration service.
-   *
-   * @var \Drupal\Core\Config\ExtensionInstallStorage
+   * @var \Drupal\config_filter\Plugin\ConfigFilterPluginManager
    */
-  protected $extensionOptionalConfigStorage;
+  protected $configFilterManager;
 
   /**
    * Ignore the configurations form these modules.
@@ -48,13 +44,6 @@ class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
    */
   protected $excludedModules = ['system'];
 
-  /**
-   * Root path of Drupal.
-   *
-   * @var string
-   */
-  protected $drupalRoot;
-
   /**
    * Form ids to mark as read only.
    */
@@ -68,12 +57,10 @@ class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
   /**
    * {@inheritdoc}
    */
-  public function __construct($drupal_root, ModuleHandlerInterface $module_handler, ConfigFactoryInterface $config_factory, ExtensionInstallStorage $extension_config_storage, ExtensionInstallStorage $extension_optional_config_storage) {
-    $this->drupalRoot = $drupal_root;
+  public function __construct(ModuleHandlerInterface $module_handler, ConfigFactoryInterface $config_factory, FilteredStorageInterface $config_storage, ConfigFilterPluginManager $filter_manager) {
     $this->moduleHandler = $module_handler;
-    $this->extensionConfigStorage = $extension_config_storage;
-    $this->extensionOptionalConfigStorage = $extension_optional_config_storage;
-
+    $this->configStorage = $config_storage;
+    $this->configFilterManager = $filter_manager;
     $config = $config_factory->get('hs_config_readonly.settings');
     $this->excludedModules = $config->get('excluded_modules') ?: $this->excludedModules;
     $this->readOnlyFormIds = $config->get('form_ids') ?: $this->readOnlyFormIds;
@@ -110,17 +97,17 @@ public function onFormAlter(ReadOnlyFormEvent $event) {
     if (!$mark_form_read_only && $form_object instanceof EntityFormInterface) {
       $entity = $form_object->getEntity();
       $mark_form_read_only = $entity instanceof ConfigEntityInterface;
-    }
 
-    // Don't block particular patterns.
-    if ($mark_form_read_only && $form_object instanceof EntityFormInterface) {
-      $entity = $form_object->getEntity();
-      $name = $entity->getConfigDependencyName();
+      // Don't block particular patterns.
+      if ($mark_form_read_only) {
+        $entity = $form_object->getEntity();
+        $name = $entity->getConfigDependencyName();
 
-      // Block config from a module.
-      $mark_form_read_only = $this->configIsLocked($name);
-      // If all config is in the whitelist, do not block the form.
-      $mark_form_read_only = $this->matchesWhitelistPattern($name) ? FALSE : $mark_form_read_only;
+        // Block config from a module.
+        $mark_form_read_only = $this->configIsLocked($name);
+        // If all config is in the whitelist, do not block the form.
+        $mark_form_read_only = $this->matchesWhitelistPattern($name) ? FALSE : $mark_form_read_only;
+      }
     }
 
     // Config forms.
@@ -183,43 +170,20 @@ protected function configIsLocked($config) {
    *   Config names.
    */
   protected function getLockedConfigs() {
-    $module_configs = [];
-    foreach ($this->moduleHandler->getModuleList() as $name => $extension) {
-      if (in_array($name, $this->excludedModules)) {
-        continue;
-      }
-
-      $install_list = $this->listProvidedItems($name);
-      $optional_list = $this->listProvidedItems($name, TRUE);
-      $module_configs = array_merge($module_configs, $install_list, $optional_list);
+    $configs = $this->configStorage->listAll();
+    if (!$this->configFilterManager->hasDefinition('config_ignore')) {
+      return $configs;
     }
-    return $module_configs;
-  }
 
-  /**
-   * Returns a list of the install storage items for an extension.
-   *
-   * @param string $name
-   *   Machine name of extension.
-   * @param bool $only_optional
-   *   FALSE (default) to list config/install items, TRUE to list
-   *   config/optional items.
-   *
-   * @return string[]
-   *   List of config items provided by this extension.
-   *
-   * @see \Drupal\config_update\ConfigLister::listProvidedItems().
-   */
-  protected function listProvidedItems($name, $only_optional = FALSE) {
-    $pathname = drupal_get_filename('module', $name);
-    $component = new Extension($this->drupalRoot, 'module', $pathname);
-    if ($only_optional) {
-      $names = $this->extensionOptionalConfigStorage->getComponentNames([$component]);
-    }
-    else {
-      $names = $this->extensionConfigStorage->getComponentNames([$component]);
+    /** @var \Drupal\config_ignore\Plugin\ConfigFilter\IgnoreFilter $plugin */
+    $plugin = $this->configFilterManager->createInstance('config_ignore');
+    foreach ($plugin->filterListAll('', []) as $ignored_config) {
+      $pos = array_search($ignored_config, $configs);
+      if ($pos !== FALSE) {
+        unset($configs[$pos]);
+      }
     }
-    return array_keys($names);
+    return $configs;
   }
 
   /**
diff --git a/docroot/modules/humsci/hs_event_series/config/install/.htaccess b/docroot/modules/humsci/hs_event_series/config/install/.htaccess
new file mode 100644
index 0000000..0e4a69a
--- /dev/null
+++ b/docroot/modules/humsci/hs_event_series/config/install/.htaccess
@@ -0,0 +1,24 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php5.c>
+  php_flag engine off
+</IfModule>
\ No newline at end of file
diff --git a/docroot/modules/humsci/hs_events/config/install/.htaccess b/docroot/modules/humsci/hs_events/config/install/.htaccess
new file mode 100644
index 0000000..0e4a69a
--- /dev/null
+++ b/docroot/modules/humsci/hs_events/config/install/.htaccess
@@ -0,0 +1,24 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php5.c>
+  php_flag engine off
+</IfModule>
\ No newline at end of file
diff --git a/docroot/modules/humsci/hs_person/config/install/.htaccess b/docroot/modules/humsci/hs_person/config/install/.htaccess
new file mode 100644
index 0000000..0e4a69a
--- /dev/null
+++ b/docroot/modules/humsci/hs_person/config/install/.htaccess
@@ -0,0 +1,24 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php5.c>
+  php_flag engine off
+</IfModule>
\ No newline at end of file
diff --git a/docroot/modules/humsci/hs_publications/config/install/.htaccess b/docroot/modules/humsci/hs_publications/config/install/.htaccess
new file mode 100644
index 0000000..0e4a69a
--- /dev/null
+++ b/docroot/modules/humsci/hs_publications/config/install/.htaccess
@@ -0,0 +1,24 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
+</IfModule>
+
+# Deny all requests from Apache 2.0-2.2.
+<IfModule !mod_authz_core.c>
+  Deny from all
+</IfModule>
+
+# Turn off all options we don't need.
+Options -Indexes -ExecCGI -Includes -MultiViews
+
+# Set the catch-all handler to prevent scripts from being executed.
+SetHandler Drupal_Security_Do_Not_Remove_See_SA_2006_006
+<Files *>
+  # Override the handler again if we're run later in the evaluation list.
+  SetHandler Drupal_Security_Do_Not_Remove_See_SA_2013_003
+</Files>
+
+# If we know how to do it safely, disable the PHP engine entirely.
+<IfModule mod_php5.c>
+  php_flag engine off
+</IfModule>
\ No newline at end of file
diff --git a/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install b/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
index 24ce6b1..933bde2 100644
--- a/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
+++ b/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
@@ -51,7 +51,25 @@ function su_humsci_profile_install() {
 /**
  * Enables config readonly module after profile finishes.
  */
-function su_humsci_profile_lock_config(&$install_state){
+function su_humsci_profile_lock_config(&$install_state) {
   install_finished($install_state);
   \Drupal::service('module_installer')->install(['hs_config_readonly']);
 }
+
+/**
+ * Create config split entities.
+ */
+function su_humsci_profile_update_8002() {
+  \Drupal::service('module_installer')->install([
+    'config_split',
+    'config_ignore',
+  ]);
+
+  $config_path = '../config/common';
+  $source = new FileStorage($config_path);
+  /** @var \Drupal\Core\Config\CachedStorage $config_storage */
+  $config_storage = \Drupal::service('config.storage');
+  foreach ($source->listAll() as $name) {
+    $config_storage->write($name, $source->read($name));
+  }
+}
diff --git a/docroot/sites/archaeology/settings.php b/docroot/sites/archaeology/settings.php
index 0230c4e..dc9b21e 100755
--- a/docroot/sites/archaeology/settings.php
+++ b/docroot/sites/archaeology/settings.php
@@ -790,98 +790,8 @@
 #   include $app_root . '/' . $site_path . '/settings.local.php';
 # }
 
-require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";
-
-// SimpleSAMLphp configuration
-// Provide universal absolute path to the installation.
-if (isset($_ENV['AH_SITE_NAME']) && is_dir('/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp')) {
-  $settings['simplesamlphp_dir'] = '/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp';
-}
-else {
-  // Local SAML path.
-  if (is_dir(DRUPAL_ROOT . '/../simplesamlphp')) {
-    $settings['simplesamlphp_dir'] = DRUPAL_ROOT . '/../simplesamlphp';
-  }
-}
-
-if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT']) && $_ENV['AH_SITE_ENVIRONMENT'] == 'prod') {
-  $config['system.file']['path']['temporary'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp";
-}
-else {
-  $config['system.file']['path']['temporary'] = sys_get_temp_dir();
-}
-
-$config['simplesamlphp_auth.settings'] = [
-  'langcode' => 'en',
-  'default_langcode' => 'en',
-  'activate' => TRUE,
-  'mail_attr' => 'eduPersonPrincipalName',
-  'unique_id' => 'uid',
-  'user_name' => 'displayName',
-  'auth_source' => 'default-sp',
-  'login_link_display_name' => 'Stanford Login',
-  'header_no_cache' => TRUE,
-  'user_register_original' => 'visitors',
-  'register_users' => TRUE,
-  'autoenablesaml' => TRUE,
-  'debug' => FALSE,
-  'secure' => FALSE,
-  'httponly' => FALSE,
-  'role' => [
-    'population' => 'administrator:eduPersonEntitlement,=,hsdo:web|administrator:eduPersonEntitlement,=,itservices:webservices',
-    'eval_every_time' => TRUE,
-  ],
-  'allow' => [
-    'set_drupal_pwd' => FALSE,
-    'default_login' => TRUE,
-  ],
-  'sync' => [
-    'mail' => TRUE,
-    'user_name' => TRUE,
-  ],
-];
-
-$config['environment_indicator.indicator']['bg_color'] = '#086601';
-$config['environment_indicator.indicator']['fg_color'] = '#fff';
-$config['environment_indicator.indicator']['name'] = 'Local';
-
-if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT'])) {
-  switch ($_ENV['AH_SITE_ENVIRONMENT']) {
-    case 'dev':
-      $config['environment_indicator.indicator']['bg_color'] = '#6B0500';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Development';
-      break;
-    case 'test':
-      $config['environment_indicator.indicator']['bg_color'] = '#4127C2';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Staging';
-      break;
-    case 'prod':
-      $config['environment_indicator.indicator']['bg_color'] = '#000';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Production';
-      break;
-    default:
-      $config['environment_indicator.indicator']['bg_color'] = '#086601';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = $_ENV['AH_SITE_ENVIRONMENT'];
-      break;
-  }
-}
-
-if (isset($_ENV['AH_SITE_ENVIRONMENT'])) {
-  if (PHP_SAPI !== 'cli') {
-    // Don't lock config when using drush.
-    $settings['config_readonly'] = TRUE;
-  }
-}
-
-// Lets whitelist everything because in our event subscriber we have the
-// ability to decide which forms are locked.
-// @see \Drupal\hs_config_readonly\EventSubscriber\ConfigReadOnlyEventSubscriber
-$settings['config_readonly_whitelist_patterns'] = ['*'];
-
 if (file_exists('/var/www/site-php')) {
   require '/var/www/site-php/swshumsci/archaeology-settings.inc';
 }
+
+require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";
diff --git a/docroot/sites/default/settings.php b/docroot/sites/default/settings.php
index 32ae005..acb2abc 100755
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -791,93 +791,3 @@
 # }
 
 require DRUPAL_ROOT . "/../vendor/acquia/blt/settings/blt.settings.php";
-
-// SimpleSAMLphp configuration
-// Provide universal absolute path to the installation.
-if (isset($_ENV['AH_SITE_NAME']) && is_dir('/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp')) {
-  $settings['simplesamlphp_dir'] = '/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp';
-}
-else {
-  // Local SAML path.
-  if (is_dir(DRUPAL_ROOT . '/../simplesamlphp')) {
-    $settings['simplesamlphp_dir'] = DRUPAL_ROOT . '/../simplesamlphp';
-  }
-}
-
-if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT']) && $_ENV['AH_SITE_ENVIRONMENT'] == 'prod') {
-  $config['system.file']['path']['temporary'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp";
-}
-else {
-  $config['system.file']['path']['temporary'] = sys_get_temp_dir();
-}
-
-$config['simplesamlphp_auth.settings'] = [
-  'langcode' => 'en',
-  'default_langcode' => 'en',
-  'activate' => TRUE,
-  'mail_attr' => 'eduPersonPrincipalName',
-  'unique_id' => 'uid',
-  'user_name' => 'displayName',
-  'auth_source' => 'default-sp',
-  'login_link_display_name' => 'Stanford Login',
-  'header_no_cache' => TRUE,
-  'user_register_original' => 'visitors',
-  'register_users' => TRUE,
-  'autoenablesaml' => TRUE,
-  'debug' => FALSE,
-  'secure' => FALSE,
-  'httponly' => FALSE,
-  'role' => [
-    'population' => 'administrator:eduPersonEntitlement,=,hsdo:web|administrator:eduPersonEntitlement,=,itservices:webservices',
-    'eval_every_time' => TRUE,
-  ],
-  'allow' => [
-    'set_drupal_pwd' => FALSE,
-    'default_login' => TRUE,
-  ],
-  'sync' => [
-    'mail' => TRUE,
-    'user_name' => TRUE,
-  ],
-];
-
-$config['environment_indicator.indicator']['bg_color'] = '#086601';
-$config['environment_indicator.indicator']['fg_color'] = '#fff';
-$config['environment_indicator.indicator']['name'] = 'Local';
-
-if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT'])) {
-  switch ($_ENV['AH_SITE_ENVIRONMENT']) {
-    case 'dev':
-      $config['environment_indicator.indicator']['bg_color'] = '#6B0500';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Development';
-      break;
-    case 'test':
-      $config['environment_indicator.indicator']['bg_color'] = '#4127C2';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Staging';
-      break;
-    case 'prod':
-      $config['environment_indicator.indicator']['bg_color'] = '#000';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = 'Production';
-      break;
-    default:
-      $config['environment_indicator.indicator']['bg_color'] = '#086601';
-      $config['environment_indicator.indicator']['fg_color'] = '#fff';
-      $config['environment_indicator.indicator']['name'] = $_ENV['AH_SITE_ENVIRONMENT'];
-      break;
-  }
-}
-
-if (isset($_ENV['AH_SITE_ENVIRONMENT'])) {
-  if (PHP_SAPI !== 'cli') {
-    // Don't lock config when using drush.
-    $settings['config_readonly'] = TRUE;
-  }
-}
-
-// Lets whitelist everything because in our event subscriber we have the
-// ability to decide which forms are locked.
-// @see \Drupal\hs_config_readonly\EventSubscriber\ConfigReadOnlyEventSubscriber
-$settings['config_readonly_whitelist_patterns'] = ['*'];
diff --git a/docroot/sites/settings/common.settings.php b/docroot/sites/settings/common.settings.php
new file mode 100644
index 0000000..ef04b25
--- /dev/null
+++ b/docroot/sites/settings/common.settings.php
@@ -0,0 +1,101 @@
+<?php
+
+/**
+ * @file
+ * This file gets added by blt.settings.php as an override configuration.
+ */
+
+// SimpleSAMLphp configuration
+// Provide universal absolute path to the installation.
+if (isset($_ENV['AH_SITE_NAME']) && is_dir('/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp')) {
+  $settings['simplesamlphp_dir'] = '/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp';
+}
+else {
+  // Local SAML path.
+  if (is_dir(DRUPAL_ROOT . '/../simplesamlphp')) {
+    $settings['simplesamlphp_dir'] = DRUPAL_ROOT . '/../simplesamlphp';
+  }
+}
+
+if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT']) && $_ENV['AH_SITE_ENVIRONMENT'] == 'prod') {
+  $config['system.file']['path']['temporary'] = "/mnt/gfs/{$_ENV['AH_SITE_GROUP']}.{$_ENV['AH_SITE_ENVIRONMENT']}/tmp";
+}
+else {
+  $config['system.file']['path']['temporary'] = sys_get_temp_dir();
+}
+
+$config['simplesamlphp_auth.settings'] = [
+  'langcode' => 'en',
+  'default_langcode' => 'en',
+  'activate' => TRUE,
+  'mail_attr' => 'eduPersonPrincipalName',
+  'unique_id' => 'uid',
+  'user_name' => 'displayName',
+  'auth_source' => 'default-sp',
+  'login_link_display_name' => 'Stanford Login',
+  'header_no_cache' => TRUE,
+  'user_register_original' => 'visitors',
+  'register_users' => TRUE,
+  'autoenablesaml' => TRUE,
+  'debug' => FALSE,
+  'secure' => FALSE,
+  'httponly' => FALSE,
+  'role' => [
+    'population' => 'administrator:eduPersonEntitlement,=,hsdo:web|administrator:eduPersonEntitlement,=,itservices:webservices',
+    'eval_every_time' => TRUE,
+  ],
+  'allow' => [
+    'set_drupal_pwd' => FALSE,
+    'default_login' => TRUE,
+  ],
+  'sync' => [
+    'mail' => TRUE,
+    'user_name' => TRUE,
+  ],
+];
+
+$config['environment_indicator.indicator']['bg_color'] = '#086601';
+$config['environment_indicator.indicator']['fg_color'] = '#fff';
+$config['environment_indicator.indicator']['name'] = 'Local';
+
+if (isset($_ENV) && isset($_ENV['AH_SITE_GROUP']) && isset($_ENV['AH_SITE_ENVIRONMENT'])) {
+  switch ($_ENV['AH_SITE_ENVIRONMENT']) {
+    case 'dev':
+      $config['environment_indicator.indicator']['bg_color'] = '#6B0500';
+      $config['environment_indicator.indicator']['fg_color'] = '#fff';
+      $config['environment_indicator.indicator']['name'] = 'Development';
+      break;
+    case 'test':
+      $config['environment_indicator.indicator']['bg_color'] = '#4127C2';
+      $config['environment_indicator.indicator']['fg_color'] = '#fff';
+      $config['environment_indicator.indicator']['name'] = 'Staging';
+      break;
+    case 'prod':
+      $config['environment_indicator.indicator']['bg_color'] = '#000';
+      $config['environment_indicator.indicator']['fg_color'] = '#fff';
+      $config['environment_indicator.indicator']['name'] = 'Production';
+      break;
+    default:
+      $config['environment_indicator.indicator']['bg_color'] = '#086601';
+      $config['environment_indicator.indicator']['fg_color'] = '#fff';
+      $config['environment_indicator.indicator']['name'] = $_ENV['AH_SITE_ENVIRONMENT'];
+      break;
+  }
+}
+
+if (isset($_ENV['AH_SITE_ENVIRONMENT'])) {
+  if (PHP_SAPI !== 'cli') {
+    // Don't lock config when using drush.
+    $settings['config_readonly'] = TRUE;
+  }
+}
+
+// Lets whitelist everything because in our event subscriber we have the
+// ability to decide which forms are locked.
+// @see \Drupal\hs_config_readonly\EventSubscriber\ConfigReadOnlyEventSubscriber
+$settings['config_readonly_whitelist_patterns'] = ['*'];
+
+// On acquia, load a salt from the server.
+if ($is_ah_env) {
+  $settings['hash_salt'] = file_get_contents('/mnt/gfs/swshumsci.prod/nobackup/apikeys/salt.txt');
+}
diff --git a/docroot/themes/humsci/su_humsci_admin/su_humsci_admin.theme b/docroot/themes/humsci/su_humsci_admin/su_humsci_admin.theme
new file mode 100644
index 0000000..109dc11
--- /dev/null
+++ b/docroot/themes/humsci/su_humsci_admin/su_humsci_admin.theme
@@ -0,0 +1,29 @@
+<?php
+
+/**
+ * @file
+ * su_humsci_admin.theme
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function su_humsci_admin_form_user_login_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if (isset($form['simplesamlphp_auth_login_link'])) {
+    // Moves the original form elements into a collapsed group.
+
+    $form['simplesamlphp_auth_login_link']['#weight'] = -99;
+    $form['manual'] = [
+      '#type' => 'details',
+      '#title' => t('Manual Login'),
+      '#open' => FALSE,
+    ];
+    $form['manual']['name'] = $form['name'];
+    $form['manual']['pass'] = $form['pass'];
+    $form['manual']['actions'] = $form['actions'];
+
+    unset($form['name'], $form['pass'], $form['actions']);
+  }
+}
```


blt/src/Commands/HumsciCommand.php
```diff
@@ -321,7 +300,24 @@ protected function updateSite($multisite) {
...
+
+  /**
+   * Get encryption keys from acquia.
+   *
+   * @command humsci:keys:send
+   */
+  public function humsciKeysSend() {
+    $send = $this->askQuestion('Are you sure you want to copy over existing keys with keys in the "keys" directory? (Y/N)', 'N', TRUE);
+
+    if (strtolower($send)[0] == 'y') {
```
sherakama: strtolower on an array? This seems odd. Don't you want 
```
if (strtolower($send[0]) == 'y') {
```

blt/src/Commands/HumsciConfigCommand.php
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      switch ($strategy) {
+        case 'core-only':
+          $this->importCoreOnly($task, $cm_core_key);
+          break;
+
+        case 'config-split':
+          try {
+            $this->importConfigSplit($task, $cm_core_key, $options['partial']);
```
sherakama: All this to get `partial` in here?

blt/src/Commands/HumsciConfigCommand.php
```diff
@@ -0,0 +1,121 @@
...
+   */
+  protected function importConfigSplit($task, $cm_core_key, $partial = FALSE) {
+    $task->drush("pm-enable")->arg('config_split');
+
+    // Local environments we don't want all the custom site created configs.
+    if ($this->getConfigValue('environment') == 'local' && !$partial) {
+      $task->drush("config-import")->arg($cm_core_key);
+      // Runs a second import to ensure splits are
+      // both defined and imported.
+      $task->drush("config-import")->arg($cm_core_key);
```
sherakama: is this second import from BLT or something you worked through?

drush/sites/archaeology.site.yml
```diff
@@ -6,6 +6,11 @@ dev:
   uri: archaeology-dev.stanford.edu
   host: srv-7503.devcloud.hosting.acquia.com
   user: swshumsci.dev
+livedev:
```
sherakama: Do you need this? Doesn't drush hit the same db even if in livedev? I guess if you are creating a drush plugin this might be important.

docroot/themes/humsci/su_humsci_admin/su_humsci_admin.theme
```diff
@@ -0,0 +1,29 @@
...
+ * @file
+ * su_humsci_admin.theme
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function su_humsci_admin_form_user_login_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
sherakama: OH, can this be submitted to the 8.x version of stanford_ssp. This is the exact kind of thing we want in that project.

blt/src/Commands/HumsciConfigCommand.php
pookmish: It came from BLT. I think it's if there are new splits. Imports the split first, then ensures the split configs get in.

drush/sites/archaeology.site.yml
pookmish: I thought I needed it because of the settings that acquia was adjusting. Probably don't now that I think about it

docroot/sites/config.setttings.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+// SimpleSAMLphp configuration
+// Provide universal absolute path to the installation.
+if (isset($_ENV['AH_SITE_NAME']) && is_dir('/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp')) {
+  $settings['simplesamlphp_dir'] = '/var/www/html/' . $_ENV['AH_SITE_NAME'] . '/simplesamlphp';
+}
+else {
+  // Local SAML path.
+  if (is_dir(DRUPAL_ROOT . '/../simplesamlphp')) {
```
sherakama: Instead of hard wiring local config it is better to create a local.settings.php file that is not tracked in the repo instead. I'd remove the local config and cadd an include file statement.

blt/src/Commands/HumsciConfigCommand.php
pookmish: yeah because i can't override the importConfigSplit() method without first taking over the `drupal:config:import` command.

drush/sites/archaeology.site.yml
pookmish: it does kinda make `drush ssh` a little faster too

docroot/sites/config.setttings.php
pookmish: ooh i found this. handy little deal https://github.com/acquia/blt/blob/9.x/settings/blt.settings.php#L238

pookmish:
I moved some things around and made `default` directory the main source for future site installs. config splits will then create configs when they are different from the default. Not sure how this will play into the update process though. For a test with a new site i added a PR for an example scenario https://github.com/SU-HSDO/suhumsci/pull/38
using https://www.drupal.org/project/config_installer you can simply do a drush command. it'll import all the config and then you have a matching site to archaeology... other than the unique items.

blt/blt.yml
```diff
@@ -42,3 +42,6 @@ command-hooks:
   post-deploy-build:
     dir: '${deploy.dir}'
     command: 'chmod +x blt/cleanup.sh; blt/cleanup.sh'
+cm:
+  strategy: config-split
+  allow-overrides: true
```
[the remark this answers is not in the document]
pookmish: It came from this https://github.com/acquia/blt/blob/9.x/src/Robo/Commands/Setup/ConfigCommand.php#L198
and that method is called in https://github.com/SU-HSDO/suhumsci/blob/HSD8-147-config-split/blt/src/Commands/HumsciConfigCommand.php#L88

pookmish:
I also ran the config_installer on the `default` directory site. that can be viewed at https://swshumscidev.devcloud.acquia-sites.com/ (ignore the ssl errors, the cert wasn't set up for that domain)

blt/src/Commands/HumsciConfigCommand.php
sherakama: Yeah, Bummer.

drush/sites/archaeology.site.yml
sherakama: Nice to have. Just typing out loud.

blt/deploy/.gitignore
```diff
@@ -1,6 +1,5 @@
 db.sql
 keys
-deploy
```
sherakama: I guess you don't need to ignore the deploy folder in the deploy folder.

docroot/themes/humsci/su_humsci_admin/su_humsci_admin.theme
pookmish: i looked at stanford_ssp... were you trying to influence me to build the 8.x version? 🤔 haha

blt/src/Commands/HumsciCommand.php
```diff
@@ -321,7 +300,24 @@ protected function updateSite($multisite) {
    * @command humsci:keys
    */
   public function humsciKeys() {
-    $this->taskDrush()->drush("rsync @swshumsci.dev:/mnt/gfs/swshumsci.prod/nobackup/apikeys/ @self:../keys")->run();
+    $this->taskDrush()
+      ->drush("rsync @swshumsci.dev:/mnt/gfs/swshumsci.prod/nobackup/apikeys/ @self:../keys")
```
sherakama: The drush alias and the location would be best configured as variables from the BLT yml file. Does this only get run on Acquia's server? If no, I suppose we can ignore making these variables.

pookmish: it would be super quick to change these into blt configuration variables.

blt/deploy/.gitignore
pookmish: backstory: theres a `deploy` directory in blt and the deploy commands would fail because those files didnt exist.

docroot/sites/settings/common.settings.php
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Lets whitelist everything because in our event subscriber we have the
+// ability to decide which forms are locked.
+// @see \Drupal\hs_config_readonly\EventSubscriber\ConfigReadOnlyEventSubscriber
+$settings['config_readonly_whitelist_patterns'] = ['*'];
+
+// On acquia, load a salt from the server.
+if ($is_ah_env) {
+  $settings['hash_salt'] = file_get_contents('/mnt/gfs/swshumsci.prod/nobackup/apikeys/salt.txt');
```
sherakama: Everything using the same salt? I think this should be separated out into 3 environment files and use the environment variables available to generate the path.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -51,7 +54,45 @@ function su_humsci_profile_install() {
...
+ */
+function su_humsci_profile_update_8002() {
+  \Drupal::service('module_installer')->install([
+    'config_split',
+    'config_ignore',
+  ]);
+
+  // Install all the default configs first which includes environment specific
+  // config splits.
+  $source = new FileStorage('../config/default');
```
sherakama: This is handy. Setting up config splits is a pain. Perhaps a drush function one day.

docroot/sites/settings/common.settings.php
pookmish: I was thinking about that. I wasn't sure if it was really necessary. i just noticed that the salt was in github and wanted to fix that

config/envs/prod/shield.settings.yml
```diff
@@ -0,0 +1,14 @@
+credential_provider: shield
```
sherakama: Don't we want this disabled on prod? Why are we tracking config for it?

config/envs/dev/shield.settings.yml
```diff
@@ -0,0 +1,14 @@
+credential_provider: shield
+credentials:
+  shield:
```
sherakama: Shouldn't there be some sort of login? Why is the user/pass empty?

config/default/user.role.contributor.yml
```diff
@@ -0,0 +1,71 @@
+uuid: caa42067-aba2-4d13-84e5-5fb3e7e78d54
```
sherakama: Role config might have to be split out into each site as each site may have different modules installed, no? This config might be better set on install and then ignored.

config/default/system.site.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Archaeology Center'
```
sherakama: This one will definitely have to be split out to each site.

config/default/system.performance.yml
```diff
@@ -0,0 +1,17 @@
+cache:
+  page:
+    max_age: 0
```
sherakama: No caching by default? Perhaps yes.

config/default/dblog.settings.yml
```diff
@@ -0,0 +1,3 @@
+row_limit: 1000
```
sherakama: shouldn't this be disabled and use syslog instead?

config/default/system.performance.yml
pookmish: I think it gets overridden by settings files at runtime on acquia.

config/default/dblog.settings.yml
pookmish: with dblog disabled they don't get 404 reports. kinda dumb IMO

config/default/system.site.yml
pookmish: yeah. i'll set the `default` to something and add the override

config/default/user.role.contributor.yml
pookmish: yeah. each site will probably have its own config. but this file at least sets the initial perms even if we ignore it. i found that with config ignore, the config will get imported but then it wont recieve changes.

config/envs/prod/shield.settings.yml
pookmish: Oooh yeah definitely don't want it on prod. we aren't even using it at all, maybe we kill it off

sherakama:
@boznik your turn.

boznik:
FYI, i had to mod my drush/sites/archaeology.yml with my personal local uri, but that fixed the sync errors.

boznik:
why are the menu position rules blacklisted?

boznik:
Should we use the State API for storing dev, test, prod specific items?

pookmish:
Menu position rules are blacklisted because they rely on the UUID of the menu item which is content. I'm tempted to removed them entirely because they can cause issues if the menu item changes and the config is locked.

Whats the state API you are referring to?

boznik:
apparently it's a way to store env specific config. i have not used it. just mentioning in case it would be useful for the different env's on acquia:
https://www.drupal.org/docs/8/api/state-api/overview

boznik:
from slack:
greg: ok, i’m curious that config split dir business is in an update hook. how does that work for new sites? what would happen if it was defined when a new site is created by adding `$config_directories[CONFIG_SYNC_DIRECTORY]` to it’s settings.php file. or did you deal with this somewhere that i’m not seeing?

mike:
so `$config_directories[CONFIG_SYNC_DIRECTORY]` is set by blt.settings.php and will be set to `config/default` for every site. the individual site’s split will probably have to be manual process for now. i’m sure we could put in some type of install hook or something that creates it though

sherakama:
@boznik The state API is for transient information that does not get tracked or exported anywhere. Although the documentation mentions that it is for specific environments, it is not something that is reproducible. Once the state has been cleared it would have to be explicitly set the next time by a script or a manual drush call. 

Also, state API is not for storing configuration items such as to which modules are enabled. That is the responsibility of the configuration API.

sherakama:
+1 to 
> would like to find a way to programmatically add the new site config dirs.

pookmish:
config dirs get created by blt. its the creation of the site specific config during installation that might be a tricky task since we'll use the config_install profile. more on that when the time comes.

pookmish:
Ok here goes nothing 🤞

docroot/themes/humsci/su_humsci_admin/Gruntfile.js
```diff
@@ -44,7 +44,7 @@ module.exports = function(grunt) {
     uglify: {
       // This is for minifying all of our scripts.
       options: {
-        sourceMap: true,
+        sourceMap: false,
```
sherakama: @pookmish: Why no source map?

pookmish: it clogs up my dblogs with 404 for the .map files. does it not?

sherakama: Oh interesting. I have not seen that. @boznik have you seen this happening?

pookmish: screen shot 2018-07-11 at 8 56 24 am

PR #43, opened 2018-07-10
Arch 191 course content

pookmish:
# READY FOR REVIEW

# Summary
- Course content type and importer

# Needed By (Date)
- EOD wednesday if possible (end of sprint)

# Urgency
- low

# Steps to Test

1. checkout branch
2. `composer update --prefer-source`
3. `blt drupal:sync --site=archaeology`
4. go to `/admin/config/system/keys` and click “add Key”
5. in the “key value” field put “http://yourdomain.local/api/hs_courses”
6. go to `admin/config/development/configuration/key-overrides/add` and set up the configuration override like the screenshot here: https://d.pr/free/i/V9z8rh
7. `drush @archaeology.local cron`
8. validate course content got created. theres should be about 124 courses.
9. validate course data matches [explore courses](http://explorecourses.stanford.edu/search?q=ARCHLGY&view=xml-20140630&collapse=&filter-departmentcode-ARCHLGY=on&filter-coursestatus-Active=on&filter-catalognumber-ARCHLGY=on)
10. attempt to view a course when logged out. it should bounce you to the explore courses page.


blt/src/Commands/HumsciConfigCommand.php
```diff
@@ -118,4 +118,27 @@ protected function importConfigSplit($task, $cm_core_key, $partial = FALSE) {
     $task->drush("config-import")->arg($cm_core_key)->option('partial');
   }
 
+  /**
+   * Import any missing entity form/display configs since they are ignored.
+   *
+   * @hook post-command drupal:config:import
+   */
+  public function postConfigImport() {
```
sherakama: Are you sure you want to make this assumption and not explicitly add the missing imports to an installation profile or update hook somewhere else? This seems a bit dangerous and opaque. Open for discussion.

config/default/core.entity_view_display.course_collections.instructor.default.yml
```diff
@@ -0,0 +1,123 @@
+uuid: 65e44979-6f86-4e60-af11-88c7f03aabb4
```
sherakama: This config acts as the 'sensible default' and is allowed to be changed in the UI?

pookmish: not on ECK entities.

blt/src/Commands/HumsciConfigCommand.php
pookmish: i thought about an update hook and that would be fine. I thought using this would eliminate the need for an update hook. instead it just always runs. the drush command to import the missing actually looks at the modules and not the config directory. so if a config lives in a module i think we can assume it should be there.

sherakama: Fair nuff, but as this a custom process and practice this will need to be documented somewhere it can get under a developer's 👃 so they are not surprised by this in the future. Perhaps outputting a message or log to the console while this is happening to let the developer know would be enough?

config/default/core.entity_view_display.course_collections.instructor.default.yml
pookmish: I recently locked down form and video display configs for everything _except_ nodes. paragraphs and ECK will be locked

blt/src/Commands/HumsciConfigCommand.php
pookmish: I think github is delayed updating the diff, but i added a line at the top of the method:
`$this->yell('Importing missing configs!');` so you really know whats happeneing. lol

sherakama: WAT!?!? 
I couldn't hear you.

Ok, I'm going to be a bit picky. Instead of calling them missing explain how they became missing with a message like
'Importing new configuration items that don't exist in the database but are ignored in config.ignore'

Then later on do some more yelling and output the name of each. This should be helpful for debugging.

boznik:
had some errors and needed to run:
blt drupal:config:import --site=archaeology --partial

config/default/core.entity_view_display.course_collections.instructor.default.yml
sherakama: Ah cool. Good to know.

boznik:
yo @shea i tested this and it working as expected. if you greenlight it, i'm gonna merge.

sherakama:
small update to the yelling required before merging: https://github.com/SU-HSDO/suhumsci/pull/43#discussion_r203563644

blt/src/Commands/HumsciConfigCommand.php
sherakama: Ah ok. LG2M.

PR #45, opened 2018-07-10
Allow view duplication form

pookmish:
# READY FOR REVIEW

# Summary
- View duplication is locked right now. this opens that up and allows for additional form ids if we find any more.

# Needed By (Date)
- Thursday Morning

# Urgency
- High

# Steps to Test

1. Checkout this branch
2. sync to production
3. enable hs_config_readonly & clear caches
4. ensure you can duplicate a view

# See Also

### DIFF
--- a/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
+++ b/docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
@@ -49,9 +49,16 @@ class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
    */
   protected $readOnlyFormIds = [
     'config_single_import_form',
-    'system_modules',
     'system_modules_uninstall',
-    'user_admin_permissions',
+  ];
+
+  /**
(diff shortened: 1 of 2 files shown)


docroot/modules/humsci/hs_config_readonly/src/EventSubscriber/ConfigReadOnlyEventSubscriber.php
```diff
@@ -49,9 +49,16 @@ class ConfigReadOnlyEventSubscriber implements EventSubscriberInterface {
    */
   protected $readOnlyFormIds = [
     'config_single_import_form',
-    'system_modules',
```
sherakama: Based on your title it doesn't seem that these subtractions were intentional. Are you sure you want to remove these?

pookmish: yeah. they are going to want to enable modules and change permissions on each site. thanks for checking

pookmish: i purposefully left the module disable since it could cause issues with the configs if they try to re-enable

PR #54, opened 2018-07-23
ARCH-303: Removing inline image floats for mobile + other adjustments.

bixgomez:
# READY FOR REVIEW

# Summary
- Removes floats, and adding 100% width, to inline images at mobile breakpoints.
- Also adding appropriate margins for left and right floats.
- *NOTE:* I did this as a "Decanter Override" because I feel we might want to include it there.

# Steps to Test
1. Visit a page with inline images embedded in the WYSIWYG, such as /about/archaeology-collections
1. Add another image (float it to the left if the one there is floating right)
1. Evaluate the behavior of these images at all breakpoints.

- [JIRA ticket](https://stanfordits.atlassian.net/browse/ARCH-303)


docroot/themes/humsci/su_humsci_theme/scss/decanter-overrides/_decanter-overrides.scss
```diff
@@ -30,3 +30,36 @@
 h1:first-child {
   @include margin(0 null null);
 }
+
+// Embedded media in WYSIWYG.
+.embedded-entity {
```
pookmish: This is drupal specific and wont be put into decanter. we should move this to another partial

pookmish:
we should move this section out of decanter overrides. also please clear up the conflicts.

josephgknox:
Confirming that code has been moved out of Decanter overrides.

PR #59, opened 2018-07-30
ARCH-155 mission statement and emphasized text styles

josephgknox:
# READY FOR REVIEW

# Summary
- Added emphsized-text style to WYSIWYG editor and added related styles
- Added custom homepage styles for mission statement for Archaeology's homepage

# Needed By (Date)
- End of sprint (8/2)

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Add emphasized-text style to collection of text on homepage's mission statement paragraph
3. Ensure that the mission statement closely resembles the mockup at https://invis.io/P7HCI9HYNA4#/299456614_Homepage

- Archeaology
- suhumsci (theme)

- https://stanfordits.atlassian.net/browse/ARCH-155

# See Also


docroot/themes/humsci/su_humsci_theme/scss/custom/archaeology/archaeology.scss
```diff
@@ -93,3 +93,10 @@ h2.text-align-center:after {
 .front .hs-basic-page.field-hs-page-hero {
   @include margin(null null 0.5em);
 }
+
+.front .paragraph.hs-text-area:first-child p.text-align-center {
```
pookmish: This is probably gonna be orphaned real quick. as soon as they dont want it centered, or want it as a second paragraph, or any changes to it.

josephgknox: Any ideas on how to accomplish this in a different way? This is the design that they approved. I can create another class (one that is special for Arch that can be assigned to this paragraph) if you think that would be better?

pookmish: I think a reusable class would be much better here. even if it wasn't in the drop down of styles, it would still be able to be used on other paragraphs AND it would persist if they changed the order of the paragraphs or something like that.

josephgknox: Awesome! I'll get this up soon!

josephgknox: Instead, we can leverage an existing class (decanter-font-lead), and I can add to the decanter-overrides file for what changes we want.

pookmish: That class already lives in the style drop down and could be used in many places on the site already

josephgknox: Ok, well then so we don't write something to custom for the homepage, I will indeed create a new class.

josephgknox:
👍

PR #65, opened 2018-08-06
styles for all people fields

josephgknox:
# READY FOR REVIEW

# Summary
- Added necessary styles for all fields on people content type

# Needed By (Date)
- End of sprint (8/16)

# Urgency
- N/A

# Steps to Test

1. Pull this branch
2. Navigate to person node that is fully populate (even though this should never happen since there are fields for staff, faculty, students, etc.)
3. Ensure that styles and collection of items together look good

- Archaeology
- suhumsci (theme)

- https://stanfordits.atlassian.net/browse/ARCH-319
- Need to pull new display settings from https://swshumsci-stage.stanford.edu/admin/structure/types/manage/hs_person/display-layout/default

# See Also


docroot/themes/humsci/su_humsci_theme/scss/nodes/hs_person.scss
```diff
@@ -5,55 +5,41 @@
...
+    @include margin(null null 0.5rem 0);
   }
 
-  .divider {
-    margin: 0 5px;
-  }
-}
-
-.hs-person.full.node {
+  .field-hs-person-title h3 {
```
pookmish: I think this should actually be more universal on the layout. lets add this to https://github.com/SU-HSDO/suhumsci/blob/develop/docroot/modules/humsci/hs_layouts/scss/three-column-w-image.scss and target all headings in the two regions that wrap the image so we don't have to override them on each node

PR #76, opened 2018-08-15
HSD8-192 Dynamically add site specific stylesheet

pookmish:


docroot/themes/humsci/su_humsci_theme/theme-settings.php
```diff
@@ -0,0 +1,32 @@
...
+  $form['humsci_settings'] = [
+    '#type' => 'details',
+    '#title' => t('Humsci Settings'),
+    '#open' => TRUE,
+  ];
+  $form['humsci_settings']['humsci_site_styles'] = [
+    '#type' => 'select',
+    '#title' => t('Choose a site style'),
+    '#empty_option' => t('- Default -'),
+    '#default_value' => theme_get_setting('humsci_site_styles'),
```
sherakama: I've had some fun in the past where `theme_get_setting` wasn't getting the right setting due to theme overrides somewhere, like in subsite.
Since then I have been in the practice of passing in the theme name to that function
Not sure if that will still be a thing in D8

docroot/themes/humsci/su_humsci_theme/su_humsci_theme.theme
```diff
@@ -29,6 +29,11 @@ function su_humsci_theme_preprocess_page(&$vars) {
   if (!isset($vars['node']) && strpos($controller, 'layout_builder') === FALSE) {
     $vars['main_class'] = 'decanter-grid';
   }
+
+  if ($site_style = theme_get_setting('humsci_site_styles')) {
+    // No errors occur if the library doesn't exist, so this is safe to do.
+    $vars['#attached']['library'][] = "su_humsci_theme/$site_style";
```
sherakama: Would you ever load a library from somewhere else, like a module? Perhaps you want the $site_style key to be the whole name of the library rather than assume this theme.

josephgknox:
Woot!

PR #83, opened 2018-08-22
HSD8-89 Allowed second line for site name

pookmish:
# NOT READY FOR REVIEW

# Summary
- Added 2nd line site name option

# Needed By (Date)
- End of sprint - aug 30

# Urgency
- Low

# Steps to Test

1. checkout this branch
1. sync site `blt drupal:sync --site=archaeology --sync-files`
1. go to the basic site settings form: `/admin/config/system/site-information`
1. populate the "Site name line 2" text field with something of your choice
1. go to the homepage and validate your text appears in the header under the site name.

# See Also


pookmish:
@josephgknox wanna review the scss change in this PR please

josephgknox:
@pookmish SCSS looks good! As we chatted about via Slack: we'll want to recompile so that we aren't inheriting new `stanford_basic` (Decanter).

pookmish:
@josephgknox I dont think theres anything to recompile. package.json is set to decanter ^3.0.0 and is independent of stanford_basic. However, i have set stanford_basic to the correct commit

docroot/modules/humsci/hs_theme_helper/hs_theme_helper.module
```diff
@@ -0,0 +1,77 @@
+<?php
```
sherakama: This second line functionality is something we have seen in past themes. Wouldn't it make sense for this to be in stanford_basic?

pookmish: I thought about that but you can't alter the form from the theme.

sherakama: Instead of altering the form, what about creating your own settings much like how OFW did it. 

You override the title and add a second line if available from the theme settings.

pookmish: Then would we have to move all of the basic site settings into the theme form? such as the home page, 404, 403 and slogan?

sherakama: No, just the titles. You have access to the templates and can apply the correct logic. 

You can swap out `{{ site_name }}` for the theme setting value and add `{{ site_name_line_2 }}` in the twig template.

pookmish: But then they would have 2 forms to go to. basic site settings for 404, home etc and theme settings for site title.

boznik: hmm, i need to think about this. it seems like it should go with the theme but i agree, two separate forms doesn't make sense.

pookmish: Moved possible work to https://github.com/SU-SWS/stanford_basic/pull/74

pookmish:
This all probably got moved to https://github.com/SU-SWS/stanford_basic/pull/74

pookmish:
Moved this to the stanford_basic

PR #124, opened 2018-10-16
simplified menu block styles

pookmish:


docroot/themes/humsci/su_humsci_theme/scss/components/molecules/_hover-menu.scss
```diff
@@ -324,93 +320,32 @@
...
-.main-container {
-  .menu-block > ul {
-    font-size: $base-font-size;
+nav.menu-block {
+  font-size: $base-font-size;
 
-    ul {
-      font-size: .9em;
-    }
+  ul {
```
josephgknox: This should be nested one level deeper so that this rule does not override the one right above it. Right now, the top menu item in the subnav is too small as it is getting the`.9em` style applied rather than the `$base-font-size`.

josephgknox: Things also start to get too small when you hit the fourth level (deepest we allow as of now).

josephgknox: Suggestion:

`nav.menu-block {
  font-size: $base-font-size;

  ul ul {
    font-size: .9em;
  }`

PR #125, opened 2018-10-18
Hsd8 426 capx

pookmish:
# READY FOR REVIEW

# Summary
- CAPx Importer

# Needed By (Date)
- 10/24

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. `blt drupal:sync --site=archaeology`
1. navigate to `/admin/structure/migrate/manage/hs_capx/migrations`
1. ensure no errors and there are 2 importers
1. navigate to `/admin/structure/migrate/capx/creds` input credentials (You probably need a new key)
1. navigate to `/admin/structure/migrate/capx` and choose a workgroup and/or organization and save
1. run cron
1. verify new content has been created and some fields are populated (if data is available) like:
    * first name
    * last name
    * bio
    * profile Picture
    * email address
    * Telephone number
    * Title

# See Also


config/default/migrate_plus.migration.hs_capx_images.yml
```diff
@@ -0,0 +1,88 @@
...
+      label: 'CAP API Etag'
+      selector: meta/etag
+  ids:
+    sunetid:
+      type: string
+process:
+  name: display_name
+  status: constants/status
+  bundle: constants/type
+  image_timestamp:
```
sherakama, line 49: There is a `LastModified` key on the response for each image. Why not use that?

docroot/modules/humsci/hs_capx/config/install/migrate_plus.migration.hs_capx.yml
```diff
@@ -0,0 +1,106 @@
...
+    -
+      name: phone
+      label: 'Phone Numbers'
+      selector: primaryContact/phoneNumbers
+    -
+      name: office
+      label: 'Office Locations'
+      selector: contacts
+    -
+      name: etag
```
sherakama, line 76: 🤔You're saving this right to the person entity? In the D7 version, we created a meta object to store all the extra information needed to handle efficient transactions without getting to the UI. It would be nice if etag checking was seamless and 'just happened' for anything that used the CAP API. This is perhaps out of scope for this project and may be best for another one. Food for thought at least.

docroot/modules/humsci/hs_capx/src/Capx.php
```diff
@@ -0,0 +1,230 @@
...
+   *   Comma separated organization codes.
+   * @param bool $children
+   *   Include all children of the organizations.
+   *
+   * @return string
+   *   CAPx URL.
+   */
+  public static function getOrganizationUrl($organizations, $children = FALSE) {
+    $organizations = preg_replace('/[^A-Z,]/', '', strtoupper($organizations));
+    $url = self::CAP_URL . "?orgCodes=$organizations&ps=1000";
```
sherakama, line 125: Is 1000 always going to be enough? How many are there currently?

docroot/modules/humsci/hs_capx/src/Capx.php
```diff
@@ -0,0 +1,230 @@
...
+
+    try {
+      $response = $this->client->request('GET', self::AUTH_URL, [
+        'query' => $parameters,
+        'auth' => [$this->username, $this->password],
+      ]);
+    }
+    catch (GuzzleException $e) {
+      // Most errors originate from the API itself.
+      return FALSE;
```
sherakama, line 164: Should log em. Debugging these things is hard.

docroot/modules/humsci/hs_capx/src/Capx.php
```diff
@@ -0,0 +1,230 @@
...
+    return $response->getStatusCode() == 200;
+  }
+
+  /**
+   * Sync the organization database with the api data from CAP.
+   *
+   * @throws \GuzzleHttp\Exception\GuzzleException
+   */
+  public function syncOrganizations() {
+    $this->insertOrgData($this->getOrgData());
```
sherakama, line 178: Any reason to return a status here?

config/default/migrate_plus.migration.hs_capx_images.yml
pookmish: Didn't we have trouble with that when we were testing it out. And that's why they added the timestamp to the url.?

docroot/modules/humsci/hs_capx/src/Capx.php
pookmish: No idea. I took that bit from earth urls

docroot/modules/humsci/hs_capx/config/install/migrate_plus.migration.hs_capx.yml
pookmish: :shallow_pan_of_food: :thought_balloon:

docroot/modules/humsci/hs_capx/src/Capx.php
pookmish: I couldn't find a use case for returning a result. maybe in the future though? I could set it to return an array of all the organization data.

docroot/modules/humsci/hs_capx/src/Capx.php
pookmish: Logging added

config/default/migrate_plus.migration.hs_capx_images.yml
sherakama: That rings a bell. Where did that conversation happen? Slack, github?

config/default/migrate_plus.migration.hs_capx.yml
```diff
@@ -0,0 +1,106 @@
...
+    sunetid:
+      type: string
+process:
+  title: display_name
+  status: constants/status
+  type: constants/type
+  field_hs_person_first_name: first_name
+  field_hs_person_last_name: last_name
+  field_hs_person_title: long_title
+  body/value: biosketch
```
sherakama: biosketch is pulling in the text version of the json response. There is an HTML option available. With the formatting on this field are you sure you want the text version and not the html version?

config/default/migrate_plus.migration.hs_capx_images.yml
```diff
@@ -0,0 +1,88 @@
...
+      regex: true
+      search: '/^(?!ts=).+/'
+      replace: ''
+    -
+      plugin: str_replace
+      search: ts=
+      replace: ''
+    -
+      plugin: concat
+  image_path:
```
sherakama, line 68: As these are now entities in Drupal, are we possibly creating orphaned entities when the the image and timestamp changes and updates the parent node without revisions? Does the previous image get cleaned up or removed?

docroot/modules/humsci/hs_bugherd/config/install/rest.resource.hs_bugherd_resource_jira.yml
```diff
@@ -1,4 +1,4 @@
-uuid: eddf0d9a-b688-4bf5-a170-ce0b622856c3
+uuid: e0b5cb53-6277-4923-b49e-8094974352e9
```
sherakama: Do you want a UUID in the module?

docroot/modules/humsci/hs_capx/config/install/migrate_plus.migration.hs_capx.yml
```diff
@@ -0,0 +1,106 @@
+uuid: 18fe55a6-a1f6-47c9-b079-992ef831c99b
```
sherakama: UUIDs generally don't go in modules.

docroot/modules/humsci/hs_capx/src/Capx.php
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+   * Authentication url.
+   */
+  const AUTH_URL = 'https://authz.stanford.edu/oauth/token';
+
+  /**
+   * The actual CAP API.
+   */
+  const CAP_URL = 'https://cap.stanford.edu/cap-api/api/profiles/v1';
```
sherakama, line 29: Do we need to plan for the possibility of a v2 endpoint or would we just hardwire the new API endpoint version to the module when we are ready? Do we expect older versions of this module to work with a new API schema? Probably not, so this is likely ok.

docroot/modules/humsci/hs_capx/src/Capx.php
```diff
@@ -0,0 +1,262 @@
...
+    if ($cache = $this->cache->get('capx:org_data')) {
+      return $cache->data;
+    }
+
+    $token = $this->getAccessToken();
+
+    // AA00 is the root level of all Stanford.
+    $result = $this->client->request('GET', self::API_URL . '/cap/v1/orgs/AA00', ['query' => ['access_token' => $token]]);
+    $result = json_decode($result->getBody()->getContents(), TRUE);
+    $this->cache->set('capx:org_data', $result, time() + 60 * 80 * 24 * 7, [
```
sherakama: 60 * `80` * 24 * 7 ?

docroot/modules/humsci/hs_capx/src/Controller/CapxOrgAutocomplete.php
```diff
@@ -0,0 +1,70 @@
...
+   * Get organizations that match the autocomplete text entry.
+   *
+   * @param \Symfony\Component\HttpFoundation\Request $request
+   *   Ajax Reqeust.
+   *
+   * @return \Symfony\Component\HttpFoundation\JsonResponse
+   *   Found results response.
+   */
+  public function autocomplete(Request $request) {
+    $string = mb_strtolower($request->query->get('q'));
```
sherakama: These are safe and sanitized?

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
```diff
@@ -0,0 +1,130 @@
...
+    $config = $this->config('hs_capx.settings');
+    $form['username'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Username'),
+      '#required' => TRUE,
+      '#default_value' => $config->get('username'),
+    ];
+
+    $keys = Key::loadMultiple();
+    foreach ($keys as &$key) {
```
sherakama, line 77: What? Why are you looping through an array and pulling out the label? If you want the last one can't you just pop it from the array?

sherakama: Also the $key variable doesn't appear to be used.

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
```diff
@@ -0,0 +1,130 @@
...
+    parent::validateForm($form, $form_state);
+
+    // Check the given username and password will authenticate with CAPx.
+    $this->capx->setUsername($form_state->getValue('username'));
+    $password_key = Key::load($form_state->getValue('password'));
+    $this->capx->setPassword($password_key->getKeyValue());
+
+    Cache::invalidateTags(['capx']);
+
+    if (!$this->capx->testConnection()) {
```
sherakama, line 109: Nice, thank you for this form validation.

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
```diff
@@ -0,0 +1,130 @@
...
+      $form_state->setError($form['password'], $this->t('Invalid Credentials'));
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    parent::submitForm($form, $form_state);
+    $this->configFactory->getEditable('hs_capx.settings')
```
sherakama, line 120: Plain text? No encryption?

docroot/modules/humsci/hs_capx/src/Overrides/ConfigOverrides.php
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   *
+   * Override the path to the key for the encryption profile.
+   */
+  public function loadOverrides($names) {
+    $overrides = [];
+    if (in_array('migrate_plus.migration.hs_capx', $names) || in_array('migrate_plus.migration.hs_capx_images', $names)) {
```
sherakama: Instead of wrapping 99% of the body of the function in an if statement you could simply return the empty array here.

docroot/modules/humsci/hs_capx/src/Overrides/ConfigOverrides.php
```diff
@@ -0,0 +1,112 @@
...
+
+      $config = $this->configFactory->get('hs_capx.settings');
+      $password = '';
+      if ($key = Key::load($config->get('password') ?: '')) {
+        $password = $key->getKeyValue();
```
sherakama: Ah, here is the key usage. Thank you.

docroot/modules/humsci/hs_capx/tests/Unit/orgs.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "~0N`S$Z4",
```
sherakama, line 2: How'd you come up with this wonderful test data 😆

config/default/migrate_plus.migration.hs_capx_images.yml
pookmish: it was in slack. https://stanfordwebservices.slack.com/archives/D2YMF9509/p1509739760000136
> Mike: The interesting thing now, we don’t even have to apply those changes in that PR. the image urls now contain a timestamp 
`https://cap.stanford.edu/profiles/viewImage?profileId=162949&type=big&ts=1509739817173`

>Shea McKinney [1:25 PM]
>Huh interesting. The path to the image finally changes
So yeah, we don't need that timestamp check
anyway

config/default/migrate_plus.migration.hs_capx.yml
pookmish: ahh. i didnt realize that. good catch

config/default/migrate_plus.migration.hs_capx_images.yml
pookmish: If i'm not mistaken, the image on the media entity will get replaced. the old image will be cleaned up during garbage collection and the media entity will keep the same ID as the target from the person node.

docroot/modules/humsci/hs_bugherd/config/install/rest.resource.hs_bugherd_resource_jira.yml
pookmish: :man_shrugging: i'm gone back and forth on it. since its the product i wasnt sure if it was a big concern.

docroot/modules/humsci/hs_capx/src/Capx.php
pookmish: i'd guess that a V2 would require refactoring of the api calls so i think this should be safe.

docroot/modules/humsci/hs_capx/src/Capx.php
pookmish: :man_facepalming: it gives extra 2.3 days of cache haha.

docroot/modules/humsci/hs_capx/src/Controller/CapxOrgAutocomplete.php
pookmish: All the guides i found showed to do that. I couldn't find an example that sanitizes https://www.qed42.com/blog/autocomplete-drupal-8
Also core uses the same thing https://api.drupal.org/api/drupal/core%21modules%21system%21src%21Controller%21EntityAutocompleteController.php/function/EntityAutocompleteController%3A%3AhandleAutocomplete/8.2.x

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
pookmish: `$keys` is used as the options in the form, so what this is doing is changing the `$key` object into a string since the foreach is looping with reference.

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
pookmish: The key would be encrypted. this just stores the machine name of the key, so this should be safe. unless we think the username should also be encrypted

docroot/modules/humsci/hs_capx/tests/Unit/orgs.json
pookmish: It came to me in a dream :1st_place_medal:

pookmish:
Any idea why the composer changes that have been made would cause all the circleci to fail?

docroot/modules/humsci/hs_capx/src/Controller/CapxOrgAutocomplete.php
sherakama: The Request object is from Symfony and I couldn't find anything that would suggest one way or the other. It looks like q gets put in to a generic parameter bag but after that I lost track of what happens to the values. Thanks for the examples, in the core autocomplete function the query gets broken up with `Tags::explode` and that would effectively wreck any javascript or code injection. I think we may want to filter for XSS and ensure plain text on this value just to be safe.

pookmish:
yay. tests passed

docroot/modules/humsci/hs_bugherd/config/install/rest.resource.hs_bugherd_resource_jira.yml
sherakama: Future Mike will be thankful for having these removed. I'm 95% sure removing the UUID is the standard practice.

pookmish: past mike created an bash alias to recursively remove `uuid` from yml files :D

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
sherakama: Ah, I missed that detail. Please add a code comment.

docroot/modules/humsci/hs_bugherd/config/install/rest.resource.hs_bugherd_resource_jira.yml
sherakama: Nice, past Mike is not always a jerk eh?

docroot/modules/humsci/hs_capx/src/Form/CapxCredsForm.php
pookmish: Comment added

docroot/modules/humsci/hs_bugherd/config/install/rest.resource.hs_bugherd_resource_jira.yml
pookmish: he's got his moments

boznik (approved):
Cron had some errors, but were due to course importer. working as expected

PR #130, opened 2018-10-31
Hsd8 432 layout groups

pookmish:
# READY FOR REVIEW

# Summary
- Added a block that allows for other blocks to be nested within it on the layout builder

# Needed By (Date)
- 11/7

# Urgency
- Medium

# Steps to Test
1. checkout this branch
1.  sync a site `blt drupal:sync --site=archaeology`
1. edit the layout of a node or a content type
1. click "Add Block"
1. find the "Group Block: Content" under the "content" section
1. give it a name and a machine name (note there are no visible form errors at this time so the machine name has to be correct)
1. validate your group was added to the layout
1. click the button "Add Block to group"
1. choose a field or more on the content type or node
1. configure as desired
1. after saving, verify it looks as expected
1. verify if no contents are in those fields, the entire group is not visible.

# See Also


boznik (approved):
This is working as expected

PR #134, opened 2018-11-07
HSD8-445 Added alter to conslidated date field sorting

pookmish:
# READY FOR REVIEW

# Summary
- Allow for multiple date fields to be used as a single sort.

# Needed By (Date)
- Soon

# Urgency
- low

# Steps to Test

1. checkout this branch
1. clear cache
1. create some events and news items (keep track of what dates are on them) alternating dates between the two content types.
1. create a new view
1. add some fields to the view and add the two date fields as sorting
1. validate the view sorts the items togeather.
1. optionally add another sort to the bottom or the top of the date fields, validate it sorts correctly still.

# See Also


docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+namespace Drupal\hs_field_helpers\Plugin\views\query;
+
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\views\Plugin\views\query\Sql as OriginalSql;
+use Drupal\views\ViewExecutable;
+
+/**
+ * Class Sql that adds extra functionality to views Sql.
```
sherakama: What kind of extra functionality?

pookmish: the secretive kind.

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
```diff
@@ -0,0 +1,122 @@
...
+
+    $expression = [];
+    $orderby_delta = -1;
+
+    // Build the expression.
+    foreach ($date_fields as $field_alias => $direction) {
+      $field = $this->fields[$field_alias];
+
+      // Set null values to 0.
+      $expression[] = 'COALESCE(' . $field['table'] . '.' . $field['field'] . ', 0)';
```
sherakama, line 66: The `COALESCE` function seems to be valid for PGSQL, MySQL, and MSSQL. Not sure if there are slight behavior differences.

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
```diff
@@ -0,0 +1,122 @@
...
+   */
+  protected function getDateFields() {
+    $field_types = ['datetime', 'daterange'];
+    $date_fields = [];
+
+    foreach ($this->orderby as $order_item) {
+      $field_alias = $order_item['field'];
+      $field_data = $this->fields[$field_alias];
+      $table = $field_data['table'];
+      if (strpos($table, '__') === FALSE) {
```
sherakama, line 112: Why this check?

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
```diff
@@ -0,0 +1,122 @@
...
+        continue;
+      }
+      list($entity_type, $field_name) = explode('__', $table);
```
sherakama, line 116: http://php.net/manual/en/function.list.php

The list function's behavior is different in PHP5 vs PHP7. Is there a way you can ensure that this module is dependant on PHP7 or protect against running backward on PHP5.6

pookmish: I read that warning and i dont believe it applies to this code. I tried out different versions using http://sandbox.onlinephpfunctions.com/
php 5 and php 7 both produce the same result when exploding a string.

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
pookmish: comment added. it basically forces us to only operate on CCK fields. Base fields like node title and items arent applicable.

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
pookmish: just noting?

docroot/modules/humsci/hs_field_helpers/src/Plugin/views/query/Sql.php
pookmish: Greatest and least expressions

boznik (approved):
Working as expected

PR #141, opened 2018-11-29
Hsd8 497 revision cleanup

pookmish:
# READY FOR REVIEW

# Summary
- Cron task to delete old revisions but keep a configured number.

# Needed By (Date)
- 12/5

# Urgency
- low

# Steps to Test

1. checkout this branch
1. sync to a site `blt drupal:sync --site=[sitenmae]`
1. review a node to find one that has more than 5 revisions
1. run cron
1. review that node again, verify only 5 revisions exist

# See Also


config/default/hs_revision_cleanup.settings.yml
```diff
@@ -0,0 +1,4 @@
+cleanup:
```
sherakama, line 1: No GUUID?

docroot/modules/humsci/hs_revision_cleanup/hs_revision_cleanup.module
```diff
@@ -0,0 +1,31 @@
...
+
+    default:
+  }
+}
+
+/**
+ * Implements hook_cron().
+ */
+function hs_revision_cleanup_cron() {
+  \Drupal::service('hs_revision_cleanup')->deleteRevisions();
```
sherakama, line 30: How heavy is this? Do we want to run every cron?

docroot/modules/humsci/hs_revision_cleanup/src/RevisionCleanup.php
```diff
@@ -0,0 +1,174 @@
...
+        '@id' => $entity_id,
+      ]);
+
+      foreach (array_keys($revisions) as $revision_id) {
+        try {
+          $this->entityTypeManager->getStorage($entity_type)
+            ->deleteRevision($revision_id);
+        }
+        catch (\Exception $e) {
+          $this->logger->error('Unable to delete revision @rid. Error: @message', [
```
sherakama, line 122: Thank you for logging.

docroot/modules/humsci/hs_revision_cleanup/src/RevisionCleanup.php
```diff
@@ -0,0 +1,174 @@
...
+    $this->config = $config_factory->get('hs_revision_cleanup.settings');
+  }
+
+  /**
+   * Delete all revisions for the configured entity types.
+   */
+  public function deleteRevisions() {
+    foreach ($this->config->get('cleanup') as $cleanup_entity) {
+      try {
+        // Subtract 1 from keep config to account for the current revision.
```
sherakama:
```suggestion
        // Subtract 1 from keep config to account for the current revision because we filter out the current revision on the possible revisions query.
```

sherakama (approved):
This is good to merge. A couple of questions to be asked. The lack of a `GUUID` in the default config may lead to import GUUID mismatch issues no?

config/default/hs_revision_cleanup.settings.yml
pookmish: I've noticed that simple settings config dont get the uuid like config entities. i'm not sure why exactly that is. This config file was generated during module install and then exported. looking through the directory most simple configs dont have uuid's

docroot/modules/humsci/hs_revision_cleanup/hs_revision_cleanup.module
pookmish: The first run is the heaviest, but every following run is mostly a single sql query. it only loads the revision entity during deletion. so i'd expect most cron runs it never loads any revisions.

config/default/hs_revision_cleanup.settings.yml
sherakama: Huh, I have not noticed that. I can confirm that this is also the case with Earth. The one thing that earth does have is a `_default_config_hash` entry. Not entirely sure what that is for.


https://github.com/stanford-earth/SE3_BLT/blob/develop/config/default/action.settings.yml

PR #142, opened 2018-11-30
Hsd8 268 run importer ui

pookmish:
# READY FOR REVIEW

# Summary
- Provide a simple path like D7 to allow users to run importers
- Granular permissions.

# Needed By (Date)
- 12/5

# Urgency
- medium

# Steps to Test

1. checkout this branch
1. sync a site `blt drupal:sync --site=[sitename] --partial`
1. go to /import page
1. review the list of migration importers
1. run one of the importers
1. verify no errors occur.

# See Also


config/default/migrate_plus.migration.hs_courses.yml
```diff
@@ -18,7 +18,7 @@ source:
   track_changes: true
   plugin: url
   data_fetcher_plugin: http
-  urls: /api/hs_courses
```
sherakama, line 21: So we are switching away from this approach? I forgot why we need to pull from the local endpoint in the first place.

docroot/modules/humsci/hs_migrate/hs_migrate.module
```diff
@@ -0,0 +1,155 @@
...
+
+  // Do not return any migrations which fail to meet requirements.
+  /** @var \Drupal\migrate\Plugin\Migration $migration */
+  foreach ($matched_migrations as $id => $migration) {
+    if ($migration->getSourcePlugin() instanceof RequirementsInterface) {
+      try {
+        $migration->getSourcePlugin()->checkRequirements();
+      }
+      catch (RequirementsException $e) {
+        unset($matched_migrations[$id]);
```
sherakama, line 147: Should we have logging here for debug?

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
```diff
@@ -0,0 +1,220 @@
...
+
+    $this->migrations = $this->migrationManager->createInstances([]);
+
+    // No need to show migrations that are dependencies. They will get executed
+    // when their dependent migration is executed.
+    foreach ($this->migrations as $migration) {
+      foreach ($migration->getMigrationDependencies()['required'] as $dependency) {
+        // Set the dependency migrations to idle so that the primary migration
+        // will still execute.
+        $this->migrations[$dependency]->interruptMigration(MigrationInterface::RESULT_STOPPED);
```
sherakama: Is this for when migrations are running and you are in the UI? Do we really want to halt them? What would the outcome be if a migration was halted mid-run and the user did not re-run it through the form?

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  /**
+   * Build the table header labels.
+   *
+   * @return array
+   *   Array of table headers.
+   */
+  protected function buildHeader() {
+    return [
```
sherakama, line 132: This is nice. Good amount of info.

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
```diff
@@ -0,0 +1,220 @@
...
+   * @return array
+   *   Form render array.
+   */
+  protected function buildRow(MigrationInterface $migration) {
+    $row['label']['#markup'] = $migration->label();
+    $row['status']['#markup'] = $migration->getStatusLabel();
+    $row['imported']['#markup'] = $migration->getIdMap()->importedCount();
+
+    if ($last_imported = $this->lastMigrations->get($migration->id(), FALSE)) {
+      $row['last_imported']['#markup'] = $this->dateFormatter->format($last_imported / 1000, 'custom', 'M j Y g:i a');
```
sherakama, line 156: What format is $last_imported in and why `/1000`?

docroot/modules/humsci/hs_migrate/src/HsMigrateBatchExecutable.php
```diff
@@ -0,0 +1,132 @@
...
+      ->createInstance($migration_id);
+    $executable = new HsMigrateBatchExecutable($migration, $message, $options);
+
+    if (empty($context['sandbox']['total'])) {
+      $context['sandbox']['total'] = $executable->getSource()->count();
+
+      // THIS is the only change from the parent class. Allow 15 items to be
+      // imported on each batch execution. The parent split the total items
+      // into 100 executions which doesn't really do anything helpful.
+      $context['sandbox']['batch_limit'] = 15;
```
sherakama: Can this be a configuration item, please? Seems like this could/would benefit from having a flexible batch limit if we ever move to different hardware.

docroot/modules/humsci/hs_migrate/src/HsMigratePermissions.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  /**
+   * Build a list of permissions for the available migrations.
+   *
+   * @return array
+   *   Keyed array of permission data.
+   */
+  public function permissions() {
```
sherakama, line 66: How granular of you. Thanks!

sherakama (changes requested):
One stickler question about halting imports when navigating to the form to be addressed.

config/default/migrate_plus.migration.hs_courses.yml
pookmish: relative urls never worked. this is mostly for new sites that haven't configured their courses source or local development. its just putting in a url with an empty xml feed to prevent log errors. The reason we are pulling from local is to modify the courses xml data so that we can create unique guids on each course section and when there are no sections

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
pookmish: yeah this is via the UI. it forces the dependent migrations run be executed. if a migration is interrupted during import, it require manual manipulation to reset it's status so that it can run again. It shouldn't be an issue here because it'll run through all the items again, skipping those that have already been imported.

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
pookmish: microtime. this is an excerpt from migrate_tools module
```
    $migrate_last_imported_store = \Drupal::keyValue('migrate_last_imported');
    $migrate_last_imported_store->set($event->getMigration()->id(), round(microtime(TRUE) * 1000));
```

docroot/modules/humsci/hs_migrate/src/HsMigrateBatchExecutable.php
pookmish: yeah. it'd be nice if migrate plus module would do that. i think they were lazy with it.

docroot/modules/humsci/hs_migrate/hs_migrate.module
pookmish: Added logging

docroot/modules/humsci/hs_migrate/src/Form/HsMigrateImportForm.php
sherakama: My question is do you really want to do this by simply visiting the form? Wouldn't you want this to be a deliberate action?

pookmish: ahh. i hear ya there. :+1: yeah i think we could do this part during the actual execution.

pookmish: I moved the migration reset into the batch preparation. so it will only reset the necessary migrations as well.

docroot/modules/humsci/hs_migrate/src/HsMigrateBatchExecutable.php
```diff
@@ -55,6 +56,25 @@ protected function batchOperations(array $migrations, $operation, array $options
...
 
+  /**
+   * Reset status of the migration and its dependency migration.
+   *
+   * @param \Drupal\migrate\Plugin\MigrationInterface $migration
+   *   Migration object to reset.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   */
+  protected function prepareMigrations(MigrationInterface $migration) {
```
sherakama, line 67: I like this much better. Thank you.

sherakama (approved):

PR #146, opened 2018-12-10
automatically set importer permissions when configured

pookmish:
# READY FOR REVIEW

# Summary
- Automatically allow site managers to run importers _after_ importer has been configured.

# Needed By (Date)
- asap

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. composer install
1. blt drupal:sync
1. go to permission page, verify site managers don't have permission to "Execute Migration Course importer" or "Execute Migration Events importer"
1. configure an events or course migration at `/admin/structure/migrate/course-importer` or `/admin/structure/migrate/events`
1. go back to permission page, verify site managers now have permission to "Execute Migration Course importer" or "Execute Migration Events importer".

# See Also


docroot/modules/humsci/hs_capx/src/Form/CapxForm.php
```diff
@@ -73,6 +101,13 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       ->save();
 
     Cache::invalidateTags(['migration_plugins']);
+
+    // Add permission to execute importer.
+    $role = $this->entityTypeManager->getStorage('user_role')->load('site_manager');
+    if ($role) {
+      $role->grantPermission('import hs_capx migration');
+      $role->save();
```
sherakama, line 109: Are you tracking role permissions anywhere in config?

sherakama:
This seems fine with just a couple of questions.

1. Are you tracking the role permission grants anywhere that would be overridden by this new feature?
2. The granting process seems to be very granular. If the approach is to catch-all should this be done in a form_alter instead of adding the grant in each form submit?

pookmish:
We're tracking permissions only for install state. After installation, permissions are ignored.
It is granular because each form sets a different permission relative to the migration entity it is used on. A form alter would be more complex in the logic and wouldn't be OOP IMO

sherakama (approved):

PR #148, opened 2018-12-14
Hsd8 529 audio embeding

pookmish:
Follow steps in https://github.com/SU-SWS/stanford_media/pull/11


boznik (approved):
This is working as expected

PR #151, opened 2019-01-07
HSD8-463 Decode local urls before saving

pookmish:
# READY FOR REVIEW

# Summary
- When a user saves a link field with a local link, the saved value must be in a decoded format.
- Decode any local urls during saving

# Needed By (Date)
- asap

# Urgency
- low

# Steps to Test

1. checkout this branch
1. optionally sync a site
1. clear cache
1. create a new basic page with a postcard paragraph
1. for the link field enter any url with a `%20`
1. after saving, check that the generated link does not have `%25%20`

# See Also


boznik (approved):
This is working as expected

PR #154, opened 2019-01-15
Hsd8 530 capx tagging

pookmish:
# READY FOR REVIEW

# Summary
- Added new config entity for capx to allow for tagging content during importing.

# Needed By (Date)
- End of sprint 1/30

# Urgency
- High

# Steps to Test
1. Checkout this branch
1. composer install/update
1. blt drupal:sync --site=[sitename]
1. Add a capx password key on `/admin/config/system/keys`
1. configure capx credentials on `/admin/structure/migrate/capx/creds`
1. add 2 or more capx importers on `/admin/structure/migrate/capx` with some field tagging
1. run the importer on `/import` page. 
1. insure the content that was imported is tagged with the appropriate terms.

# See Also


sherakama:
Your summary is really short for how big a PR and deal this is. Digesting this today.

pookmish:
it ties into that conversation we had before winter closure. tag content during import based on what workgroup is configured to import

sherakama:
I like this. The interface feels familiar and consistent what we had in D7.

Couple of things:
1. Can you add description fields to the tagging to explain what it is doing
2. It would be nice if you could add an `execute` or `update` option in the operations drop down for each importer so someone could run it from that interface as well
3. Unrelated, but I cannot search for organization by the org code.

docroot/modules/humsci/hs_capx/src/CapxImporterListBuilder.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Drupal\hs_capx;
+
+use Drupal\Core\Config\Entity\ConfigEntityListBuilder;
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Provides a listing of Capx importer entities.
+ */
+class CapxImporterListBuilder extends ConfigEntityListBuilder {
```
sherakama, line 11: TIL, thanks.

docroot/modules/humsci/hs_capx/src/Entity/CapxImporter.php
```diff
@@ -0,0 +1,134 @@
...
+ *       "add" = "Drupal\hs_capx\Form\CapxImporterForm",
+ *       "edit" = "Drupal\hs_capx\Form\CapxImporterForm",
+ *       "delete" = "Drupal\hs_capx\Form\CapxImporterDeleteForm"
+ *     },
+ *     "route_provider" = {
+ *       "html" = "Drupal\hs_capx\CapxImporterHtmlRouteProvider",
+ *     },
+ *   },
+ *   config_prefix = "capx_importer",
+ *   admin_permission = "administer site configuration",
```
sherakama: It would be nice if this was `administer capx importers`

sherakama (approved):
GTG with a few feature additions.

pookmish:
So I tried and tried, but was not able to get a import button for a single importer. the config overridder gets cached and can't be changed at runtime depending on a route. which means i cant change the urls to target only a single url. its either all or nothing.

I did fix the other requests though
back to you

PR #156, opened 2019-01-24
HSD8-468 Global grunt sass compiler

pookmish:
# MAYBE READY FOR REVIEW

# Summary
- Global sass compiler for all humsci code

# Needed By (Date)
- no rush.

# Urgency
- the lowest possible

# Steps to Test

1. checkout this branch
1. `npm install`
1. `grunt watch`
1. edit some humsci scss files
1. validate compiles as expected.

# See Also


pookmish:
I'm not sure if this is a good approach. wanted to get a POC up to start the converstation

josephgknox (commented):
This is nice! The biggest headache (as we so often discussed) was having to manage multiple Gruntfile.js files and having to navigate to different levels of the file system in order to ensure you compile correctly.

QQ: with this model, can we drop the Gruntfile.js files at the lower levels (e.g., the one at`docroot/themes/humsci/su_humsci_theme`)?

Also, it doesn't appear that it's finding the include paths (error thrown that it cannot find a Decanter mixin - button primary - that I know is there). Not sure where the error is originating from, though, as the file is not specified in the message. I could spend more time looking at this, but wanted to throw it up here for now in case you know what's happening as I explore.

pookmish:
Yeah we would be able to remove all grunt file and node_modules directories everywhere except at the root level. with some tweaks you wouldn't even need the stanford_basic node_modules directory to compile in there.

as for the errors, i wonder if its attempting to compile something in a node_modules directory. try to delete all of those directories. you can find them with `find docroot -type d -name node_modules`. execute that from the root directory. let me know if that does anything

sherakama:
Hmm. This is interesting. 

This looks to have a lot of similarities to webpack. Perhaps it is time to investigate: https://github.com/webpack-contrib/sass-loader

josephgknox:
@pookmish yep, that worked!

Happy to look into Webpack with you, if you'd like. If we want to use something in the interim, this seems to be working really well from my perspective.

sherakama:
Where is this at? Does it need more review? Maybe ready? Maybe I'll review it. Maybe I won't.

pookmish:
i couldn't get webpack to work correctly in my first attempts. maybe we can circle back to webpack later?

josephgknox:
Haha! I think we've determined it works well, but were exploring Webpack simultaneously. Not sure we want to allow that to block this from being merged in the interim.

josephgknox:
Yeah, what Pookie said before me. :)

Gruntfile.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+      }
+    }
+  });
+
+  grunt.loadNpmTasks('grunt-sass');
+  grunt.loadNpmTasks('grunt-contrib-watch');
+
+  grunt.registerTask('default', ['availabletasks']);
+  grunt.registerTask('mike', ['stuff',], function () {
```
sherakama: Probably can remove this now.

Gruntfile.js
```diff
@@ -0,0 +1,84 @@
...
+    patterns.map(function(pattern){
+      glob.sync(pattern).map(function(file){
+        libraries.push(file);
+      });
+    });
+
+    return libraries;
+  }
+
+  grunt.initConfig({
```
sherakama, line 26: We have also become somewhat reliant on the autoprefixer add on. https://github.com/SU-SWS/decanter/blob/master/Gruntfile.js#L56 

It would be nice if you could add that here.

sherakama (approved):
Add the autoprefixer and this is GTG.

josephgknox (approved):
G2G

sherakama:
@pookmish Thanks for adding the prefixer, but it is not in the available tasks list, nor is it being called after sass is compiled.

josephgknox:
@pookmish branch has been restored for above request.

pookmish:
Moved to https://github.com/SU-HSDO/suhumsci/pull/181 since this PR has been closed

PR #160, opened 2019-01-29
Hsd8 545 events exporter

pookmish:
# READY FOR REVIEW

# Summary
- Moved all views related plugins and hooks into a separate module
- Created an optional view that would be removed later. It only would be added to MRC since MRC has the fields that are required. This view creates a rest export endpoint for consumption of other sites.
- adjust the events importer to accept a couple drupal fields.
- Modify core normalizer plugin so that the xml output could contain nested nodes.
- Modify core view serializer style to allow customization of xml tags.

# Needed By (Date)
- asap

# Urgency
- Medium

# Steps to Test
1. checkout this branch
1. sync mrc site. `blt drupal:sync --site=mrc`
1. view the local mrc site page `/api/events?_format=xml`
1. validate it displays an xml feed of all events on the site in a similar structure to events.stanford.edu
1. on another local site go to the events importer settings: `/admin/structure/migrate/events`
1. input the url for the mrc local site. something like `http://mrc.suhumsci.loc/api/events?_format=xml`
1. run the importer. `drush @mrc.local mim hs_events_image_importer; drush @mrc.local mim hs_events_importer`
1. Validate MRC events now live on the new local site and content matches.

# See Also


pookmish:
poke

docroot/modules/humsci/hs_views_helper/config/optional/views.view.data_export.yml
```diff
@@ -0,0 +1,1516 @@
...
+          table: node__field_s_event_speaker
+          field: field_s_event_speaker
+          relationship: none
+          group_type: group
+          admin_label: ''
+          label: ''
+          exclude: false
+          alter:
+            alter_text: true
+            text: '<div data-attribute-tag="person">{{ field_s_event_speaker }}</div>'
```
sherakama, line 1071: What is this data property for?

pookmish: https://github.com/SU-HSDO/suhumsci/pull/160/files#diff-aa13bdef96022fd9205ffe1d82cbe38bR37

its the only way i could find to have nested xml tags

sherakama: Hmmm. Yeah, this all seems to work and I see why you went this route. I tried to swap out for a new view mode but the template output was not great. I am not sure why the only formatter is `bricks` and why it only has one default mode. This seems like something that could have been expanded on but might not be worth the effort. Nice work on the custom root and item tags tho.

sherakama (approved):
@boznik What do you think about the approach for rendering the nested people items in the view?

docroot/modules/humsci/hs_views_helper/config/optional/views.view.data_export.yml
pookmish: even with different view modes its still renders out just plain html. so even with paragraphs/eck its the same outcome.

pookmish:
kinda needing these. production release has been delayed

boznik:
i'm ok with it. i'm wrestling my local suhumsci setup, so i may revisit.

PR #161, opened 2019-01-30
Hsd8 546 node clone action

pookmish:
# READY FOR REVIEW

# Summary
- Clone node action that works with VBO module.
- VBO module enhances core actions module by allowing configurable actions.
- Changed core bulk actions on content list page for VBO actions

# Needed By (Date)
- asap

# Urgency
- Medium

# Steps to Test
1. checkout this branch
1. Sync a site `blt drupal:sync --site=[sitename]`
1. view the content list `/admin/content`
1. check a box on a node to clone
1. in the actions dropdown at the top, choose "Clone selected content" and click "apply to selected items"
1. in the following page, choose a number of how many times to clone the node and submit
1. validate the node was cloned with all field content in place.

# See Also


pookmish:
poke

sherakama:
```
Notice: Undefined index: clone_count in Drupal\hs_actions\Plugin\Action\CloneNode->execute() (line 108 of /app/docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php) 
```

I get this error when running clone on this form:
http://archaeology.suhumsci.loc/admin/content/manage

Perhaps you should default to 1.

sherakama:
```
Action processing results: Access denied (1).
```
When trying to clone a course type.

pookmish:
admin/content/manage is a different page entirely. i'll add VBO to those pages. `/admin/content` should work

sherakama:
Yep, VBO actions show up it lots of places. Trying them all out.

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
```diff
@@ -0,0 +1,176 @@
...
+    $this->entityFieldManager = $entity_field_manager;
+    $this->entityTypeManager = $entity_type_manager;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function defaultConfiguration() {
+    return [
+      'clone_count' => 0,
```
sherakama: Perhaps default to 1. When would you want to clone 0 times?

pookmish: Prevent un-intentional cloning?

sherakama: That very well may be what happens but I would rather see prevention like that be more intentional.

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
```diff
@@ -0,0 +1,176 @@
...
+    $this->configuration['clone_count'] = $form_state->getValue('clone_count');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function access($object, AccountInterface $account = NULL, $return_as_object = FALSE) {
+    /** @var \Drupal\node\NodeInterface $object */
+    $result = $object->access('update', $account, TRUE)
+      ->andIf($object->getOwner()->access('edit', $account, TRUE));
```
sherakama: Am I getting this right, the `andIf()` statement is checking if the current user can edit the original owner (user entity). Why would we need that?

pookmish: Yeah. i copied that from the `AssignOwnerNode` action. i'm changing to if the user can update & create

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
pookmish: i've set it to default to 1, but also set that during execution incase it hasn't been set.

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
pookmish: Should be better logic now

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
```diff
@@ -0,0 +1,176 @@
...
+  protected function getReferenceFields($entity_type_id, $bundle) {
+    $fields = $this->entityFieldManager->getFieldDefinitions($entity_type_id, $bundle);
+
+    if ($this->entityTypeManager->hasDefinition('eck_entity_type')) {
+      $eck_types = $this->entityTypeManager->getStorage('eck_entity_type')
+        ->loadMultiple();
+      $clone_target_types = array_keys($eck_types);
+    }
+
+    $clone_target_types[] = 'paragraph';
```
sherakama, line 166: What about users, taxonomy terms, blocks, etc?

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    return $duplicate_entity;
+  }
+
+  /**
+   * Get fields that need to have their referenced entities cloned.
+   *
+   * @param string $entity_type_id
+   *   The entity type ID. Only entity types that implement.
```
sherakama: Incomplete sentence.

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
```diff
@@ -0,0 +1,176 @@
...
+
+    if ($this->entityTypeManager->hasDefinition('eck_entity_type')) {
+      $eck_types = $this->entityTypeManager->getStorage('eck_entity_type')
+        ->loadMultiple();
+      $clone_target_types = array_keys($eck_types);
+    }
+
+    $clone_target_types[] = 'paragraph';
+
+    // Filter out fields that we dont care about.
```
sherakama: What fields don't you care about and why? This could use a more detailed comment.

docroot/modules/humsci/hs_actions/src/Plugin/Action/CloneNode.php
pookmish: users, terms, etc don't get deleted if the node is deleted. ECK and paragraphs do. so we gotta clone them to produce new entity IDs

sherakama: 🤔 Hmm, that is an interesting conundrum. So you want _some_ of the references cloned and _some_ not. Makes sense here.

pookmish: yup. media entities, terms, etc we want to reference the same entity id as the parent.

pookmish: I think ECK and paragraphs are the only ones that require that

sherakama:
Thanks for making the change to the access check. This works much better now.

sherakama (approved):

pookmish:
kinda needing these. production release has been delayed

PR #164, opened 2019-02-05
HSD8-312 HSD8-504 Show external link icon and style to match

pookmish:
# READY FOR REVIEW

# Summary
- Show and style external link icons.

# Needed By (Date)
- 2/13

# Urgency
- low

# Steps to Test
1. checkout this branch
1. sync a site like archaeology `blt drupal:sync --site=archaeology --partial` (use partial to keep existing views
1. verify any external link in the main content region has an external link icon inline with the text. including heading tags.

# See Also


docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
```diff
@@ -2,6 +2,11 @@
   'use strict';
   Drupal.behaviors.suHumSciTheme = {
     attach: function (context, settings) {
+      var headings = ['h2', 'h3', 'h4', 'h5', 'h6'];
```
sherakama, line 5: Why are you attaching these classes with javascript?

pookmish: can't apply css based on children of an element.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
```diff
@@ -34,7 +34,7 @@
     {% endif %}
 
     {% if title %}
-      <h3 class="horizontal-card__title">{{ title }}</h3>
+      <div class="horizontal-card__title">{{ title }}</div>
```
sherakama, line 37: Why are you changing this from a heading tag? Seems like a heading tag would be better for the document outline?

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
sherakama: Where are you unable to control the markup that you cannot wrap a heading in a link?

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
```diff
@@ -158,3 +158,33 @@ figure {
...
+    font-size: 60%;
+    padding: 0 5px;
+
+    .fa-arrow-right {
+      transform: rotate(-45deg);
+    }
+
+    // Duplicated text "Link is External" occurs in screen readers.
+    // Hide this one from screen readers to prevent repeated text.
+    .visually-hidden {
```
sherakama, line 186: Wouldn't we want to remove the class from the link instead of changing the behavior of the visually-hidden css class?

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
pookmish: I've seen H&S more often than not set the title to be an `h2` tag. Also because we arent the site builders we can't control the structure of the page. with this being an `h3` there _must_ be an `h2` above it.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
pookmish: the class is added via extlink module javascript with no triggers to act upon it.

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
pookmish: depending on what the user configures in the ui for field markup.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
sherakama: with HTML5 I'm not sure that is true anymore.
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#Defining_headings

This is long but full of info and kind of breaks apart what Mozilla says:
https://css-tricks.com/document-outline-dilemma/
 
Then there is this from 2009: https://www.456bereastreet.com/archive/200901/headings_heading_hierarchy_and_document_outlines/
which has the rules

1. Each document should have a logical heading structure that starts at level 1 and doesn’t skip any levels
2. Anything that visually looks like a heading should be marked up as a real heading
3. The title of the document’s main content should be a level 1 heading
4. There should be only one level 1 heading in each document

Finding specs: https://www.w3.org/TR/WCAG20-TECHS/H69.html
> Authors are encouraged to nest headings hierarchically

Doesn't seem to be required but helpful. 

Looks like the best practices are to have an "as hierarchical" structure as possible and to use heading tags instead of other tags with css to show visual importance as screen readers won't pick up on that.

pookmish: I had a chat with jiatyan and she said any skipped headings are a :no_good_man:  (thats a "no_good_man" emoji)

sherakama: Is it a navigation problem with the screen readers that gaps create? IE: "Tell me all h2s"?

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
sherakama: Drat. Ok, this is fine then.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
pookmish: I think so. besides that currently with the way it is like what is on the live site is an `h2` inside an `h3`. definitely not a good thing there

sherakama: Tell me more

pookmish: i cant remember what site it was but they used this pattern in a view. then the title field they configured to render as an h2

sherakama: Aye, I agree. H2s in H3s is very bad but that can be solved with tag stripping. 

This is going to be a large problem for us with products open to general use. I want to make sure we get this right and do the best we can to protect from poor config choices. 

Preventing nested headings is a good idea. Missing out on headings or incorrectly setting heading levels is going to be hard.

pookmish: And H&S would prefer to have "Unstyled" patterns. they would rather add styling via field classes to accomplish their desired outcome.

sherakama (approved):
Ok, The functionality is working as intended so this PR is approved. This has stemmed a bigger picture discussion on the component approach with Drupal. I'll put some time on the calendar tomorrow afternoon for a deep discussion.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
sherakama: Interesting.

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
sherakama: Sounds like we need field level components (atoms)

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
josephgknox: Yep, the "unstyled" patterns have been a discussion with them. To what point they want things to be unstyled is variable, too. They want structure and general layout, but maybe not size (dictated by heading tag or otherwise), spacing, etc.

What if our patterns/components never used heading tags by default, but rather they leveraged the display variables (A, B, C, D, etc.) to make a title div look like a heading. So, heading tags are only used in plain, WYSIWYG like content. While this alone wouldn't close the door on users making poor decisions, it would verify that we are providing them with tools that by default may lead to the wrong decision.

josephgknox:
Yep, the functionality is working as intended for me, too. Approved.

josephgknox (approved):
Approved, pending conflict cleanup.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.html.twig
pookmish: From what i understand from them they literally want no styling on a pattern.

PR #169, opened 2019-02-06
HSD8-579 Use ultimate cron to break up migration importers

pookmish:
# READY FOR REVIEW

# Summary
- use ultimate cron to create separate cron tasks for each migration group.
- Cron queue was not a viable solution since we want these items to execute on every cron. cron queue doesn't do that.

# Needed By (Date)
- 2/13

# Urgency
- medium

# Steps to Test
1. Checkout this branch
1. sync a site like philit `blt drupal:sync --site=philit --partial`
1. view the page `/admin/config/system/cron/jobs` and ensure 3 cron jobs for "H&S Migrate" module category.
1. run each of the migration cron jobs and ensure they are successful (Capx will take a bit of time)
1. run cron from drush and ensure it's successful
1. review the `/admin/config/system/cron/jobs` page again to ensure no errors on any cron job.

# See Also


sherakama:
Ultimate

config/default/ultimate_cron.job.ultimate_cron_cron.yml
```diff
@@ -0,0 +1,15 @@
+uuid: 488cb667-dd02-4bb7-8763-958f99390a10
+langcode: en
+status: true
+dependencies: {  }
+title: 'Runs internal cleanup operations'
+id: ultimate_cron_cron
+weight: 0
+module: ultimate_cron
+callback: ultimate_cron_cron
```
sherakama, line 9: Department of redundancy department.

docroot/modules/humsci/hs_migrate/hs_migrate.module
```diff
@@ -87,14 +88,36 @@ function hs_migrate_entity_delete(EntityInterface $entity) {
...
   $migrations = hs_migrate_migration_list();
-  // Take it one group at a time, importing the migrations within each group.
-  foreach ($migrations as $migration_list) {
-    array_walk($migration_list, 'hs_migrate_execute_migration');
+
+  $migration_group = str_replace('hs_migrate_', '', $cron_entity->id());
+  // Execute the migration entities in the provided migration group.
+  if ($migration_group && !empty($migrations[$migration_group])) {
+    try {
+      array_walk($migrations[$migration_group], 'hs_migrate_execute_migration');
```
sherakama, line 107: Does this create blocking iteration? As in, if you have 4 migrations to run and the second one throws an error does this block the third and fourth execution? You may actually want to refactor this to use a loop or send in an anonymous function with the try/catch in it.

pookmish: it would only block migrations within the same group. for example theres a `hs_capx_images` importer and a `hs_capx`. if the images fail to import, the profiles wont import. which in my mind is an ideal thing

sherakama: That totally makes sense. Thanks. I'll run through the test steps now.

sherakama:
This all seems to be functioning without errors.

I did notice on `admin/config/system/cron/settings/launcher` the setting for `run in thread` is required but did not have a setting configured.

screen shot 2019-02-06 at 10 43 29 am

sherakama (approved):
GTG, with the addition of the missing setting.

@boznik you're up.

pookmish:
Interesting about the thread. All of the configs were generated when the module was enabled so they must not require that from the config side

boznik:
so what's up with the setting change? is there more work to be done?

pookmish:
nothing more to do. just an observation shea had i believe

PR #189, opened 2019-02-12
HSD8-555 shortcut filter

pookmish:
# READY FOR REVIEW

# Summary
- Remove shortcut links if the user doesnt have access.
- Added permission that allows for roles to hide the "Manage" button in the toolbar

# Needed By (Date)
- Someday

# Urgency
- Low

# Steps to Test
1. checkout this branch
1. update database `drush @sitename.local updb -y`
1. add a link to an advanced admin page link `/admin/structure/views`(shortcuts can be configured at `/admin/config/user-interface/shortcut`)
1. add a user with a "Site Manager" role
1. masquerade as that user.
1. ensure you don't see the link you created for the admin page.
1. remove the permission  "View Toolbar Manage Button" from the "Site manager" role `drush @sitename.local role:perm:remove site_manager "view toolbar manage"`
1. verify you only see the shortcuts in the toolbar and not the "Manage" button
1. unmasquerade and verify you see the "Manage" button in the toolbar again.

# See Also

### DIFF
--- a/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
+++ b/docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
@@ -174,7 +174,7 @@ function su_humsci_profile_simplify_condition_forms(array &$condition_elements, 
 /**
  * Implements hook_ENTITY_TYPE_insert().
  */
-function hs_field_helpers_eck_entity_type_insert(EntityInterface $entity) {
+function su_humsci_profile_eck_entity_type_insert(EntityInterface $entity) {
   $eck_type = $entity->id();
   // When a new ECK entity type is create, set initial permissions so that
   // site builders aren't required to search for the necessary permissions.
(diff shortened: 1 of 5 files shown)


docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
```diff
@@ -174,7 +174,7 @@ function su_humsci_profile_simplify_condition_forms(array &$condition_elements,
 /**
  * Implements hook_ENTITY_TYPE_insert().
  */
-function hs_field_helpers_eck_entity_type_insert(EntityInterface $entity) {
+function su_humsci_profile_eck_entity_type_insert(EntityInterface $entity) {
```
sherakama, line 177: whoops. How'd we let that one through?

pookmish: and it still worked which was surprising. :man_shrugging:

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
```diff
@@ -192,3 +192,50 @@ function hs_field_helpers_eck_entity_type_insert(EntityInterface $entity) {
...
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function su_humsci_profile_preprocess_menu(&$variables) {
+  if ($variables['menu_name'] != 'shortcut_menu') {
+    return;
+  }
+
+  $current_user = \Drupal::currentUser();
+  _su_humci_profile_clean_shortcut_links($variables['items'], $current_user);
```
sherakama, line 205: Is the "_" private function still a standard in D8? I can't seem to find anything on this officially: https://www.drupal.org/docs/develop/standards/coding-standards

sherakama:
I haven't pulled down and tested to local yet but shouldn't the shortcut menu items the user doesn't have access to be hidden already? Why do you have to check and trim them out?

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
```diff
@@ -192,3 +192,50 @@ function hs_field_helpers_eck_entity_type_insert(EntityInterface $entity) {
...
+    }
+  }
+}
+
+/**
+ * Implements hook_page_attachments().
+ */
+function su_humsci_profile_page_attachments(array &$attachments){
+  // Conditionally attach an asset to the page.
+  if (!\Drupal::currentUser()
```
sherakama: Do we want to also add a check here so that we avoid attaching this css for anonymous users?

pookmish:
Unfortunately there are no access checks by shortcut module. so we gots ta do it ourself :( see [here](https://api.drupal.org/api/drupal/core%21modules%21shortcut%21shortcut.module/function/shortcut_toolbar/8.2.x) it only checks for access to the shortcut set. not the individual items.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
pookmish: good thinkin. :+1:

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
pookmish: hmm not sure. I thought it was always just convention but not really a standard

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.profile
pookmish: instead of checking for anonymous, i checked if user has permission to view the toolbar. essentially it results in the same, but would also avoid adding the css when we go about a private page functionality later.

boznik (approved):
This is working as expected.

PR #192, opened 2019-02-15
HSD8-596 Added plugin for clone action to change field values

pookmish:
# READY FOR REVIEW

# Summary
- Add to clone node action to allow for the user to incrementally increase date field values. Used for recurring events.

# Needed By (Date)
- 2/27

# Urgency
- High

# Steps to Test
1. Checkout this branch
1. `composer install --prefer-source` for good measure
1. sync a site `blt drupal:sync --site=[sitename]`
1. On `/admin/content` page, find an event (or make one if none exist) and click the checkbox for that row and choose "Clone" action at the top
1. in the next form, expand the "Adjust Cloned Field Values" and set the incremental values for the date field.
1. Choose to clone at least 3 times for best testing
1. click apply to clone the node
1. View the new nodes and validate each one has incremented date value. for example if the original event was 1/15/2019 12:01PM and you configured to increment by 5 hours. the cloned content should be 1/15/2019 5:01PM, 1/15/2019 10:01PM, 1/16/2019 3:01AM respectively.

# See Also


docroot/modules/humsci/hs_actions/src/Plugin/Action/FieldClone/Date.php
```diff
@@ -0,0 +1,122 @@
...
+  protected $entityIds = [];
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state) {
+    $form = [];
+
+    $increment = range(0, 12);
+    unset($increment[0]);
```
sherakama, line 33: Why? Why not `range(1, 12)`? Do the keys have to align?

sherakama: Oh, yes, I guess they do.

sherakama (approved):

sherakama:
Overall, interesting approach. Did you check to see if there was something in the contrib space that seemed similar?

pookmish:
There wasn't even a clone action, let alone anything that would modify field values like this.

sherakama:
Good to know, Thanks for looking.

boznik:
during blt sync:drupal:

ypeError: Argument 1 passed to Drupal\Component\Utility\NestedArray::getValue() must be of the type array, boolean given, called in /Users/ggarvey/www/suhumsci/docroot/modules/contrib/config_ignore/src/Plugin/ConfigFilter/IgnoreFilter.php on line 147 in Drupal\Component\Utility\NestedArray::getValue() (line 69 of /Users/ggarvey/www/suhumsci/docroot/core/lib/Drupal/Component/Utility/NestedArray.php).

PR #193, opened 2019-02-15
HSD8-600 Generic 3 column pattern with variants

pookmish:
# READY FOR REVIEW

# Summary
- Created a new Generic pattern with no styles, only layout structures with some variants.
- Added a twig filter to check if the variable has markup. unfortunately the filter has to be in a module and can't live in the theme.

# Needed By (Date)
- 2/27

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. `composer install --prefer-source` for good measure
1. clear cache/sync a site `blt drupal:sync --site=[sitename]`
1. Create a new test view to display any content type you want
1. add some fields. change the view "SHOW" settings from "fields" to "pattern" and choose the "Generic Three Column" pattern.
1. map the fields to the left, middle, and right regions
1. view the output.
1. change the settings of the pattern to use a different variant. like the "Larger right" or something
1. observe the output changes.

# See Also


docroot/themes/humsci/su_humsci_theme/patterns/molecules/generic-three-column/generic-three-column.html.twig
```diff
@@ -0,0 +1,41 @@
...
+{% elseif '50_25_25' in variant %}
+  {% set leftAttributes = leftAttributes.addClass('larger') %}
+{% elseif '25_25_50' in variant %}
+  {% set rightAttributes = rightAttributes.addClass('larger') %}
+{% endif %}
+
+{% spaceless %}
+  <div{{ attributes }}>
+
+    {%- if left|has_markup or static -%}
```
sherakama, line 22: I don't think you need the `-` here when you wrap the whole thing in `{% spaceless %}`.

docroot/themes/humsci/su_humsci_theme/patterns/molecules/generic-three-column/generic-three-column.html.twig
```diff
@@ -0,0 +1,41 @@
...
+      <div{{ leftAttributes }}>
+        {{ left }}
+      </div>
+    {%- endif -%}
+
+    {%- if middle|has_markup or static -%}
```
sherakama, line 28: Although I like this filter is it really necessary? Why is the https://twig.symfony.com/doc/2.x/filters/striptags.html filter not enough?

pookmish: we'd have to specify every time all the tags that would present valid content. it would have to be something like 
`middle|strip_tags(" <iframe> <object> <picture> <figure>...")` for every time we want to check

sherakama: So in this case, you would have to do it three times? And you would gain the flexibility and granularity of specifying those tags for each instance. I get the DRY thinking of this but I think it adds something to maintain when you already have a tool to use.

pookmish: in the future if we discover a tag that is appropriate, we'd have to add it to every line where this would be used. hyptothetically if we left out `<iframe>` and realize its valid markup, we'd have to manually add it to every line, not just this one patter (when we use it more)
Also i dont think its much to maintain. its a 1 line static method vs several lines in twigs. :man_shrugging:

sherakama: Yeah, it is a trade for sure. We need more opinions. 

@boznik @jbickar may we have your thoughts on this.

pookmish: the two options are 
1. have a special twig filter that can be referenced with `foo|has_markup`
  * Small code to maintain
2. Use a built in twig filters like `foo|render|striptags(' <a> <iframe> <object> <picture> <figure>')|trim` for each condition check.

boznik: i'm inclined to go with pookmish on this one. maybe we should put this in a seperate twig helper module where we could all reuse special filters across different sites?

boznik: watcha think @sherakama @jbickar ?

sherakama: I like that idea. I am worried that we will be creating templates that cannot be shared across our projects and will be dependant on modules. Let's start a project to contain these twig plugins and require it in our themes.

boznik: so should this get merged  and moved to a separate module later?

sherakama: Yeah, I don't want to hold this up any longer. Go forth and refactor later.

sherakama: I've added a topic to our next dev chat to go through this.

PR #307, opened 2019-06-20
Hsd8-653 Generic 3 column 1-5-1 variant

pookmish:
# READY FOR REVIEW

# Summary
- Added variant to 3 column pattern for 1 - 5 - 1 organization. This allows the middle column to always be 5 times wider than the sides.

# Needed By (Date)
- ASAP

# Urgency
- Medium. Math site needs it.

# Steps to Test
1. Checkout this branch
1. clear cache
1. view `/patterns/generic_three_column` page and ensure the "Small Side Columns" variant has a middle are that is 5 times wider than the sides

# See Also


docroot/themes/humsci/su_humsci_theme/patterns/molecules/generic-three-column/scss/generic-three-column.scss
```diff
@@ -34,4 +34,9 @@
     }
   }
 
+  &[class*="1-5-1"] {
```
sherakama: Why the wildcard class selector? There are only two classes with the convention you want. Why not just name them here? I would suspect that explicitly declaring the classes to act upon is the more performant (but less sexy) code.

pookmish: you make a good point. i'm on board :boat:

sherakama (approved):

boznik (approved):
This is working as expected

PR #324, opened 2019-07-02
Fix table header color

pookmish:
# READY FOR REVIEW

# Summary
- Set background color and text color on table headers
- changed weight on humsci theme css library

# Needed By (Date)
- asap

# Urgency
- high

# Steps to Test
1. checkout branch
1. review `/peoplegraduate-students/phd-alumni` on mathematics for accessible header color

# See Also


sherakama (approved):

sherakama (changes requested):
Are you sure you want to move all the stylesheets up to the `theme` level in the cascade?

sherakama:
Why do the base level elemet styles need to be brought up to the `theme` level. This goes against the intentions of those levels. 

```
Drupal 8 follows a SMACSS-style categorization of its CSS rules:

Base — CSS reset/normalize plus HTML element styling.
Layout — macro arrangement of a web page, including any grid systems.
Component — discrete, reusable UI elements.
State — styles that deal with client-side changes to components.
Theme — purely visual styling (“look-and-feel”) for a component.
```

By putting everything at the `theme` level, you are making it harder to override them for others later.

pookmish:
well maybe just up to the `layout` level then. they gotta come up one level because `stanford_basic` is currently taking precedence https://github.com/SU-SWS/stanford_basic/blob/8.x-2.x/stanford_basic.libraries.yml#L93

sherakama:
You can adjust the weight in each of those 5 levels. Why not do that?

sherakama:
https://www.drupal.org/docs/8/theming/adding-stylesheets-css-and-javascript-js-to-a-drupal-8-theme
```
weight | Adjusts order relative to other assets (within the same SMACSS group).Default: 0. Use a numeric between -50 to +50. | { weight: 1 }
```

pookmish:
Ahh thank you. thats what i was looking for!

pookmish:
That should work out better. i chose 5 to allow for a small possibility of stuff being in between. :man_shrugging: doubt it'll ever happen but i like even 5's

sherakama (approved):
I like that.

PR #337, opened 2019-07-15
updated blt to 10.x-dev

pookmish:
# READY FOR REVIEW

# Summary
- Updated BLT & Drush

# Needed By (Date)
- some day

# Urgency
- low

# Steps to Test
1. checkout this branch
1. try out some blt and drush commands to ensure it still works.

# See Also


blt/.schema_version
```diff
@@ -1 +1 @@
-9001010
\ No newline at end of file
+10001000
```
sherakama, line 1: https://www.youtube.com/watch?v=_4TPlwwHM8Q

blt/blt.yml
```diff
@@ -62,7 +62,7 @@ cm:
   allow-overrides: true
 sync:
   commands:
-    - 'custom:hello'
+    - 'blt:init:settings'
```
sherakama, line 65: You want to int the settings on every sync? Why?

blt/src/Blt/Plugin/Commands/CircleCiCommands.php
```diff
@@ -273,7 +273,7 @@ protected function runBehatTest(array $sites) {
   protected function setupSite() {
     $tasks[] = $this->installDependencies();
     $tasks[] = $this->waitForDatabase();
-    $tasks[] = $this->taskExec('service apache2 start');
+    $tasks[] = $this->taskExec('apachectl stop; apachectl start');
```
sherakama, line 276: restart or reload not work?

blt/src/Blt/Plugin/Commands/CircleCiCommands.php
```diff
@@ -377,8 +377,7 @@ protected function installDrupal($profile = 'minimal') {
       ->drush('site-install')
       ->args($profile)
       ->option('verbose')
-      ->option('yes')
-      ->option('db-url', static::DB_URL, '=');
```
sherakama, line 381: Is the DB URL option now handled by something else?

blt/blt.yml
pookmish: `custom:hello` no longer exists and the sync command actually runs `blt:init:settings` as the 2nd step. so this is just doing it twice. I do this because i dont like how blt tries to run composer in step 1

blt/src/Blt/Plugin/Commands/CircleCiCommands.php
pookmish: It does, but stop and restarting allows for `$_ENV` variables to pass into apache.

blt/src/Blt/Plugin/Commands/CircleCiCommands.php
pookmish: through `docroot/sites/settings/ci.settings.php`

composer.json
```diff
@@ -205,7 +211,11 @@
     "su-sws/stanford_media": "dev-8.x-1.x",
     "su-sws/stanford_ssp": "dev-8.x-1.x"
   },
+  "replace": {
+    "dealerdirect/phpcodesniffer-composer-installer": "*"
```
sherakama, line 215: Perhaps Greg or I should take a pass at this issue. This seems a tad risky.

docroot/sites/settings/acquia.settings.php
```diff
@@ -0,0 +1,55 @@
...
+  default:
+    $config['environment_indicator.indicator']['bg_color'] = '#086601';
+    $config['environment_indicator.indicator']['fg_color'] = '#fff';
+    $config['environment_indicator.indicator']['name'] = EnvironmentDetector::getAhEnv();
+    break;
+}
+
+if (EnvironmentDetector::isProdEnv()) {
+  $config['system.file']['path']['temporary'] = "/mnt/gfs/$group.$environment/tmp";
+}else {
```
sherakama: formatting.

composer.json
pookmish: I could not for the life of me get this to work. Even patching the library didnt work. I also don't think it belongs in `dependencies`. If anything BLT should require it as a `devDependency` IMO. All this library does is automatically install phpcs standards when a dependency changes/added/removed. Not really needed.

docroot/sites/archaeology/settings.php
```diff
@@ -790,8 +790,13 @@
 #   include $app_root . '/' . $site_path . '/settings.local.php';
 # }
 
-if (file_exists('/var/www/site-php')) {
-  require '/var/www/site-php/swshumsci/archaeology-settings.inc';
```
sherakama, line 794: This is all handled in BLT's black magic now? I don't see this require anywhere else in this pr?

pookmish: yep. more obscurity https://github.com/acquia/blt/blob/10.x/settings/blt.settings.php#L125

boznik:
i had to create the local.settings.php file and add back pulling the db info from the settings dir

sherakama (approved):
If Greg is OK with it I am ok with it.

I'd like to remove the replace section as soon as we can so keep that in mind.

pookmish:
me too. i hate having that there. but :man_shrugging: I did change the release for 8.1.5 which isn't scheduled for production release until aug 7th so we'll have a little time to find if anything goes :boom:

boznik:
this is working and i didn't find a way around the replace method.

PR #346, opened 2019-07-17
HSD8-682 Fix table pattern when used in views

pookmish:
# READY FOR REVIEW

# Summary
- Fix table pattern when using it in views.

# Needed By (Date)
- 7/31

# Urgency
- medium

# Steps to Test
1. checkout this branch.
1. temporarily edit `docroot/sites/blt.development.services.yml` file and set `debug: false` (this keeps html comments and stuff from showing up in the dom, so you can simulate prod site)
1. go to `/test-table-view` and ensure each row has the same number of columns.

# See Also


boznik (approved):
This is working as expected

sherakama (approved):

PR #348, opened 2019-07-18
Fix group block links from generating js errors

pookmish:
# READY FOR REVIEW

# Summary
- Fix group block links in layout builder
- The data attributes have to match since contextual.js file will look at them and run an ajax call. Without the attributes matching exactly it will throw an error and cause contextual links to fail.

# Needed By (Date)
- asap

# Urgency
- high

# Steps to Test
1. checkout this branch
1. temporarily edit `config/envs/local/config_ignore.settings.yml`
1. add `'core.entity_view_display*'` into the `ignored_config_entities` set (this will allow you to do the following step without drush syncing the view display settings)
1. visit `/admin/structure/types/manage/hs_event/display/default/layout` page
1. ensure you get contextual links on all fields, group blocks and anywhere appropriate in the layout builder
1. validate the "Add block to group" still functions correctly.

# See Also

### DIFF
--- a/docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
+++ b/docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
@@ -4,6 +4,7 @@ namespace Drupal\hs_blocks\Plugin\Block;
 
 use Drupal\block_content\Access\RefinableDependentAccessInterface;
 use Drupal\block_content\Access\RefinableDependentAccessTrait;
+use Drupal\Component\Utility\Crypt;
 use Drupal\Component\Uuid\UuidInterface;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Block\BlockBase;
@@ -11,7 +12,9 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Plugin\Context\ContextRepositoryInterface;
 use Drupal\Core\Plugin\PluginBase;
+use Drupal\Core\PrivateKey;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Site\Settings;
 use Drupal\Core\Url;
 use Drupal\layout_builder\SectionComponent;
 use Drupal\layout_builder\SectionStorageInterface;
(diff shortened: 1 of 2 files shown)


docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
```diff
@@ -180,25 +192,36 @@ protected function buildAdministrativeSection(array &$components) {
...
+      'delta' => $section_delta,
+      'group' => $this->configuration['machine_name'],
+    ];
+
+    // Build a contextual id that won't generate errors from ajax.
+    // @see _contextual_id_to_links()
+    // Use contextual link data attributes so that layout builder doesn't
+    // disable the link from being clicked.
+    // @see behaviors.layoutBuilderDisableInteractiveElements() in layout-builder.js
+    $contextual_id = $this->configuration['machine_name'] . ':' . http_build_query($route_params) . ':langcode=en';
```
sherakama, line 207: I know we don't do multilang but wouldn't it be better to grab the active language instead of hardcoding it here?

docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
```diff
@@ -180,25 +192,36 @@ protected function buildAdministrativeSection(array &$components) {
...
-        [
-          'attributes' => [
-            'class' => ['use-ajax', 'new-block__link'],
-            'data-dialog-type' => 'dialog',
-            'data-dialog-renderer' => 'off_canvas',
-            'data-contextual-id' => 'add-group-link',
-          ],
-        ]
-      ),
+      '#url' => Url::fromRoute('hs_blocks.choose_block', $route_params, ['attributes' => $route_attributes]),
```
sherakama, line 224: Yeeesh, that was a lot for these simple links.

sherakama (approved):
Haven't validated code runs but approach looks sane.

docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
pookmish: it doesnt really do anything in the end. i just grabbed what another contextual link has. it could probably even be an empty string. just need 2 of the `:` in there for the `list` and `explode` in contextual module

docroot/modules/humsci/hs_blocks/src/Plugin/Block/GroupBlock.php
pookmish: you're telling me. :face_with_head_bandage:

boznik:
validated this is working

PR #349, opened 2019-07-18
HSD8-681 prune courses

pookmish:
# READY FOR REVIEW

# Summary
- Delete orphans of courses migration.

# Needed By (Date)
- 7/31

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. run a fresh install `drush si config_installer -y; drush cim -y; drush uli`
1. on `/admin/structure/migrate/course-importer` set the course importer to a really small set of courses like `https://explorecourses.stanford.edu/search?q=MS%26E111&view=xml-20140630` 
1. run migration `drush mim hs_courses`
1. view `/admin/content` and count and take note of what courses have imported
1. back on `/admin/structure/migrate/course-importer` change the importer to another set of courses, something like `https://explorecourses.stanford.edu/search?q=MS%26E112&view=xml-20140630`
1. run migration `drush mim hs_courses`
1. view `/admin/content` and verify the original courses have been removed and you now only have courses from the new xml source.

# See Also


docroot/modules/humsci/hs_migrate/src/EventSubscriber/EventsSubscriber.php
```diff
@@ -0,0 +1,127 @@
+<?php
+
```
sherakama, line 2: Perhaps a file comment about what this subscriber is doing would be helpful here.

pookmish: I'll update the class comment since `@file` are not allowed in namespaced files https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards#file

config/default/migrate_plus.migration.hs_courses.yml
```diff
@@ -16,6 +16,7 @@ migration_group: hs_courses
 label: 'Course importer'
 source:
   track_changes: true
+  delete_orphans: true
```
sherakama, line 19: Migrate or migrate plus doesn't already have this as an option somewhere?

pookmish: no. migrate was originally built to just consume one time data. if it was wrong it has a roll back option. but no syndication built in.

sherakama: https://www.drupal.org/node/2809433

sherakama: https://www.drupal.org/project/migrate_tools/issues/2863426

pookmish: ahh interesting. looks like the first link has a similar direction. but neither of these issues allows for a config option to sync vs just import. maybe at a later date this can be removed.

sherakama (approved):
I'm surprised that this option is not available already. Looks like it is an ongoing issue and nothing stable to rely on. What you have provided seems reasonable but the Drupal issue threads linked in this PR seem to be getting to some nitty-gritty details that would be worth reviewing before committing to this solution.

config/default/migrate_plus.migration.hs_courses.yml
pookmish: the 2nd link is just a drush command. i dont wanna have to run a drush command for per migration per site.

boznik (approved):
This is working as expected

PR #355, opened 2019-07-25
HSD8-686 stage file proxy

pookmish:
# READY FOR REVIEW

# Summary
- Use stage_file_proxy module on dev, stage and local
- disable dblog
- custom module to record 404 & 403 reports with custom pages at the same url as the ones provided by dblog

# Needed By (Date)
- 7/31

# Urgency
- Medium

# Steps to Test
1. checkout this branch
1. `blt drupal:sync --site=[siteofyourchoice] --partial`
1. ensure theres a "Top 404 Pages" and "Top 403 Pages" reports pages under the "Reports" menu in the toolbar
1. try a random url and ensure it appears in the 404 page list.
1. try to view `/admin` when logged out in an incognito window and ensure it appears in the 403 list.

# See Also


sherakama:
Do you have to log the 404, 403s? Isn't this information available somewhere on Acquia's infrastructure? Perhaps this is an easier way of gathering them though. Might be expensive to parse the log files. Anything from the cloud API?

pookmish:
There is through version 2 of cloud api, but i tried to authenticate with oauth for v2 and had no lucky. I'm not sure how expensive it would be to parse that data. https://cloudapi-docs.acquia.com/#/Environments/getEnvironmentsLogs

sherakama:
I guess it would be a trade then, a DB write for every 404, 403 that hits the system vs an expensive parse and added development for gathering the information on demand. You'll likely want to clear out these 404, 403 logs on a regular basis as well as those tables are likely to bloat rapidly and slow everything down.

pookmish:
Good point. maybe a maximum of 100 rows, trimmed during cron. and then a manual truncate button. i think that should be enough. thoughts?

sherakama:
My first vote is to change over to on-demand fetching from the API or raw logs but if you are blocked by time or effort I'd say probably no more than a 1000 records with the added functionality you suggested.

pookmish:
raw logs are compressed for each day. There would only be a report for the current day of 404/403 reports without uncompressing each of those. https://docs.acquia.com/acquia-cloud/monitor/logs/#cloud-download-logs-cli

I'll put in a record trimming. The reason why i thought 100 records is because this solution doesnt record each 404 as a new entry in the database. it increments it and only displays the top 25 paths in the report page. 1000 would probably be unnecessary

pookmish:
Cron and manual form delete has been implemented. wanna review again?

boznik (approved):
This is working as expected

PR #364, opened 2019-08-08
When an image is in a masonry, dont do lazy loading

pookmish:
# READY FOR REVIEW

# Summary
- Dont lazy load masonry images.

# Needed By (Date)
- ASAP

# Urgency
- HIGH

# Steps to Test
1. checkout this branch
1. `blt drupal:sync --site=archaeology --partial`
1. view `/research/publications` to ensure masonry appears to have applied correctly.

# See Also


docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
```diff
@@ -232,3 +232,28 @@ body *:focus {
 .decanter-font-lead {
   max-width: 100%;
 }
+
+@-moz-keyframes spin {
```
sherakama, line 236: Your sass compiler could benefit from an autoprefixer so you wouldn't have to add them yourself.  
https://www.npmjs.com/package/autoprefixer

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
```diff
@@ -46,7 +46,13 @@
         });
       });
 
-      $("img.lazy", context).lazyload();
+      $('img.lazy', context).each(function () {
+        if ($(this).closest('.masonry').length) {
```
sherakama, line 50: So this is to find the images in a masonry group and then set their src and load them right away because all images get the loading image at first. Yeah, I guess you have to do this with assuming all images are 'lazy'.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
pookmish: oh thats already there, maybe i just need 
```
spin {
  100% {
   transform: rotate(360deg);
  }
}
```
???

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
pookmish: yeah that seemed to be a quick fix. the library i'm using doesn't provide callbacks when an image is loaded to rearrange the masonry. and the thumbnails i found don't stretch themselves to the `width` and `height` attributes. the loading.png only stretches its width but not height.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
sherakama: If you want to animate you will still need that property but you will only need one keyframe declaration

https://github.com/SU-SWS/decanter/pull/443/files#diff-66c59ec628d07d6fc01b96308e6094e5R28

docroot/themes/humsci/su_humsci_theme/js/su_humsci_theme.js
pookmish: Another alternative that might work is to use more jquery. set inline styles that set the height with the correct aspect ratio of the attributes, but takes the actual width into consideration.
like an image: `` but its container is only 250px wide. the jquery would set inline styles to `style="width: 250px; height: 375px"`. but then it would need to adjust that on responsive. :man_shrugging:

sherakama: You would have to add the https://imagesloaded.desandro.com/ package.

Lots of people use masonry and that package together but your solution seems light enough. You don't get the lazy load in masonry tho. Might be worth looking in to that library.

pookmish: thats what is being used in the masonry. https://github.com/SU-HSDO/suhumsci/blob/develop/docroot/themes/humsci/su_humsci_theme/patterns/organisms/masonry/masonry.ui_patterns.yml#L37
but the lazy loading library i'm using doesnt have a callback to trigger the `.layout()` method

sherakama: Gotcha.

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
pookmish: hmm. i see autoprefixer in the Gruntfile.js but it just compiles to 
```
@keyframes spin {
  100% {
    transform: rotate(360deg);
  }
}
```
maybe i'm doing something wrong

sherakama (approved):

docroot/themes/humsci/su_humsci_theme/scss/base/_global.scss
sherakama: That probably means it is compatible with the browsers we support

https://caniuse.com/#search=keyframe

pookmish: oh ha! look at that, it still works. :+1:

PR #365, opened 2019-08-08
Fix lazy masonry loaders different library

pookmish:


sherakama (approved):
Code looks good. Haven't pulled down to test. Relying on Mike's good graces.

PR #411, opened 2019-10-01
Use api v2

pookmish:
# READY FOR REVIEW

# Summary
- Change Acquia cloud api calls to use the new V2 api
- added cron job in circleci to automate Cert renewal on the 1st of every month.

# Need Review By (Date)
- No rush

# Urgency
- low

# Steps to Test
1. manual review of code.

# See Also


.circleci/config.yml
```diff
@@ -180,6 +180,19 @@ update_dependencies: &update_dependencies
...
+    - *add_ssh
+    - *add_known_hosts
+    - checkout
+    - restore_cache: *restore_cache
+    - run: composer install --optimize-autoloader
+    - run:
+        name: Update Prod SSL Certs
+        command: |
+          vendor/bin/blt humsci:update-cert prod
+
```
sherakama, line 195: Testing multiline comment. Looks good.

blt/src/Blt/Plugin/Commands/AcquiaApi.php
```diff
@@ -0,0 +1,239 @@
...
+   *   Environment to effect.
+   *
+   * @return bool|array
+   *   API Response.
+   */
+  public function getCerts($environment) {
+    if ($response = $this->callAcquiaApi("/environments/{$this->envIds[$environment]}/ssl/certificates")) {
+      return json_decode($response, TRUE);
+    }
+    return $response;
```
sherakama, line 169: This seems trixy. Wouldn't it be more clear if the last return is `return FALSE` instead of relying on the return from the method call?

blt/src/Blt/Plugin/Commands/AcquiaApi.php
```diff
@@ -0,0 +1,239 @@
...
+        'urlAuthorize' => '',
+        'urlAccessToken' => 'https://accounts.acquia.com/api/auth/oauth/token',
+        'urlResourceOwnerDetails' => '',
+      ]);
+
+      // Try to get an access token using the client credentials grant.
+      $accessToken = $provider->getAccessToken('client_credentials');
+    }
+    catch (\Exception $e) {
+      echo $e->getMessage();
```
sherakama, line 216: Echo? Why not use `$this->say()`?

blt/src/Blt/Plugin/Commands/AcquiaApi.php
```diff
@@ -0,0 +1,239 @@
...
+    );
+
+    // Send the request.
+    $client = new Client();
+    try {
+      $response = $client->send($request, $options);
+      return (string) $response->getBody();
+    }
+    catch (GuzzleException $e) {
+      echo $e->getMessage();
```
sherakama, line 234: Echo? Why not use $this->say()?

blt/src/Blt/Plugin/Commands/AcquiaApi.php
pookmish: because this is a bare bones class without all the robo/blt stuff. I wasn't quite sure what might be the best for this

blt/src/Blt/Plugin/Commands/HumsciServerCommands.php
```diff
@@ -200,6 +147,67 @@ public function humsciLetsEncryptAdd($environment = 'dev', $options = ['domains'
...
+   *
+   * @param string $environment
+   *   Environment to update cert.
+   *
+   * @command humsci:update-cert
+   *
+   * @throws \Robo\Exception\TaskException
+   */
+  public function updateCert($environment) {
+    $cert_name = $environment == 'test' ? 'swshumsci-stage.stanford.edu' : "swshumsci-$environment.stanford.edu";
```
sherakama, line 184: How fun. 🙃

blt/src/Blt/Plugin/Commands/AcquiaApi.php
sherakama: Oh, I see you're not extending anything. Why not extend the Robo tasks?

pookmish: i suppose extending robo would be better than extending blt. that sounds good

blt/src/Blt/Plugin/Commands/AcquiaApi.php
pookmish: yeah i suppose it would be the same result. just easier to follow.

blt/src/Blt/Plugin/Commands/AcquiaApi.php
pookmish: oh now i remember. by doing `say()` it doesn't actually output the response. thats why i used echo. I dont know why it doesnt

sherakama: The output should probably use the api as robo has a `stop-on-error` option. Perhaps `printError()` https://github.com/consolidation/Robo/blob/1.x/src/Log/ResultPrinter.php#L64

PR #412, opened 2019-10-02
HSD8-721 Fix validation for date ranges when in ECKs

pookmish:
# READY FOR REVIEW

# Summary
- Fixed date range field validation when the field is used in an ECK on a node form.

# Need Review By (Date)
- 10/8

# Urgency
- medium

# Steps to Test
1. checkout this branch
1. `blt drupal:sync --site=archaeology --partial`
1. on `/admin/structure/types/manage/hs_event/form-display` display the "Conference Sessions" field and use an "Inline Entity Form (Complex)" widget setting
1. create an event
1. in the "Conference Sessions" area enter different values for the session date field.
1. validate that it doesnt give errors when the date & times are valid.

# See Also


sherakama (approved):
The code looks good. I will do an empirical review later.

boznik (approved):
This is working as expected

PR #434, opened 2019-12-10
STN-56:  sass grunt build

silverli:
# READY FOR REVIEW

# Summary

STN-56: experiments with sass grunt build for NEW humsci_basic theme

- research: compiling scss to css works with the current grunt task in the humsci_basic theme
-research: compiling nested and imported scss to css works with the current grunt task in the humsci_basic theme
- research: compiling and using decanter mixins from scss to css works with the current grunt task in the humsci_basic theme
- feat: new scripts to package json to run in the cli
  - one for `watch`
  - one for `build:css` for building sass to css and postcss compilation too
- todo: adds todo about broken postcss task

# Steps to Test
1. Try out new npm tasks: `npm run watch` and `npm run build:css` (note that the latter will not completely work since the postcss task isn't working)

# See Also

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -45,6 +45,9 @@ module.exports = function (grunt) {
         ]
       },
       dist: {
+        // TODO: edit this glob to be: 'docroot/**/humsci/**/*.css'
+        // so that it includes all the css files.
+        // Note: be ready for a large number of file changes.
         src: [
           'docroot/**/humsci/*.css'
         ]
@@ -92,6 +95,4 @@ module.exports = function (grunt) {
 
   grunt.registerTask('default', ['availabletasks']);
   grunt.registerTask('compile', ['sass:dist', 'postcss:dist']);
-
-  grunt.registerTask('default', ['availabletasks']);
 };
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
     "test": "tests"
   },
   "scripts": {
+    "watch": "grunt watch",
+    "build:css": "grunt compile",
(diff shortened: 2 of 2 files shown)


Gruntfile.js
```diff
@@ -83,6 +83,12 @@ module.exports = function (grunt) {
           ]
         }
       }
+     // sparkbox:multiple-themes: so for our multiple theme build, we'd have to add
+     // in a task here in this more "generic" function file
+     // both Nathan and I think it might make more sense to make use of a theme specific build
+     // task, file, script, or another Gruntfile.
+     // It also appears that the only task that works in this file really is the watch one (we fixed the postcss one).
+     // So, it doesnt seem that this is used all that frequently! (except maybe scss-> css builds)
```
silverli: additional commentary on these ideas: 
PROS of creating our own scss build system in the humsci_basic theme:

- that way we have custom commands for our multiple theme build
- could use npm tasks for our compiling rather than grunt (which seems a bit outdated, and hard to manipulate for complex things, like multiple theme scss builds)
- could import specific npm modules (e.g., decanter) so as not to break other themes' dependencies

bryanbraun (approved):

Gruntfile.js
```diff
@@ -45,6 +45,8 @@ module.exports = function (grunt) {
         ]
       },
       dist: {
+        // TODO: edit this task to work with 'docroot/**/humsci/**/*.css'
```
bryanbraun: Actually, can we add a "why" in here? Maybe like: "This glob doesn't currently include all our css files. Edit this task..."

PR #437, opened 2019-12-13
[STN-60] Multi-theme Sass build

nrambeck:
# Summary

Configures a Sass build in the Humsci Basic theme that generates different stylesheets for each of the sub-themes.

- Uses theme specific Grunt task to run Sass and PostCSS tasks on Sass files in the theme
- NPM scripts are setup to run the Grunt tasks (`npm run watch` and `npm run build:sass`)
- The build generates a stylesheet for each of the 3 themes in the `dist/css` folder
- The 3 Drupal themes are configured to load their respective stylesheets found in Humsci Basic
- linting is also set up as npm script (`npm run lint:sass`) and will be run during the watch task at each change
- sass-true tests have been set up with mocha (`npm run test:sass`)

# TODO

-  ✅ We should setup a Grunt task for linting the Sass stylesheets: https://stylelint.io/
- ✅ Setup mixin that would allow adding styles for multiple themes
- ✅ play around with a card to do Adam's version of set up with sass themes in subthemes https://github.com/SU-HSDO/suhumsci/pull/438
- ✅ set up true Sass tests

# Need Review By (Date)
_['10/30', 'asap', etc.]_

# Urgency
_['low', 'medium', 'high', etc.]_

# Steps to Test
Make some changes to the sass files in humsci_basic and see if these scripts work with them:

- `npm run build:css` should build your new sass to css and apply postCss where necessary
- `npm run lint:sass` should check your new sass for errors (add an error if you'd like)
- `npm run watch` will start watching the humsci_basic src/scss folder. Make a change and see it run.
- `npm run test:sass` make sure sass tests pass

# See Also
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)

PR #438, opened 2019-12-16
STN-60: feat:  multiple themes sass build in humsci_basic and child subtheme

silverli:
# READY FOR REVIEW

# Summary
Configures a Sass build for the Humsci Basic theme, and airy sub theme that makes use of the existing root Gruntfile tasks. I was able to import sass mixins/variable from humsci_basic into the airy sass and compile based on the use there.

- Uses Grunt to run Sass and PostCSS tasks on Sass files in the theme
- NPM scripts are setup to run the Grunt tasks (`npm run watch` and `npm run build:css` and `npm run lint:css`)
- The airy Drupal theme is configured to load its respective stylesheets

# Need Review By (Date)
_['10/30', 'asap', etc.]_

# Urgency

# Steps to Test
1. _[First testing step]_
2. ...

# See Also

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,12 +28,19 @@ module.exports = function (grunt) {
     watch: {
       css: {
         files: ['**/*.{scss,sass}'],
-        tasks: ['sass'],
+        tasks: ['sass', 'run:stylelint'], // run stylelint for humsci_basic/airy at same time after each change
         options: {
           interrupt: true
         }
       }
     },
+    run: {
+      stylelint: {
+        exec: 'npm run lint:css'
+      }
+    },
+    // if we want to use postcss, might just have to do it for our own project right now
+    // or specific gruntfiles AND package.jsons in each humsci_basic, airy, etc directory
     postcss: {
       options: {
         map: true, // inline sourcemaps
@@ -92,6 +99,7 @@ module.exports = function (grunt) {
   grunt.loadNpmTasks('grunt-sass');
   grunt.loadNpmTasks('grunt-contrib-watch');
   grunt.loadNpmTasks('grunt-postcss');
+  grunt.loadNpmTasks('grunt-run');
 
   grunt.registerTask('default', ['availabletasks']);
   grunt.registerTask('compile', ['sass:dist', 'postcss:dist']);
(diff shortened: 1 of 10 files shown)


Gruntfile.js
```diff
@@ -28,12 +28,19 @@ module.exports = function (grunt) {
     watch: {
       css: {
         files: ['**/*.{scss,sass}'],
-        tasks: ['sass'],
+        tasks: ['sass', 'run:stylelint'], // run stylelint for humsci_basic/airy at same time after each change
```
silverli, line 31: So I continued using this main or root gruntfile because having a gruntfile in each directory felt a bit tedious...

docroot/themes/humsci/humsci_basic/scss/base.scss
```diff
@@ -0,0 +1,10 @@
+@mixin unbuttonize() {
+  padding: 0;
+  margin: 0;
+  border: 0 none;
+}
+
+.horizontal-card {
```
silverli, line 7: this is example of how something might be overriden by the airy theme

docroot/themes/humsci/humsci_airy/scss/_horizontal-card.scss
```diff
@@ -0,0 +1,10 @@
+//Able to import base styles/mixins/variables from humsci_basic here
+@import '../../humsci_basic/scss/base.scss';
```
nrambeck, line 2: While working on the Decanter import card, I discovered how import paths work, so that this import would not have to be relative like this. Just an FYI though. Nothing actionable.

nrambeck (commented):
This looks like a mostly workable solution. The only thing we might have a problem with is pulling in the latest version of Decanter without affecting the existing themes that use the 3.x version.

The 2 downsides I see are:

1. The styles for each theme are in separate locations, making it harder to work on individual components across each theme at the same time.
1. The CSS overrides add some extra bloat of additional rules for each sub-theme. We would have some of that using the brand/theme mixin method, but not as much.

silverli:
I chatted with @nrambeck and Adam about the two approaches we we're trying for our sass builds. 

Adam said: 
> for this PR: Pros: less ceremony, Cons: lack of context when editing/writing styles
> for the other PR: Pros: lots of context when editing and writing styles, it's familiar to Sparkbox and we know it works for multi-brands/themes

Nathan and I both prefer the sass build in https://github.com/SU-HSDO/suhumsci/pull/437 - it feels like a familiar way of working with multiple themes, we also like the single grunt file and npm scripts encapsulated in humsci_basic, rather than having to work in the root gruntfile, or have 2+ separate grunt files in the parent and child subthemes.

Additionally, separating out sass here (in this PR) feels cumbersome to us -- as though it could be hard to track where styles are coming from.

In sum, we are going with https://github.com/SU-HSDO/suhumsci/pull/437 for now!

PR #443, opened 2020-01-08
docs: updates lando readme for set up

silverli:
# READY FOR REVIEW

# Summary
- adds more information about local set up with lando

# See Also


bryanbraun (approved):
I'm not sure if we'll want to maintain this full list localhost entries as it gets really large, but this seems good to me for now.

PR #444, opened 2020-01-10
 feat(STN-83): news + horizontal card work

silverli:
**In progress thoughts...**
- trying to think through whether we put styles for horizontal card pattern under patterns/ in humsci_basic or the src/scss folder
- as of right now, src/scss seems easier, but we may want to reconsider if patterns start to bloat that folder and then move them

# Summary
[JIRA card](https://sparkbox.atlassian.net/secure/RapidBoard.jspa?rapidView=80&modal=detail&selectedIssue=STN-83)

Creates the styles and HTML for the horizontal card and News Item content type. We are not worrying about events in this context yet.

Additionally, I created some SCSS functions and mixins for looking up the primary, secondary, and tertiary colors for the colorful-teal theme variation. Also, added sass-true tests.

Design: https://www.figma.com/file/toNZnFLadtpNpG5dPzNGWz/%E2%9A%99%EF%B8%8F-Colorful-Theme---Working?node-id=1408%3A1313

# Steps to Test
1. Add a news test page to your local drupal:
- On an existing site (using the colorful theme), create a few pieces of dummy “News Item” content. Remember to add some without images, one without an external link, etc.
- Create an empty “basic page.”
- On the basic page, add a paragraph:
  - View → News
    - Display: → List
- Ensure that the news items are being displayed on the page and the default looks good
(Note that layout is not created yet, so everything will be 100%)
2. inside humsci_basic theme folder, run `npm run test` and make sure everything passes
# See Also
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)
- [Sparkbox PR Checklist](../docs/SparkboxPRChecklist.md)

PR #447, opened 2020-01-21
Add the three_column_w_image template & basic styles to humsci_basic

bryanbraun:
# READY FOR REVIEW

# Summary
This brings the `three-column-w-image` layout over to the `humsci_basic` theme, and includes some basic layout styles that are applied to the page.

I used as much of the Decanter breakpoint system as I could, (everything but the largest breakpoint).

I opted for a custom flexbox layout over the Decanter one just because our designs used different sizes for the sidebar and gutters. Also, having the sidebars scale down the same rate as the main content wasn't working that well.

# Need Review By (Date)
_1/21_

# Urgency
'high'

# Steps to Test
1. Browse to the page for a News item.
2. Verify that the single-column content is centered at large screen widths, and has screen margins at small screen width.
3. Using layout-builder, add some content to the left-sidebar, save the layout, and confirm that the "sidebar + main content" layout displays appropriately across all screen widths.
4. Using layout-builder, add some content to the right-sidebar, save the layout, and confirm that the "sidebar + main content" layout displays appropriately across all screen widths.

Note: content can be added to both sidebars though that use-case isn't optimized. TBD whether this is out of scope for this issue.

# See Also


nrambeck:
All the templating stuff looks good. As far as styling is concerned, it does seem like we could reduce the number of columns sooner as we scale down the screen size. The 3 columns are really tight on medium size screens. For the sake of time, I'm going to ask @kaseybon or @silverli to review the styling.

image

bryanbraun:
@nrambeck thanks, and good point. You're right that 3-column on tablet is the most awkward situation right now. 

I think it might be easier to iterate on it once we have more accurate content in place (correct heading sizes, correct text sizes, real featured image, etc.

Also, Alex mentioned that it isn't a primary use-case. I'll check with Austin and see whether this specific thing is out of scope for this card.

_Update: I checked with Austin... since the three column configuration isn't used for news it's out of scope for this sprint. FWIW, we don't have designs for this case or ACs in the card either.)_

docroot/themes/humsci/humsci_basic/src/scss/_base.scss
```diff
@@ -21,3 +21,10 @@ html {
     padding: 3rem;
   }
 }
+
+html {
+  box-sizing: border-box;
+}
+*, *:before, *:after {
```
kaseybon: This feels really heavy handed, are we sure we don't want to adjust box-sizing as we need to? I'm mostly thinking about how it will affect performance.

kaseybon: If we do decide to keep this however can we put an empty line between 27 and 28? I'm surprised the linter didn't require that.

docroot/themes/humsci/humsci_basic/src/scss/layouts/_layouts.general.scss
```diff
@@ -0,0 +1,3 @@
+.page-width {
```
kaseybon: Lindsays work hasn't been merged yet but we decided it is best to start namespacing and prefixing classes now while we only have a few styles. The comment that discusses the naming convention can be found [here](https://github.com/SU-HSDO/suhumsci/pull/444#discussion_r368746278).

We are also going to be prefixing variables and mixins so we don't have any confusion between Decanter's SCSS and ours.

kaseybon (changes requested):

pookmish:
[Tugboat](https://tugboat.qa) has finished building the preview for this pull request!

Link:
- https://pr447-ycddapyfownxrlzkw50y9usajtuii2bh.tugboat.qa


Dashboard:
- https://dashboard.tugboat.qa/5e2734ae6ffa7bc27104cee4

Visual Diffs:
- https://dashboard.tugboat.qa/5e2734ae6ffa7bc27104cee4/visualdiffs

kaseybon (approved):

silverli:
closing this in favor of merging work with Lindsays in https://github.com/SU-HSDO/suhumsci/pull/448

PR #454, opened 2020-01-24
(feat)STN-105: vertical card news

kaseybon:
# READY FOR REVIEW

## Summary
[JIRA - STN-105](https://sparkbox.atlassian.net/browse/STN-105)

This pull request modifies the vertical card to mimic the horizontal card. The styles were refactored so that both horizontal and vertical cards will use the same base styles while using modifier classes to account for different layouts. This will also allow us to use the card classes for other card types.

## Steps to Test
1. The setup for the vertical card should be very similar to the horizontal card. I can pair with whoever reviews this card to get that setup locally.
2. Create a new basic page with 4 views: 
    -  Vertical card (display: list)
    -  Vertical card (display: featured)
    -  Horizontal card (display: list)
    -  Horizontal card (display: featured)
3. Ensure that the refactored styles do not affect the horizontal card and it still has the correct breakpoint styles.

## PR Checklist


docroot/themes/humsci/humsci_basic/patterns/horizontal-card/horizontal-card.html.twig
```diff
@@ -1,46 +1,46 @@
-{%- set attributes = attributes.addClass('hb-cmp-horizontal-card') -%}
+{%- set attributes = attributes.addClass('hb-cmp-card hb-cmp-card--horizontal') -%}
```
nrambeck: To add multiple classes, use an array type syntax.
https://befused.com/drupal/twig-addclass

docroot/themes/humsci/humsci_basic/src/scss/_main.scss
```diff
@@ -87,7 +87,7 @@
   // 6. Components
   //    Specific UI components. This is where majority of our work takes place
   //    and our UI components are often composed of Objects and Components
-  'components/pattern.horizontal-card',
+  'components/card',
```
nrambeck, line 92: Love this change!

nrambeck (changes requested):
Everything seems to work in the browser as expected. I left one todo item regarding setting classes in Twig.

The only things is that I noticed our current Vertical Card Display Mode:
http://swshumsci.suhumsci.loc/admin/structure/types/manage/hs_news/display/hs_vertical_card

Does not use a heading element, but instead uses a span. Do we need to changed this and update the configuration Yaml files? It will affect exist themes, but I'd imagine they would be broken if using a span.

Manage display | H S Sandbox 2020-01-24 13-33-23

nrambeck:
Also, we need to rebase off the `8.2.1-release` branch.

docroot/themes/humsci/humsci_basic/src/scss/_main.scss
kaseybon: Yeah, we don't need to prefix our scss partials. We will do that sometimes if the code base is small and we don't nest the partials in directories but since these are nested we don't need the prefixes.

kaseybon:
@nrambeck this is ready for review again. Regarding your comment about updating the config file I did go ahead and commit the change. Bryan asked around and it sounds like there are checks in place to prevent overriding existing sites but all new sites moving forward will use the `h2` tag.

And to be doubly sure what we're doing will not break existing sites:
@pookmish this includes a small config change to `core.entity_view_display.node.hs_news.hs_vertical_card.yml`. Am I right that this type of change config-ignored and thus won't impact existing sites?

pookmish:
@kaseybon Thats correct. Existing sites will not be affected by changes to the entity display or entity form config changes on node types. 

Keep in mind that paragraph types **ARE** tracked and synced. so any changes to a paragraph type display/form will propagate through to all sites.

nrambeck (approved):

nrambeck:
Changes look great @kaseybon. Approved from my end.

PR #468, opened 2020-02-03
feat(STN-99):  secondary nav and styling and decanter JS

silverli:
# READY FOR REVIEW

### Notes
- 👍 get Decanter secondary-nav--buttons JS working on the page. All the HTML should be in place, we may need to write a couple more styles to hide/expand things.
  -- I need to write up a description of how this is not optimal, not all that accessible, and as consumers of the design system, we need more flexibility, especially in a drupal world
  --- dont like that they insert the button element, should just rely on classes
  --- undefined button text is bad a11y and usability
  -- should be able to specify if we want to use decanter assets or not (chevron buttons)
  -- also unrelated, but decanter styles really dont work well with ours because of html font-size set to 62.5% and then all rems are different sizes and too large
 
- 👍 investigate if it is bad to override the main menu block with secondary nav template? can we try renaming and seeing if it uses the right block? (need to write up information about all this)
  -- by default, in the config, it is defined that for a basic page, as well as basically all other content types and nodes, that the layout will use `suggestion: menu_block__main` for layoutbuilder for menu blocks. Thus, by default, that is the block template we overrode for secondary nav. The config can be viewed here: `docroot/modules/humsci/hs_basic_page/modules/hs_basic_page_display/config/install/core.entity_view_display.node.hs_basic_page.default.yml`.
- 👍 adjust heading on drupal side to not be link? to work better with link?
 -- style adjusted for when heading might be a link
 -- should also work when it's just a heading
- 👍 try with 4 levels.
- 👍 fix hover color for left bar

## Summary
Designs: https://www.figma.com/file/toNZnFLadtpNpG5dPzNGWz/%E2%9A%99%EF%B8%8F-Colorful-Theme---Working?node-id=2024%3A1360

Adds HTML markup and styling for secondary nav. This should work for as many levels of nav as there might be. We've tested to 4 levels. Uses decanter JS for secondary nav accordions.

Screen Shot 2020-02-04 at 11 01 18 AM

## Steps to Test
1. _[First testing step]_
2. ...

## PR Checklist


silverli:
## Screenshots thus far

Screen Shot 2020-02-03 at 1 04 57 PM
Screen Shot 2020-02-03 at 1 07 49 PM
Screen Shot 2020-02-03 at 1 14 57 PM
Screen Shot 2020-02-03 at 1 16 37 PM
Screen Shot 2020-02-03 at 4 15 03 PM

docroot/themes/humsci/humsci_basic/src/scss/components/_secondary-nav.scss
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    &:hover,
+    &:focus {
+      &::before {
+        color: darken($su-color-driftwood, 50%);
+      }
+    }
+
+    &[aria-expanded="true"] {
```
kaseybon, line 135: I love targeting aria attributes for these types of styles!

docroot/themes/humsci/humsci_basic/templates/menus/macros/secondary-nav-menu.twig
```diff
@@ -5,36 +5,40 @@
...
     {% if item.below is not empty %}
       {% set list_attributes = list_attributes.addClass(class_prefix ~ "__item--parent") %}
+      {% set list_attributes = list_attributes.addClass("su-secondary-nav__item--parent") %}
     {% endif %}
     {% if item.in_active_trail == true %}
       {% set list_attributes = list_attributes.addClass(class_prefix ~ "__item--active-trail") %}
+      {% set list_attributes = list_attributes.addClass("su-secondary-nav__item--active-trail") %}
     {% endif %}
-    {% if item.is_active == true %}
+    {# {% if item.is_active == true %}
```
kaseybon: Should we delete these comments?

kaseybon: Same thing on lines 29 and 41.

kaseybon (changes requested):
Everything looks good except we don't have any focus states for the show/hide toggles.

docroot/themes/humsci/humsci_basic/templates/menus/macros/secondary-nav-menu.twig
silverli: yep thank you

kaseybon (approved):

PR #501, opened 2020-02-26
STN-142 : Style Drupal Admin Links

heatherdesigns:
# [JIRA STN-142](https://sparkbox.atlassian.net/browse/STN-142) READY FOR REVIEW

## Summary
Style Drupal admin links.

Design reviewed and approved by Merani ✅

## Need Review By (Date)
3/2/2020

## Urgency
low

## Steps to Test
1. In the CLI run `npm run test` and verify our tests still pass.
2. Go to http://economics.suhumsci.loc and review the Drupal admin links located below the masthead. There are no designs for this. Reference the first comment in [JIRA STN-142](https://sparkbox.atlassian.net/browse/STN-142) for a screenshot with the request to implement something similar.
3. Go to http://economics.suhumsci.loc/node/1/layout and confirm the admin bar looks good there as well.

drupal admin

## PR Checklist
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)
- [Sparkbox PR Checklist](../docs/SparkboxPRChecklist.md)

PR #502, opened 2020-02-26
fix(STN-200): add contextual link hack

silverli:
# READY FOR REVIEW

## Summary
[Jira Card](https://sparkbox.atlassian.net/browse/STN-200)

Adds hack so that contextual links don't overlap.

## Steps to Test
1. Look at a page with paragraph types on it
2. Check that the contextual links don't overlap.
3. Make sure `npm run test` passes.

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/src/scss/_main.scss
+++ b/docroot/themes/humsci/humsci_basic/src/scss/_main.scss
@@ -111,7 +111,6 @@
   'elements/base',
   'elements/forms',
   'elements/tables',
-  'elements/wysiwyg-text-area', // WYSIWYG editor
 
   // =====================================================================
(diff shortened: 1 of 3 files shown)


heatherdesigns (commented):
Works as expected! Great job!

clipping chick

I wonder if `components` is the right place for this? It isn't a component that you would see on the site but rather something that we are customizing in the admin panel. I was unsure of where to put the style changes I made to the Drupal admin panel in STN-142 #501 and chose `elements` with a comment. It feels like this and the Drupal admin links could go into their own section. Perhaps we should consider a sub-section like `6b. Drupal admin components`.  🤔 What do you think?

heatherdesigns (approved):

PR #506, opened 2020-03-02
STN-54: Main Navigation

kaseybon:
# READY FOR REVIEW

## Summary
This pull request adds behavior changes to the main navigation and page layout:
- The sidebar navigation is hidden at small-medium screen sizes but visible at the large breakpoint
- the mobile nav will include all pages available at small-medium screens
- the desktop nav will only include the top 2 levels of pages available at large screens

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. Build both CSS and JS
2. Run `npm test` to confirm there are no errors
3. Adjust the main navigation to show 5 levels instead of the default 2:
    `Structure > Blocks Layout` and then configure the Main Navigation block in the `menu` region.
4. Confirm that all the navigation meets all acceptance criteria (link in the Jira card)

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.three_column_w_image.scss
```diff
@@ -8,6 +8,10 @@
 
   &.hb-page-width {
     @include hb-page-width;
+    // scss-lint:disable ImportantRule
```
silverli: I've added this in my PR, so I think you can remove all the new green stuff :) it will all be part of the hb-page-width mixin now

docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.three_column.scss
```diff
@@ -8,27 +8,29 @@
 
   &.hb-page-width {
     @include hb-page-width;
+    // scss-lint:disable ImportantRule
```
silverli: I've added this in my PR, so I think you can remove all the new green stuff :) it will all be part of the hb-page-width mixin now

docroot/themes/humsci/humsci_basic/src/scss/components/_main-nav.scss
```diff
@@ -26,5 +16,89 @@
...
+      }
+    }
+  }
+
+  &__menu-container {
+    padding-left: 20px;
+    margin-top: 8px;
+    margin-bottom: 8px;
+
+    &[aria-hidden="true"] {
```
silverli: do we need some noJS fallbacks for aria-hidden classes that may or may not happen if JS is turned off?

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
...
+  for (let i = 0; i < togglers.length; i += 1) {
+    const toggler = togglers[i];
+    const togglerID = toggler.getAttribute('id');
+    const togglerContent = document.querySelector(`[aria-labelledby="${togglerID}"]`);
+    const togglerParent = toggler.closest('.su-main-nav__item--parent');
+    const shouldBeOpen = togglerParent.classList.contains('su-main-nav__item--active-trail') ? true : false;
+    let isHidden;
+
+    const collapseMenu = () => {
+      togglerContent.setAttribute('aria-hidden', true);
```
silverli: I am on the fence with the aria-hidden stuff. technically, I think you should never hide a tabbable element (such as the a tags under the ul). I think it would be best to apply display: none; instead. we could add a class and then do that in the styles

silverli: I think what happens is that even if you set aria-hidden to true, you can still tab around, but nothing will be said, which is not a great experience, and can be very confusing for screen reader users

docroot/themes/humsci/humsci_basic/src/scss/components/_main-nav.scss
silverli: you can use the .no-js class that will be on the top html element too

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
```diff
@@ -1,6 +1,6 @@
 {#
 /**
- * Macro for creating nested menus.
+ * Macro for creating nested menus in the main navigation.
```
silverli, line 3: in `docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig` can you remove the 'no-js' class that is getting added?

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+const mainMenu = document.querySelector('.su-main-nav__menu-lv1');
+const mobileNavBreakpoint = 992;
+let windowWidth = window.innerWidth;
+let isHidden;
+let wasDesktopSize;
+
+const collapseNav = () => {
+  menuToggle.setAttribute('aria-expanded', false);
+  mainMenu.setAttribute('aria-hidden', true);
+  menuToggle.innerHTML = "Menu";
```
silverli: we actually dont have to set the text at all here. If we keep it "Menu" in the menu block code template, then that content can be translated by drupal if we ever need. Also, a screenreader user will hear, with the aria-expanded attribute: Menu Button Expanded / Menu button collapsed and that's all they really need to know!

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+  isHidden = true;
+}
+
+const showNav = () => {
```
silverli: I tried out a DRY-er function for these show/CollapseNav functions, as well as the ones in nested-toggler:

if you want to make a new file called changeNav.js in this same main-nav directory, add the following:


```
const changeNav = (toggle, parent, isOpen) => {
  toggle.setAttribute('aria-expanded', isOpen);
  parent.setAttribute('aria-hidden', !isOpen);
// just set isHidden variable in the local file underneath the call to this function
}

export default changeNav;

```

and then to import it and use it in **this file** you can do:

```
import changeNav from './changeNav';


// rather than calling showNav or collapseNav, you can just do:

 // Toggle the nav when the the button is clicked
  menuToggle.addEventListener('click', () => {
    if (isHidden) {
      changeNav(menuToggle, mainMenu, true);
      isHidden = false;
    } else {
      changeNav(menuToggle, mainMenu, false);
      isHidden = true;
    }
  });
```

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+  menuToggle.setAttribute('aria-expanded', true);
+  mainMenu.setAttribute('aria-hidden', false);
+  menuToggle.innerHTML = "Close";
+  isHidden = false;
+}
+
+if (menuToggle) {
+  // If below a certain breakpoint, hide the nav on page load
+  if (window.innerWidth < mobileNavBreakpoint) {
```
silverli: I think this should already be set up by default in the template code and styles. I don't think we should use JS to accomplish it. I would use no-JS and .js styles to do so instead in the styles.

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+
+const showNav = () => {
+  menuToggle.setAttribute('aria-expanded', true);
+  mainMenu.setAttribute('aria-hidden', false);
+  menuToggle.innerHTML = "Close";
+  isHidden = false;
+}
+
+if (menuToggle) {
+  // If below a certain breakpoint, hide the nav on page load
```
silverli, line 11: this is more like if there is JS on the page, then collapse the nav. we should set aria-expanded to true on the button trigger in this template code `docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig` by default on page load, and then this JS can come in if there is JS enabled on the page

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
silverli: oh i see.. you do set it to display:none;  still feels like you could add a class instead of aria-hidden since it's not doing much  when the display CSS is used

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+      collapseNav();
+    }
+  });
+
+  // Handle the showing/hiding of the nav when resizing the browser
+  window.addEventListener('resize', () => {
+    windowWidth = window.innerWidth;
+
+    // When resizing from mobile to desktop, show the navigation
```
silverli: can you add this comment here: 
```
// When resizing from mobile to desktop, ensure navigation is displayed, not hidden
// If wasDesktopSize is false, it means we haven't gotten there yet and will to run this check
// Otherwise, if wasDesktopSize is true, we are above the mobileNavBreakpoint and don't need to keep showingNav
```

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+    if (windowWidth >= mobileNavBreakpoint) {
```
silverli: can you then amend the conditional check to the following so showNav doesnt get continuously run over and over as we resize the screen upwards?

`if (windowWidth >= mobileNavBreakpoint && !wasDesktopSize)`

docroot/themes/humsci/humsci_basic/src/js/main-nav/menu-toggle.js
```diff
@@ -0,0 +1,56 @@
...
+      showNav();
+      wasDesktopSize = true;
+    }
+
+    // When resizing from desktop to mobile, hide the navigation
+    if (windowWidth < mobileNavBreakpoint && wasDesktopSize) {
+      collapseNav();
+
+      // This keeps the navigation from collapsing every time the screen is resized
```
silverli, line 47: can you add a comment like this one, or saying something similiar:

`// After the first time we resize to below the mobileNavBreakpoint, reset wasDesktopSize var`

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
```diff
@@ -27,9 +27,26 @@
     {% endif %}
 
     <li{{ list_attributes }}>
-      <a{{ link_attributes }}>{{ item.title }}</a>
-      {% if item.below %}
-        {{ menus.nav_menu(item.below, menu_level + 1, class_prefix) }}
+      {% if item.url.toString() is empty %}
```
silverli, line 40: can you explain this if check a bit more? I am not exactly sure how or why we decide between button or link

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
```diff
@@ -27,9 +27,26 @@
...
+        {% set link_attributes = link_attributes.addClass('hb-nested-toggler') %}
+        <a{{ link_attributes }} id="{{ item.title|replace({' ': ''}) }}{{ menu_level }}">{{ item.title }}</a>
```
silverli: should we add role=button?

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
+const togglers = document.querySelectorAll('.hb-nested-toggler');
+const mobileNavBreakpoint = 992;
+let windowWidth = window.innerWidth;
+
+if (togglers) {
+  for (let i = 0; i < togglers.length; i += 1) {
+    const toggler = togglers[i];
+    const togglerID = toggler.getAttribute('id');
+    const togglerContent = document.querySelector(`[aria-labelledby="${togglerID}"]`);
+    const togglerParent = toggler.closest('.su-main-nav__item--parent');
```
silverli: .closest isnt supported in ie11

silverli: .parentNode might work though!

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
...
+    const shouldBeOpen = togglerParent.classList.contains('su-main-nav__item--active-trail') ? true : false;
```
silverli: i find this naming a bit confusing

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
+const togglers = document.querySelectorAll('.hb-nested-toggler');
+const mobileNavBreakpoint = 992;
+let windowWidth = window.innerWidth;
+
+if (togglers) {
+  for (let i = 0; i < togglers.length; i += 1) {
```
silverli, line 8: hmm I dont see this logic happening to all my nested-togglers. Only the first one it seems 🤔 maybe we can pair on this

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
...
+    window.addEventListener('resize', () => {
+      windowWidth = window.innerWidth;
+
+      // When resizing from mobile to desktop, show the navigation
+      if (windowWidth >= mobileNavBreakpoint) {
+        collapseMenu();
+      }
+    });
+
+    // When tabbing through the navigation the previously opened dropdown closes
```
silverli: are these two event listeners for desktop only?

docroot/themes/humsci/humsci_basic/src/scss/components/_main-nav.scss
```diff
@@ -1,20 +1,10 @@
-// Override some decanter styles for Main Nav
```
silverli, line 1: I would keep some version of this comment here

docroot/themes/humsci/humsci_basic/src/scss/components/_main-nav.scss
```diff
@@ -26,5 +16,89 @@
...
+        .su-main-nav__button,
+        .su-main-nav__menu-container {
+          display: none;
+        }
+      }
+    }
+  }
+
+  &__menu-container {
+    padding-left: 20px;
```
silverli: px not rem? just wanted to check

docroot/themes/humsci/humsci_basic/src/scss/components/_main-nav.scss
```diff
@@ -26,5 +16,89 @@
...
+  }
+
+  &__button {
+    @include unbuttonize;
+    cursor: pointer;
+
+    position: absolute;
+    top: 0;
+    right: 0;
+    width: 44px;
```
silverli: px or rem?

docroot/themes/humsci/humsci_basic/src/scss/components/_masthead.scss
```diff
@@ -12,36 +13,15 @@
...
-        top: hb-calculate-rems(67px);
-      }
+    margin: 0 !important;
+    width: 100% !important;
+    z-index: $hb-z-index-main-menu;
+
+    @include grid-media-min('sm') {
+      flex: unset;
+      width: 100% !important;
+      max-width: none;
```
silverli, line 23: never seen this! interesting

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
silverli: or adding a comment would be awesome

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    // At larger screen sizes:
+    // =========================================================================
+    // All menus collapse when resizing larger than the lg breakpoint
+    window.addEventListener('resize', () => {
+      windowWidth = window.innerWidth;
+
+      // When resizing from mobile to desktop, show the navigation
+      if (windowWidth >= mobileNavBreakpoint) {
```
silverli: youll need a wasDesktopSize here too so collapseMenu doesn't keep getting called. I think the logic would go like this:


```
if (windowWidth >= mobileNavBreakpoint && !wasDesktopSize) {
  collapseMenu();
  wasDesktopSize = true;
}

```

but I wonder if you need this since we'll close the nav on desktop in the menu-toggle.js?

silverli (changes requested):
This is really great work. I know there are a lot of comments, but I think they are mostly pretty small. We can pair up in the morning too if we want to see if there are some things that can be implemented later.

The one big thing I had a concern with was that for my nested-togglers, it looked like only the first set was getting the appropriate aria attributes, and the others weren't.

I also wanted to check that we should be seeing 5 levels deep on mobile, and 2 on desktop? I am not sure I am seeing this behavior, I might be missing some nav config.

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
kaseybon: I added the role and also adjusted the JS to make sure that the spacebar will also toggle the anchor tags that are buttons.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
kaseybon: I updated the description with one config change that I had forgotten about. Let me know if this fixes things.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
kaseybon: I prefer to use aria-hidden and then target that attribute with the CSS because it helps prevent against future changes breaking things. If we rely on a class and something happens where the class is being toggled but aria attributes are not we probably won't notice right away.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
kaseybon: Yep. We don't want to close the mobile nested navs in either of these scenarios.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
kaseybon: Adding this creates a new edge case when resizing the browser. Personally I think this edge case happening at desktops isn't a huge issue but the new issue it creates is. I would suggest we skip this and if for some reason the edge case is a problem in the future we can tackle it then.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
kaseybon: We have Babel to handle this. It looks like there is an error in the secondary nav JS from Decanter that is causing our JS to not run. ~I'm going to move all Decanter imports to the bottom of the index.js for now so our JS still runs.~ That didn't do anything.

docroot/themes/humsci/humsci_basic/src/js/main-nav/nested-toggler.js
silverli: that is fine with me!! not a huge deal :)

docroot/themes/humsci/humsci_basic/templates/menus/macros/nav-menu.twig
```diff
@@ -11,7 +11,16 @@
     {# Link Attribtues #}
     {% set link_attributes = item.attributes %}
     {% set link_attributes = link_attributes.addClass(class_prefix ~ "__link") %}
-    {% set link_attributes = link_attributes.setAttribute('href', item.url|render) %}
+    {#
+      If there isn't a URL we want to set the href to #0 to prevent a page refresh
```
silverli, line 15: good catch 👏

silverli (approved):
Good work!

PR #511, opened 2020-03-05
feat(STN-195): hero integration

silverli:
# READY FOR REVIEW

## Summary
Hero integration to allow it to be full-width on a basic page. The vertical margin of the full width hero at the top of the page will be different than a hero inserted with other components.

I also added a build task and refactored how hb-full-width works.

I tested and checked that default image styles (which is a responsive media field using hero image styles) are working as expected.

J and I paired on the mobile design, hero with no image, and hero with no overlay. Screenshots below:

#### No overlay
hero with no overla

#### No image
hero without image
Screen Shot 2020-03-05 at 5 16 48 PM

## Steps to Test
1. Add a hero to the home page using the hero image option. (See screenshot below). As for the display, go to Manage Display, and ensure that it's using the three column grid and the "Full width" section width option.
2. Do some browser testing and verify it looks good.
3. Go to another basic page, and add a hero to the page as the first thing using the paragraph type. Check out manage display for that page, and toggle back and forth between using the three column grid and the "Full width" section width option and the "limited width" option.
2. Do some browser testing and verify it looks good.

Screen Shot 2020-03-05 at 12 56 49 PM

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/package.json
+++ b/docroot/themes/humsci/humsci_basic/package.json
@@ -5,11 +5,12 @@
   "main": "Gruntfile.js",
   "scripts": {
     "start": "npm run build:sass && npm run watch",
-    "watch:sass": "grunt watch",
     "build:sass": "grunt compile",
+    "build:js": "webpack",
+    "build": "npm run build:sass & npm run build:js",
     "lint:sass": "stylelint 'src/scss/**/*.scss' --config .stylelintrc",
     "test:sass": "mocha 'tests/true-sass-spec.js'",
-    "build:js": "webpack",
+    "watch:sass": "grunt watch",
     "watch:js": "webpack --watch",
     "watch": "npm run watch:sass & npm run watch:js",
     "test": "npm run lint:sass && npm run test:sass"
--- a/docroot/themes/humsci/humsci_basic/package.json
+++ b/docroot/themes/humsci/humsci_basic/package.json
@@ -7,7 +7,7 @@
     "start": "npm run build:sass && npm run watch",
     "build:sass": "grunt compile",
     "build:js": "webpack",
-    "build": "npm run build:sass & npm run build:js",
+    "build": "npm run build:sass && npm run build:js",
     "lint:sass": "stylelint 'src/scss/**/*.scss' --config .stylelintrc",
     "test:sass": "mocha 'tests/true-sass-spec.js'",
     "watch:sass": "grunt watch",
--- a/docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.three_column.scss
+++ b/docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.three_column.scss
@@ -6,7 +6,11 @@
     flex-wrap: nowrap;
   }
 
-  &.decanter-grid {
+  // If this uses the "limited width" section width
+  // option in layout builder, then apply our
+  // hb-page-width mixin. Otherwise, the section will
+  // go full width.
+  &:not(.hs-full-width) {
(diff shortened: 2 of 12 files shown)


docroot/themes/humsci/humsci_basic/package.json
```diff
@@ -5,11 +5,12 @@
   "main": "Gruntfile.js",
   "scripts": {
     "start": "npm run build:sass && npm run watch",
-    "watch:sass": "grunt watch",
     "build:sass": "grunt compile",
+    "build:js": "webpack",
+    "build": "npm run build:sass && npm run build:js",
```
heatherdesigns, line 10: :tada:

docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.three_column.scss
```diff
@@ -6,7 +6,11 @@
     flex-wrap: nowrap;
   }
 
-  &.hb-page-width {
+  // If this uses the "limited width" section width
+  // option in layout builder, then apply our
+  // hb-page-width mixin. Otherwise, the section will
+  // go full width.
+  &:not(.hs-full-width) {
```
heatherdesigns, line 13: Excellent comments! :tada:

heatherdesigns (changes requested):
This is looking great! I still need to do browser testing but wanted to share some feedback thus far.

I found a couple things to look into.

If someone adds a link within the body of the hero, it displays as a button.
style links in body

Lists are available in the WYSIWYG for the body but they do not display. Should we verify that is the expected behavior? Putting a list in the hero would be kinda weird. 🤷‍♀
fyi lists do not display

 The hero without an overly is an interesting scenario. This is a screenshot when logged out. This probably isn't notable when logged in because of the admin nav. There spacing in between the nav and hero image is awkward. And there's no space between the hero image and main content.

no_overlay

heatherdesigns:
Browser Testing:
- [x] Safari
- [x] Chrome
- [x] Chrome - Windows
- [x] Firefox - Mac
- [x] Firefox - Windows
- [x] Edge
- [x] IE

heatherdesigns (commented):
Browser testing is complete. Everything looks fantastic! Everything stacks in IE, which is perfect! It meets standards for IE. :tada:

The only thing I noticed in IE is that the image can get swishy at smaller screen sizes.

IE

another IE

---

One other thing to note. When I first opened the log in page, there's an extra space above the admin tool bar. I tried to debug it but couldn't figure it out. I'd refresh and nothing. So I opened the same page in another browser window and the extra space was gone. Super weird. I'm using Safari for my logged in state. I'm not sure if that has anything to do with it. I don't think we need to be too worried about it, but if we keep seeing it, then we should make a bug card for it.

extra

silverli:
@heatherdesigns in response to the things you noticed:

**List styles not showing up in the hero text:**
- the body is trimmed to 200 characters, so if you place lists towards the top, you'll see them :)
- also as for links showing up as buttons, I've fixed that now!
Screen Shot 2020-03-09 at 10 27 30 AM
Screen Shot 2020-03-09 at 10 43 39 AM
- I have also added better link and list styles for contrast based on Merani's guidance

**Hero image alone with extra vertical margin**
- I think I've fixed this with new styles that target when the image is alone

**Image gets squishy in IE11**
- I am not too worried. IE11 is a silver browser, and I think the functionality/appearance is good enough for that compliance. (Object-fit CSS property that we are using isn't supported in IE11... so that's why this is happening).

**Safari logged in state with weird bug**
- I wasn't able to replicate

heatherdesigns (approved):

PR #516, opened 2020-03-10
STN-215: QA Feedback Pt2

bryanbraun:
# Almost READY FOR REVIEW

## Summary
Misc QA feedback defined in https://sparkbox.atlassian.net/browse/STN-215

## Need Review By (Date)
asap

## Urgency
medium?

## Steps to Test
1. ...

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
+++ b/docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
@@ -45,12 +45,16 @@
         {% if time['#sources'] %}
           {% for field in time['#sources'] %}
             {% if field|render|striptags('<drupal-render-placeholder><a>')|trim  %}
-              {{ field }}
+              <div class="hb-card__subcontent-item">
+                {{ field }}
+              </div>
             {% endif %}
           {% endfor %}
         {% else %}
           {% if time|render|striptags('<drupal-render-placeholder><a>')|trim  %}
-            {{ time }}
+            <div class="hb-card__subcontent-item">
+              {{ time }}
+            </div>
           {% endif %}
         {% endif %}
 
--- a/docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
+++ b/docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
@@ -39,11 +39,21 @@
       </div>
     {% endif %}
 
-    {% if time|render|trim  %}
+    {% if time %}
       <div class="hb-card__subcontent">
-      {% if time|render|striptags('<drupal-render-placeholder><a>')|trim  %}
-        {{ time }}
-      {% endif %}
+
+        {% if time['#sources'] %}
+          {% for field in time['#sources'] %}
+            {% if field|render|striptags('<drupal-render-placeholder><a>')|trim  %}
+              {{ field }}
+            {% endif %}
+          {% endfor %}
+        {% else %}
+          {% if time|render|striptags('<drupal-render-placeholder><a>')|trim  %}
+            {{ time }}
+          {% endif %}
+        {% endif %}
+
       </div>
     {% endif %}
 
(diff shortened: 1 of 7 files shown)


bryanbraun:
I noticed that the separator was looking fine for the date stacked vertical card, but it was missing on the horizontal card. I think that's because those cards are still using the "Subcontent Item" class:

image

docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
```diff
@@ -37,18 +37,16 @@
...
     {% endif %}
 
-    {% if time|render|trim or location|render|trim %}
-
-      {% set general_variants = {'time': time, 'location': location} %}
-
+    {% if time|render|trim  %}
       <div class="hb-card__subcontent">
-        {% for key, variant in general_variants %}
-          {% if variant|render|striptags('<drupal-render-placeholder><a>')|trim  %}
```
silverli, line 48: do we need this piece about striptags('<drupal') etc... in there?

docroot/themes/humsci/humsci_basic/src/scss/components/_more-link.scss
```diff
@@ -0,0 +1,13 @@
+.more-link {
```
silverli, line 1: 🎉

docroot/themes/humsci/humsci_basic/src/scss/utilities/_admin.contextual-links.scss
```diff
@@ -13,3 +13,10 @@ $sel: '';
     }
   }
 }
+
+// Prevent humsci_basic link styles from applying to contextual link items.
```
silverli, line 17: 🎉

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-editor.scss
```diff
@@ -0,0 +1,6 @@
+// Prevent humsci_basic link styles from applying to editor inputs
```
silverli, line 1: 🎉

silverli (changes requested):
Everything looks good, but I have two other comments I added that that might need a change too, and I haven't done browser testing yet.

docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
bryanbraun: Yeah I believe the purpose of this was to make sure that if a content field was empty for a piece of content then the pipe separator doesn't show up.

(for example: you could imagine a list of news items where one news item doesn't have an external link but the other ones do)

So that's the case we should test here. I don't know if we need the exact `striptags('<drupal'` solution, but we should handle this somehow.

docroot/themes/humsci/humsci_basic/src/scss/components/_card.scss
```diff
@@ -67,35 +67,49 @@
...
+        font-size: hb-calculate-rems(16px);
+      }
 
-    &:last-of-type {
       &::after {
         content: '';
-        margin: 0;
-        border: 0;
+        margin: 0 hb-calculate-rems(8px);
+        border-left: 1px solid darken($su-color-driftwood, 45%);
```
silverli, line 86: can we add a comment that this is the pipe CSS

docroot/themes/humsci/humsci_basic/patterns/date-stacked-vertical-card/date-stacked-vertical-card.html.twig
silverli: yeah if you all could test for this tomorrow, that would be good to check! I am seeing behavior on my end where there is a pipe after just one item:

Screen Shot 2020-03-10 at 5 01 20 PM

docroot/themes/humsci/humsci_basic/src/scss/components/_card.scss
silverli: It also might be better to use border-right starting on the second child every time (see the bug I commented on with the screenshot above)

kaseybon: The border is only applied to the pseudo element which has no width, not the field itself. We remove the pipe from the last-child in the next line. If you're seeing an pipe when you shouldn't be it means there is an empty field in the DOM.

bryanbraun:
Rebasing this, since there's conflicts.

silverli (approved):
All the pipes with vertical/hcard/date stacked looked great! good work.

PR #518, opened 2020-03-12
March 11 fixes

pookmish:
# READY FOR REVIEW

# Summary
- Added humanitiescore multisite
- Switched the link button in the wysiwyg for the one from core (similiar to what we did for the D8CORE)
- Fixed a generic.png file missing on existing sites for media files in the media library.
- Fixed composer dependency issues and failing tests.
- added a post multisite hook for blt similar to what i implemented for ace-gryphon that modifies the drush alias and the blt.yml files

# Need Review By (Date)
- ASAP please

# Urgency
- high

# Steps to Test
1. checkout this branch
1. `composer install`
1. install a new site `blt drupa:install`
1. add a page with a text area, add an image to the wysiwyg and verify you can add a link to that media item.
1. review code for the new multisite/blt commands/drush aliases.

# See Also


docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.post_update.php
```diff
@@ -183,3 +184,18 @@ function su_humsci_profile_post_update_8202() {
       ->save();
   }
 }
+
+/**
+ * Copy the media module icon to the site's files directory.
```
sherakama, line 189: Why?

sherakama (approved):

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.post_update.php
pookmish: the directory is normally created during install time, but due to the way a site is provisioned the files might not exist.
https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/modules/media/media.install#L40

pookmish: these icons are used in the media library as the "default images" even for documents and pdfs.

PR #529, opened 2020-03-19
fix(STN-273): calc bug with IE11

silverli:
# READY FOR REVIEW

## Summary

Fixes an IE11 bug but also a more general issue with my use of `calc` on the grid.  You cannot mix certain operations in css with different units and you must be very careful about order of operations

## Steps to Test
1. See that the bug described here (using a view within a grid for 3-12 columns) no longer exists in ie11: https://sparkbox.atlassian.net/browse/STN-273
2. ensure that `npm run test` passes.

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.layout.scss
+++ b/docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.layout.scss
@@ -23,13 +23,14 @@
 // Cols over 12 are not supported.
 @mixin hb-column($col-number: 2) {
   $percentage-width: 100% / $col-number;
-  $one-less-col: $col-number - 1;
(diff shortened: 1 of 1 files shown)


docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.layout.scss
```diff
@@ -55,7 +56,7 @@
 
       // above 3 cols
       @include grid-media('lg') {
-        @include hb-column-width($percentage-width, $total-margin, $col-number);
+        @include hb-column-width($percentage-width, hb-calculate-rems($total-margin), $col-number);
```
heatherdesigns, line 59: Good catch adding `hb-calculate-rems`! :tada:

heatherdesigns (approved):

PR #536, opened 2020-03-25
feat(STN-170): adds Colorful pairing theme setting to humsci_colorful

silverli:
# READY FOR REVIEW

## Summary
- adds Colorful pairing theme setting to humsci_colorful
- adds hard coded default configs to each theme for global theme
  settings

## Steps to Test
1. Ensure tests pass `npm run test`.
2. On a site such as the default site, uninstall the Humsci Colorful and Humsci Basic themes. Install Stanford Humsci Theme. Switch back and install the Humsci Basic theme (basically you need to do a fresh install). 
3. Go to your home page and check the classes that are being applied to the brandbar, lockup, and global footer. Ensure. a `--default` class is being applied to all. For the  local footer, ensure that no default class is applied.
4. Install the Humsci Colorful theme now.
5. Go to your home page and check the classes that are being applied to the brandbar, lockup, local footer, and global footer. Ensure. a `--default` class is being applied to all. 
6. Check that the class on the `html` element is `hc-pairing-ocean`.
6. Go into the theme settings for Humsci Colorful, and change the selected option for the color pairing for this theme. Check the classes on the `html` element again and make sure it switched.
7. Go into the theme settings for Humsci Basic and ensure that the color pairing setting does not exist there.

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_colorful/humsci_colorful.theme
+++ b/docroot/themes/humsci/humsci_colorful/humsci_colorful.theme
@@ -2,10 +2,23 @@
 
 /**
  * @file
- * Functions to support theming in the HumSci Basic theme.
+ * Functions to support theming in the HumSci Colorful theme.
  */
 
 use Drupal\Component\Utility\Html;
+
+/**
+ * Prepares variables for the html.html.twig template.
+ */
+function humsci_colorful_preprocess_html(&$vars) {
+
+  // Theme color pairing setting to set the way the Colorful theme will appear.
+  $color_pairing = theme_get_setting('theme_color_pairing');
+  if ($color_pairing) {
+    // Set html_attributes for html DOM element
+    $vars['html_attributes']->addClass('hc-pairing-' . $color_pairing);
+  }
+}
 
 /**
  * Implements hook_preprocess_page().
@@ -13,7 +26,7 @@ use Drupal\Component\Utility\Html;
 function humsci_colorful_preprocess_page(&$vars) {
   // Variant setting for the local footer.
   $lfv = theme_get_setting('local_footer_variant_classname');
-  if ($lfv !== "none") {
+  if ($lfv) {
     $vars['local_footer_variant_classname'] = 'hb-local-footer--' . $lfv;
   }
 }
--- a/docroot/themes/humsci/humsci_colorful/theme-settings.php
+++ b/docroot/themes/humsci/humsci_colorful/theme-settings.php
@@ -16,13 +16,31 @@ $theme_name = \Drupal::theme()->getActiveTheme()->getName();
  * Form override for theme settings.
  */
 function humsci_colorful_form_system_theme_settings_alter(array &$form, FormStateInterface $form_state) {
+  // Colorful theme color pairing setting
+  // theme_color_pairing
+  $form['options_settings']['humsci_colorful_color_pairing'] = [
+    '#type' => 'fieldset',
+    '#title' => t('Color Pairing'),
+  ];
+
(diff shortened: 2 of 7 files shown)


docroot/themes/humsci/humsci_colorful/theme-settings.php
```diff
@@ -16,13 +16,31 @@
...
+      'ocean' => t('Ocean'),
+      'mountain' => t('Mountain'),
+      'stanford' => t('Stanford'),
+    ],
+    '#default_value' => theme_get_setting('theme_color_pairing'),
+  ];
+
   // Local Footer
-  $form['options_settings']['humsci_basic_local_footer'] = [
+  $form['options_settings']['humsci_colorful_local_footer'] = [
```
bryanbraun, line 38: Good catch!

docroot/themes/humsci/humsci_colorful/humsci_colorful.theme
```diff
@@ -2,11 +2,26 @@
...
+function humsci_colorful_preprocess_html(&$vars) {
+
+  // Theme color pairing setting to set the way the Colorful theme will appear.
+  $color_pairing = theme_get_setting('theme_color_pairing');
+  if ($color_pairing !== "none") {
+    $vars['theme_color_pairing'] = 'hb-colorful hb-colorful--' . $color_pairing;
+
+    // Set html_attributes for HTML DOM element
+    $vars['html_attributes']->addClass($vars['theme_color_pairing']);
+  }
```
bryanbraun:
```suggestion
  $vars['html_attributes']->addClass('hb-colorful');
  
  // Theme color pairing setting to set the way the Colorful theme will appear.
  $color_pairing = theme_get_setting('theme_color_pairing');
  if ($color_pairing) {
    $vars['html_attributes']->addClass('hb-colorful--' . $color_pairing);
  }
```

Here are a few adjustments that seem to make sense to me (untested!):

1.  Pull the `hb-colorful` class addition outside of the color pairing condition, since I think that class should always be there, regardless of the color pairing situation.
2. Change the condition from `if ($color_pairing !== "none")` to just `if ($color_pairing)`. I'm not sure how the value could ever equal the string `"none"`. I noticed that some other code we have does this check (comparing to `"none"`), but I suspect something's wrong there...  [the docs suggest that `theme_get_setting()` will return `NULL` if the setting doesn't exist](https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/theme_get_setting/8.2.x).
3.  I don't think we need to create a `$vars['theme_color_pairing']` variable, since we aren't using it in our template, so we can just add the class to `$vars['html_attributes']` directly instead.

bryanbraun (changes requested):
Functionally this is working great! I have a couple code suggestions for the preprocess hook, if you think they make sense.

docroot/themes/humsci/humsci_colorful/theme-settings.php
```diff
@@ -16,13 +16,31 @@
...
+    '#title' => t('Color Pairing'),
+  ];
+
+  $form['options_settings']['humsci_colorful_color_pairing']['theme_color_pairing'] = [
+    '#type' => 'select',
+    '#title' => t('Color Pairing'),
+    '#options' => [
+      'ocean' => t('Ocean'),
+      'mountain' => t('Mountain'),
+      'stanford' => t('Stanford'),
```
silverli: change to sand

docroot/themes/humsci/humsci_colorful/humsci_colorful.theme
silverli: sounds great! thank you for the feedback

bryanbraun (approved):

PR #570, opened 2020-04-20
STN-157 : Date Stacked Horizontal UI Pattern

heatherdesigns:
# [STN-157](https://sparkbox.atlassian.net/browse/STN-157) READY FOR REVIEW

## Summary
This PR creates a `Date Stacked Horizontal` card designed to be used specifically for events.

This PR has been browser tested ✅, accessibility tested ✅, and reviewed by a designer ✅.

### 📝FYI 
The label style for the Date Stacked Horizontal UI Pattern is under design consideration by H&S. There may be an update to the color of the label.

## Need Review By (Date)
4/22/2020

## Urgency
medium

## Steps to Test
1. In the CLI, run `npm run test` and confirm our Sass tests pass.
1. Because this PR brings in a new pattern, you need to clear the cache by running `lando drush @economics.local cr`. This allows Drupal to recognize the new pattern.

### Add a new view
1. Go to the H&S/Sparkbox shared Google Drive Development directory and download an updated version of the `Custom Econ Events` view.
1. Go to Configuration > Development > Configuration Synchronization > Import
1. Click on the `Single item` tab, under `Configuration type` select `View`, paste settings from `custom-econ-events.txt`, and click `Import`.

### Create a basic page with the new view set to use the Date Stacked Horizontal pattern
1. Create a `Basic Page`
    * Content > Add Content > Basic Page
1. Add a `View` set to `Custom Econ Events` with a `Display` of `Date Horizontal - Single`
1. Save and check out your new page!

### Confirm that you can change the label styles
1. Edit the `Custom Econ Events` view `Date Horizontal - Single` display.
1. Under `Fields`, click on `Content: Speaker (Speaker)`
1. Under `Style Settings`, check `Customize field and label wrapper HTML`
1. Check `Create a CSS class` and add `hb-highlighted-label`
1. Apply and Save

## PR Checklist


kaseybon (changes requested):
Everything looks good except the linter is failing.

bryanbraun:
Quick question on this, @heatherdesigns... what's the fallback behavior for when there's no image?

I was reminded of this when looking at one of the in-progress traditional designs below. These styles don't apply to the colorful theme, but I think the fallback behavior we see here is what they're hoping to achieve: 

image

docroot/themes/humsci/humsci_basic/src/scss/components/_card.scss
```diff
@@ -86,6 +86,67 @@
...
+
+      .field-content {
+        display: block;
+        font-size: hb-calculate-rems(16px);
+        padding: 0 hb-calculate-rems(4px) 0 hb-calculate-rems(28px);
+        line-height: 127%; // keep icons from being cropped at smaller screen sizes
+      }
+
+      // only apply calendar icon when the label exists
+      .views-label-field-hs-event-date + .field-content {
```
kaseybon, line 128: We should add new classes for these icons fields (like `.hb-card__date-icon`) instead of targeting the `.views-label-` classes.

heatherdesigns: I chatted with Bryan on this to make sure I was on the right path forward. Targeting the Drupal `.vews-label-` classes is fine but they needed to be scoped inside of a class specifically for the stying of the icons.

I created a `hb-card__icon` base class that has modifiers for `date`, `location`, and `speaker`. By pulling out the icon classes out from being nested under `hb-card__subcontent-detail` we make them available for placement in other twig templates which use the `hb-card` classes. (For future use cases.)

kaseybon (approved):

PR #580, opened 2020-04-28
STN-231: IE 11 Navigation Bugs

kaseybon:
# READY FOR REVIEW

## Summary
This pull request does 2 things:
- Refactors the secondary nav to use our custom JS and CSS
- Fixes any console errors in IE 11 (this is also the reason why we removed Decanter JS for secondary nav, it was causing errors)

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. Run `npm test`
2. Run `npm start` to build new assets
3. Review the main nav and secondary nav in a modern browser to ensure everything still works as expected at all screen sizes.
4. Review the main and secondary nav in IE 11 and ensure that it now behaves like modern browsers do.
5. Checkout the Courses page and ensure that the JS is not breaking the main navigation on the page.

## PR Checklist
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)
- [Sparkbox PR Checklist](../docs/SparkboxPRChecklist.md)

PR #578, opened 2020-04-24
STN-322 : More Link

heatherdesigns:
# READY FOR REVIEW

## Summary
This branch was created off of `sparkbox-sprint-15` with commits cherry-picked from #575 
Original work completed by @kaseybon 
Reviewed and approved by @heatherdesigns

## Steps to Test
See #575

PR #577, opened 2020-04-24
feat(STN-306): background color for Row Paragraph type

silverli:
# READY FOR REVIEW

## Summary
Adds a new field for the paragraph style on the Row paragraph type.

**Browser testing**
- [x] Chrome
- [x]  Firefox
- [x] Safari
- [x] IE11

## Steps to Test
1. In humsci_basic, run `npm run test`
2. If you have not synced the latest from econ, I would recommend doing that now, on my branch: `lando blt drupal:sync --site=economics --sync-files --partial`.
3. If you have pulled from econ recently, then you can just import the new database changes: `lando drush @economics.local config-import --partial`
4. After logging into the site, go to a basic page, and either edit or add a new Paragraph "Row" type. Make sure to check for the existence of the new select field called "Paragraph Style". Select a paragraph style and then save. Check your page to see if the changes were applied.
5. Check both limited width and full width options. Also make sure to check a row on a page with a sidebar (both options should become limited width).

## PR Checklist
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)
- [Sparkbox PR Checklist](../docs/SparkboxPRChecklist.md)

PR #576, opened 2020-04-22
STN-324 : Detail Page Image Styles

heatherdesigns:
# [STN-324](https://sparkbox.atlassian.net/browse/STN-324) READY FOR REVIEW

## Summary
The work in this PR empowers H&S to display images using different images styles when the need arises.

### Acceptance Criteria
* Identify a solution for displaying a non-resized featured-image, within the 3-column-with-image layout, that meets H&S’s needs.
* Ensure that the solution allows for selection of different image styles.
* Implement the approach and demonstrate it’s functionality on the publications detail pages.
* Ensure the approach meets our browser support and accessibility guidelines.

## Need Review By (Date)
4/27

## Urgency
medium

## Steps to Test
1. In the CLI, run `npm run test` and confirm all of our Sass tests pass.
2. Run `npm start` to build the styles. The compiled CSS files were added during a rebase. 🤷  I will rebuild and update them once everything has been reviewed and approved.
2. Review the following pages with the default image style of `Medium Square 360 x 360`
    * You can **check / update the image style** used on a detail page by going to `Structure > Content types > Event _(or whichever node detail page you are working on)_ > Manage Display`. You can also use the contextual link located on the detail page.
    * Click on `Manage layout`
    * Find the `"Image" field` block, click on the contextual link, and select `Configure` 
    * Find `Image Style` and confirm / select `Medium Square (360 x 360)`
    * Update. Save layout. Back to site.
- [ ] Event: http://economics.suhumsci.loc/events/events-all-fields-filled
- [ ] Publication: http://economics.suhumsci.loc/publications/publications-all-fields-filled
- [ ] News: http://economics.suhumsci.loc/news/joeys-test-news-item
- [ ] Person: http://economics.suhumsci.loc/people/ran-abramitzky
3. The work in this PR gives H&S the flexibility to use the image styles set in the configuration. I will be adding recommendations on which image styles to use to the Pattern Doc. Following the steps above, test various image styles: `Medium Scaled (360)`, `Small Square (200 x 200)`, etc.
4. Test another option. 
H&S requested styles for when the image is placed within the main body.
    * Go to http://economics.suhumsci.loc/publications/publications-all-fields-filled
    * Click on the contextual link and select `Manage Display`
    * Click on `Manage layout`
    * Move `"Publications Image"` into the main body section
    * Click on the contextual link for the publications image
    * Add the utility class of `hb-main-body-detail-image` to  the `Field Formatter Class`
    * Update. Save layout. Back to site.

## PR Checklist
[...]

PR #581, opened 2020-04-28
Sparkbox Sprint 15

silverli:
## Summary
Changes and stories addressed in Sparkbox Sprint 15:

- word break lockup fix | STN-315 #567 
- Add Hi-Res support to WYSIWYG image styles | STN-259 #568 
- QA Feedback: Sprint 14 | STN-319 #571 
- Main menu flash on page load | STN-289 #574 
- More link style missing in WYSIWYG | STN-322 #578 
- Full-width background color on row | STN-306 #577 
- Revisit image styles on detail pages | STN-324  #576 
- IE Navigation bugs | STN-231 #580 
- Date-stacked Horizontal Card UI Pattern | STN-157 #570 

### TASK LIST:
- [x] Add compiled CSS & JS
- [ ] Ensure Code Climate passes against the `develop` branch. Fix any errors.
- [x] Assign to Mike for review.

## Need Review By (Date)
4/28

## Urgency
high

## Steps to Test
1. `npm run test` should pass in the root folder
2. If you are interested in the testing steps for associated PRs, please take a look at the individual PR's description and "Steps to Test" section.

## PR Checklist


bryanbraun (approved):
This looks good to me.

A few notes for Mike...

1. There's a [Code-climate issue for some JavaScript duplication](https://codeclimate.com/github/SU-HSDO/suhumsci/pull/581). We looked into this as part of https://github.com/SU-HSDO/suhumsci/pull/580... it's because the click listeners for the secondary nav look similar to the ones for the main nav on mobile. We could try to consolidate these, but it feels a bit like over-abstraction. I'd like to mark this specific issue in code-climate as a "wont-fix" but I don't have that ability. Are you comfortable with marking this as a non-issue?
2. There were two platform-wide updates this sprint that can impact existing live sites. They are:
    1. Adding hi-res effects to some existing image styles (https://github.com/SU-HSDO/suhumsci/pull/568)
    2. Adding a "style" dropdown to the "Row" paragraph. H&S has agreed to make any necessary adjustments to the field-permissions on sites using themes that don't currently support this field. 

Let me know if you have any other questions. 👍

docroot/modules/humsci/hs_paragraph_types/modules/hs_row/modules/hs_row_display/hs_row_display.module
```diff
@@ -8,6 +8,20 @@
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 
+/**
+ * Implements template_preprocess_paragraph().
+ */
+function hs_row_display_preprocess_paragraph__hs_row(&$variables) {
+  $paragraph = $variables['paragraph'];
+
+  if (isset($paragraph->field_paragraph_style)) {
```
pookmish: this is not the proper way to get field data from an entity. see https://www.metaltoad.com/blog/drupal-8-entity-api-cheat-sheet for some quick references

docroot/themes/humsci/humsci_basic/humsci_basic.libraries.yml
```diff
@@ -1,3 +1,8 @@
 base:
   js:
-    src/js/build/scripts.js: {}
+    scripts/build/scripts.js: {}
+
+no-js:
+  header: true
+  js:
+    scripts/no-js.js: {}
```
pookmish: a `no-js` javascript file? this doesn't make sense

docroot/themes/humsci/humsci_basic/patterns/date-stacked-horizontal-card/date-stacked-horizontal-card.ui_patterns.yml
```diff
@@ -0,0 +1,47 @@
+date_stacked_horizontal_card:
```
pookmish, line 1: I've talked with H&S about this, all of these patterns don't have associated libraries with them. we try to package a library with the pattern. it would be worth some investigation if this is possible to compile a scss file along side these pattern definitions. see https://github.com/SU-HSDO/suhumsci/blob/develop/docroot/themes/humsci/su_humsci_theme/patterns/molecules/horizontal-card/horizontal-card.ui_patterns.yml#L44

docroot/themes/humsci/humsci_basic/scripts/no-js.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Add .js class, and remove .no-js (if JavaScript loads on the page)
+ * This allows us to apply specific styles for no-JS solutions.
+ *
+ * We load this as a standalone script so we can prioritize it to the
+ * head of the document and avoid a flash of unstyled content. See:
+ * https://stackoverflow.com/a/12410668/1154642
+ */
+document.documentElement.classList.add('js');
+document.documentElement.classList.remove('no-js');
```
pookmish: drupal core `modernizer`  library does this already. this is redundant

docroot/themes/humsci/humsci_basic/src/scss/components/_buttons.scss
```diff
@@ -13,7 +13,7 @@ a.button {
     min-height: auto;
 
     @include hb-colorful {
-      @include hb-pairing-color('color', 'secondary');
+      @include hb-pairing-color('color', 'secondary', $important: true);
```
pookmish: please avoid any `important` in  css if at all possible

docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig
```diff
@@ -28,8 +28,47 @@
...
   {% endspaceless %}
 </nav>
+
+{#
+  All menus in the main nav are expanded by default, which allows navigation if JavaScript is
+  disabled. If JavaScript is enabled, the code below will collapse the extraneous navigation.
+  We run this inline to ensure it runs immediately after the elements are added to the DOM,
+  preventing unsightly flashing on the page.
+#}
+<script>
```
pookmish: remove scripts from twig. this should live in its own js file.

pookmish (changes requested):

docroot/themes/humsci/humsci_basic/scripts/no-js.js
bryanbraun: Modernizer isn't currently being loaded in our themes, mostly because we haven't needed it thus far. However, we did want to support this one feature (js/no-js classes), so that's what this file was intended to do.

I suppose we could load core's modernizer instead of this custom implementation if you think that's more appropriate. Want us to do that instead?

(Modernizer does come with some extra features we wouldn't be using, but the file's only 4.5k, so not that big of a deal).

docroot/themes/humsci/humsci_basic/humsci_basic.libraries.yml
bryanbraun: Yeah it would make more sense to call this something like `js-class.js` if we end up keeping it. 😄

docroot/modules/humsci/hs_paragraph_types/modules/hs_row/modules/hs_row_display/hs_row_display.module
bryanbraun: Good callout and thanks for the link. We'll look into this first thing tomorrow. 👍

docroot/themes/humsci/humsci_basic/scripts/no-js.js
pookmish: yeah adding `core/modernizer` would be a preferred approach

docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig
bryanbraun: I hear what you're saying. Originally I had this script in it's own JS file, and was planning to a attach it to the menu with using a preprocess hook or twig's `attach_library()`. However, those techniques didn't give us enough control over load order to solve this specific problem.

For some background on the issue, we intentionally load our menus as expanded by default since it allows users to navigate when JS is disabled on their browser. However, this means that while the page is loading, the menus render as expanded before collapsing. H&S found this distracting and it was particularly bad on pages with lots of content, like the courses page with extra-long tables.

This inline script collapses the menus, and it's very sensitive to load order. If we load it in the header, there's nothing to collapse because menu DOM hasn't been constructed yet. If we load it near the footer, we run into the flashing issue because we have to wait for the full DOM to be rendered, plus any JS that must be downloaded and executed before it.

We've extracted all of the non-load-order-dependent menu javascript into their own JS files. The only part we're putting inline is the part that's required to collapse the menu, and we've commented the code to explain why it's needed. I feel comfortable moving forward with this exception, but if you know of a better way to accomplish this goal, then I'm open to it.

docroot/themes/humsci/humsci_basic/src/scss/components/_buttons.scss
bryanbraun: Hm. Our mixins don't use !important by default, but maybe this was a special case?

@kaseybon, I'm seeing that this change was made [here in our QA PR](https://github.com/SU-HSDO/suhumsci/pull/571/files#diff-b79f5db116264e3c9c0f5657e44f6fa2R16). Do you remember the reasoning behind this, and whether there's another way to get the specificity we need without !important?

docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig
pookmish: this is a scenario my team is very unlikely to budge on. this needs to be extracted into a javascript file and loaded. using `attach_library` might be allowed in some scenarios. proper javascript file loaded into the header should be able act upon appropriate context when its available.

pookmish: also maybe some css could help with the flash. something like 
```
.js .thing ( visibility: hidden }
```
or whatever the use case is

docroot/themes/humsci/humsci_basic/patterns/date-stacked-horizontal-card/date-stacked-horizontal-card.ui_patterns.yml
bryanbraun: Yeah, there's a few trade-offs that we'd need to make in order to move in that direction. Our current style organization prioritizes "multiple-theme maintainability" over "UI patterns being portable." It provides some benefits, but I recognize that it isn't the same convention used in Jumpstart-UI and other projects.

You're right that it's worth investigation. I've created a JIRA card for us to explore what a transition would look like and what benefits that could provide.

docroot/themes/humsci/humsci_basic/humsci_basic.libraries.yml
bryanbraun: This was removed as part of the modernizr recommendation. 👍

docroot/modules/humsci/hs_paragraph_types/modules/hs_row/modules/hs_row_display/hs_row_display.module
```diff
@@ -8,6 +8,20 @@
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 
+/**
+ * Implements template_preprocess_paragraph().
+ */
+function hs_row_display_preprocess_paragraph__hs_row(&$variables) {
+  $paragraph = $variables['paragraph'];
+  $row_style = $paragraph->get('field_paragraph_style')->value;
```
silverli: @pookmish let me know if this looks better for getting field data from an entity.

pookmish: I would still prefer `->getString()`. also this should always check if the field is there before trying to get it.

docroot/themes/humsci/humsci_basic/templates/menus/menu--main.html.twig
bryanbraun: We can't rely on the `.js` class alone since the menus still need to be expandable. We may be able to come up with another class for representing load-progress though... I'm looking into this now. 👍

bryanbraun: Ok @pookmish, I just pushed up 13b74d8, which uses another approach to address this. Let me know if that looks good to you. 👍

docroot/themes/humsci/humsci_basic/src/scss/components/_buttons.scss
kaseybon: We do try to avoid `!important` whenever possible but there are some scenarios where we do need it. We need to use `!important` when styling buttons in the dark footer variant which causes a bit of a chain reaction with the other buttons which is why the reset button here is getting `$important: true`.

bryanbraun:
Ok @pookmish, I believe we've addressed or responded to all the items you mentioned above. Please let me know if there's anything else I can do to help with this PR. 👍

docroot/themes/humsci/humsci_basic/src/scss/components/_buttons.scss
pookmish: is it possible to use greater specification of the css selector instead of using important?

bryanbraun: @pookmish, Kasey is out today. I assume she explored some of those options, but I'm not sure. I've got some meetings but I could take a pass at this later today if you want.

pookmish: yeah i'd just like to see if its possible to avoid. typically its only needed when overriding styles set with javascript. this is the only thing holding up the PR at this point.

bryanbraun: @pookmish, I did some reshuffling and was able to get this working without `!important` in ddf5a75. How does that look?

pookmish: Cool thanks for that

PR #594, opened 2020-05-08
fix(STN-295): Basic Page New Width Config

silverli:
# READY FOR REVIEW

## Summary
On a fresh site install, the layout settings for the basic node page were not set to full width, even though it appeared they were.

We never exported the default new layout builder setting for the basic page after we changed the value. 

Originally, the full width page setting was equal to null or an empty string, but in a PR a few weeks/months ago we changed it to 'hs-full-width'. 

On a fresh install, the default value was installed as '' (empty string) so no default width was selected for a basic page.

TODO - we can delete this old branch trying to fix this issue: https://github.com/SU-HSDO/suhumsci/compare/sparkbox-sprint-16...fix-295--hero-db?expand=1

## Steps to Test
1. Ensure `npm run test` passes in humsci_basic
2. Pull down this branch and do a fresh local site sync. Go to add a new basic page, and a new hero section. (DO NOT touch any layout settings). Save the page and you should see the hero take up the full width by default.

## PR Checklist

### DIFF
--- a/config/default/core.entity_view_display.node.hs_basic_page.default.yml
+++ b/config/default/core.entity_view_display.node.hs_basic_page.default.yml
@@ -20,8 +20,8 @@ third_party_settings:
         layout_id: three_column
         layout_settings:
           label: ''
-          section_width: ''
-          region_widths: ''
+          section_width: hs-full-width
+          region_widths: center
         components:
           13b1ba92-73e2-42fc-8cd2-e4743faf8899:
             uuid: 13b1ba92-73e2-42fc-8cd2-e4743faf8899


config/default/core.entity_view_display.node.hs_basic_page.default.yml
```diff
@@ -20,8 +20,8 @@ third_party_settings:
         layout_id: three_column
         layout_settings:
           label: ''
-          section_width: ''
-          region_widths: ''
+          section_width: hs-full-width
+          region_widths: center
```
silverli, line 24: maybe this center value doesnt need to be updated 🤔

bryanbraun, line 24: This is so simple!

It looks like it'll apply to new sites going forward but non necessarily existing sites. And that's probably fine. 👍

Are we pretty sure this won't negatively impact sites on the `su_humsci_theme`?

silverli: I showed it to Mike, and he's ok with it! Yes, will affect all sites, but so did our value change to 'hs-full-width' so this will be good.

But I think the center value is probably an unnecessary change since we didn't do anything with that, or change that value.

silverli: Yeah this is all we changed last time: https://github.com/SU-HSDO/suhumsci/commit/fd1e876816ba7538c6e037a9e57e2d4fa7a84131#diff-5d94d30d84cf534b949034c88ad89648

bryanbraun: I'm not too concerned about this... it looks like it's the pre-selected value anyways:

image

bryanbraun: Cool. I was looking into what we can do for existing sites, and it looks like it's just a matter of browsing to `/admin/structure/types/manage/hs_basic_page/display/default/layout` and saving the top section.

In other words, we make changes to this "default" layout even when other individual `basic_page`s  have diverged from the default.

So fixing this for existing sites should just be a one-time, manual configuration change. This may be worth showing during demo, just so their team knows how to get here. 👍

bryanbraun (approved):

config/default/core.entity_view_display.node.hs_basic_page.default.yml
silverli: This is talking about the basic page layout for ALL basic pages.

silverli: one offs may act different

PR #609, opened 2020-05-20
feat(STN-297): equal grid cards

silverli:
# READY FOR REVIEW

## Summary
Implements both Sass and JS that work together when a special wrapper class is applied in layout builder to ensure that all the cards in a grid are the same height (and the titles also have the same height).

If you add `hb-vertical-linked-card--equal` to the Components area in layout builder, this class will prompt both Sass and JavaScript in that context to make the cards the same height.

Mike was also hoping that we make our JS as extensible and re-usable as possible, so keep that in mind while reviewing.

**Browser Testing**
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] IE 11. Note: the exact same functionality is not replicated here. Card titles will be the same height, but descriptions won't be.
~~TODO: IE does not like the new flexbox rules... I need to keep looking at this~~

## Steps to Test
1. Ensure that `npm run test` passes.
2. For now this feature is only being asked for in reference for vertical linked cards. So go to a page with those, such as: http://economics.suhumsci.loc/qa/qa-cards and http://economics.suhumsci.loc/research/research-fields and do the following:
   - go into layout builder, configure the "components" section, and add the new wrapper class: `hb-vertical-linked-card--equal`
  - go back to the page you were on and check that everything looks right
  - Note: this is meant to be used on cards that all have the same amount of content (meaning there shouldnt be just one card without an image when the rest have it, etc.)
  - Also note: you may need to change up some description text link, and title length to see this really working
3. Check the same thing in the colorful theme.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.vertical-linked-card.equal-row.scss
```diff
@@ -0,0 +1,28 @@
...
+.hb-vertical-linked-card--equal {
+  // TODO breaking things in IE
+
+  // For use in paragraphs in a row:
+  @include grid-media-min('md') {
+    .paragraph-item,
+    .paragraph-item > div {
+      // Since the default values for flex are flex-grow: 1 this
+      // allows content to grow to the height of the container
+      display: flex;
```
silverli, line 12: ie is broken now because of this

bryanbraun:
Alternative class names that don't use BEM modifiers (since this isn't applied under a BEM block):

- hb-align-vertical-linked-cards
- hb-equalize-vertical-linked-cards
- hb-equal-height-vertical-linked-cards
- hb-stretch-vertical-linked-cards
- hb-stretch-align-vertical-linked-cards

Any others?

bryanbraun:
The class name change is working for me @kaseybon.

kaseybon (approved):

PR #623, opened 2020-06-09
STN-379: Sprint 17 Feedback

kaseybon:
# READY FOR REVIEW

## Summary
This PR fixes all of the issues found in the Sprint 17 QA and cleans up some unused styles.

_QA Fixes:_
- form iframes fill the entire width of a container and have a larger height
- video iframes fill the entire width of a container and maintian a 16:9 aspect ratio
- list items flow along with floated content
- bullet points and numbers no longer overlay floated images
- wysiwyg tables and table views have margin bottom
- Traditional secondary button has a more noticable hover state
- When an h2-h6 is the only element in a row well all margin is removed to "vertically center" the heading in the well
- text areas in row wells now will the width of the container even when the content doesn't
- FA email icons are hidden by default and can be added with a utility class `hb-mailto-link`
- vertical card patterns now display the author and year fields
- subcontent fields bug where a subcontent field didn't show when only one field was mapped has been fixed

_Code Cleanup:_
- styles were originally added to remove the external icon in various scss partials but are no longer necessary since we are now hiding the icon by default

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. Check that all tests are passing
1. - [x] Check that there are no external icons showing up in places that are not using the `hb-external-link` utility class.
1. - [x] Modify the vertical card on the [Publications](http://economics.suhumsci.loc/research/publications) page to map the author and publication year to the author and year fields (note: there are a few other changes required for this one, I'll export the view config and send it your way).
    Screen Shot 2020-06-09 at 9 40 08 AM
    Multiple authors should display in a comma separated list and the year should display below the author field.
1. - [x] In the ☝️ same view map a single field to the subcontent field and confirm that a single view still displays.
1. - [x] Checkout the [QA Form](http://economics.suhumsci.loc/qa/qa-form) page and confirm that the iframe fills the entire width of the content area and is 70vh high.
1. - [x] Add a video to the page and confirm that the iframe fills the entire width of the content area and maintains a 16:9 aspect ratio.
1. - [x] On the [Image Styles](http://economics.suhumsci.loc/qa/qa-image-styles) page add unordered lists and confirm that:
    - When an image is floated to the right the list items flow under the floated images
        Screen Shot 2020-06-09 at 9 44 57 AM
        Screen Shot 2020-06-09 at 9 45 03 AM
    - When an image is floated to the left the bullet point/numbers do not lay on top of the image in both the wysiwyg editor and on the front end.
1.  - [x] On the [Faculty](http://economics.suhumsci.loc/people/faculty) page the FA email icon in structured cards should not be displayed unless you add the class `hs-mailto-link` to the email fied.
1. - [x] Confirm that email icons do not show up when using a mailto link in a hero or carousel button.
1.  - [x] Create a basic page and add the following and confirm there is correct margin bottom to each of the table elements:
    - A row with table in a text area
    - A row with 2 tables table in a text area
    - A row with the courses table view (to make it easier I modified it to only show 5 table rows)
    - Another row with the courses table view
1. - [x] Create a basic page and add a series of rows with limited and full width wells. Each row should contain a single text area with a single heading tag (h2-h6 will work). When a single heading is in the row the margin on the heading is removed to appear vertically centered.
    Screen Shot 2020-06-09 at 9 52 39 AM
    Screen Shot 2020-06-09 at 9 54 28 AM
1. - [x] In the ☝️ same page adjust the text alignment for a shorter heading and confirm that the text will center and right align.
    Screen Shot 2020-06-09 at 9 53 56 AM

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/components/_card.scss
```diff
@@ -404,6 +404,19 @@
     }
   }
 
+  &__author {
+    // vertical card, list spacing: "xxxs": 10px,
+    margin-bottom: hb-spacing-width('xxxs');
+
+    ul {
+      // @include hb;
```
silverli: hm what should go here?

silverli (changes requested):
Wow! This is a ton of work. Looking really great. I noticed two issues:

For image styles floated with bullet points or numbered lists, I see that at smaller breakpoints, the lists get cut off because of overflow: hidden:
Screen Shot 2020-06-10 at 2 06 46 PM
Screen Shot 2020-06-10 at 2 06 59 PM

For the rows of limited or full width with just a heading inside of them, I am still seeing extra margin getting added below the heading because of the `field-hs-text-area` class.

Screen Shot 2020-06-10 at 3 01 49 PM

docroot/themes/humsci/humsci_basic/src/js/index.js
```diff
@@ -3,3 +3,22 @@ import './tables/scope';
 import './tables/table-pattern';
 import './navigation/index';
 import './equal-height-grid/index';
+
+
+function resizeIFrameToFitContent( iFrame ) {
+  console.log(iFrame);
```
silverli: get rid of console log

docroot/themes/humsci/humsci_basic/src/js/index.js
```diff
@@ -3,3 +3,22 @@ import './tables/scope';
 import './tables/table-pattern';
 import './navigation/index';
 import './equal-height-grid/index';
+
+
+function resizeIFrameToFitContent( iFrame ) {
```
silverli: shall we make another file for this function? Also, can you write a few comments about what it's for?

silverli: I am a bit confused since I know you added 70vh as CSS for this and I'm not sure why JS is needed

docroot/themes/humsci/humsci_basic/src/scss/components/_card.scss
kaseybon: This is and the li styles below a leftover from when we thought the multiple authors would be in a list. Merani wants them in an inline comma separated list.

docroot/themes/humsci/humsci_basic/src/js/index.js
kaseybon: This entire bit of JS needs to be deleted. I wound up taking the code over to CodePen to play around with it and realized that the JS just wouldn't work for what we wanted to do but I never deleted it here.

The `70vh` was supposed to be a fallback, instead it's now the set height of the forms. I wanted to use JS to find the height of the contents of the iframe so we could set the iframe height to match. This would work if the iframe source was hosted at the same domain as the website but since it's hosted on Google we run into cross site origin issues. Basically we can't do anything with the source content of the iframe for [security reasons](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).

kaseybon:
The spacing in the row wells has been adjusted to be the same.

For the lists with floated images I decided that the best course of action is to adjust the images when necessary. Images will now:
- have a max-width of 100% of the container at the smallest screen size
- have a max-width of 50% of the container at the small breakpoint
- have a max-width of 75% of the container at the medium breakpoint

PR #639, opened 2020-06-24
feat(STN-417):  QA Feedback for Sprint 18

silverli:
# READY FOR REVIEW

## Summary
QA Feedback for Sprint 18. Refer to QA feedback card here https://sparkbox.atlassian.net/browse/STN-417

Browser Testing:
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] IE11

## Steps to Test
1.Ensure that `npm run test` passes.
2. You'll need a page like http://economics.suhumsci.loc/qa/row-wells that has an accordion created on its own in (not in a row) and then one in row with a full width row setting.  You should add a long title like:

Regular accordion with some looooooong looooooooooong looonnng title Regular accordion with some looooooong looooong loonooiooonnnng title
- [x] check that both these accordions spans the full width of the row and that a long title never overlaps with the plus sign icon.
3. Add a custom taxonomy field to the event content type. if you've synced from staging this will be easier as they have added one already called "SITE Year" or "custm_event_site_year". All you will have to do is ensure it is visible to use. To do so go to:
http://economics.suhumsci.loc/admin/structure/types/manage/hs_event/form-display
- Make sure SITE year is not disabled and shows up under "Event tags" like so:

Screen Shot 2020-06-24 at 11 02 36 AM

- Save those changes and add a SITE year to an existing event
- Go into layout builder for that event and add 'hb-categories' for SITE YEAR
- Play around with showing or hiding the label
- [ ] Ensure nothing looks broken (fyi I wasn't able to make this perfect with the link -- there is a hover on the pseudo element)

4. Use browser stack for this next one. Use an iphone 11 with chrome and/or safari browser and open this page: http://economics.suhumsci.loc/people/faculty
- [x] check that the select does not exceed the mobile browser width, nor create a horizontal scroll. (especially if "Environmental, Resource and Energy Economics" is selected)

5. Go to http://economics.suhumsci.loc/ and ensure that for a date tile with a week day, month, and day, the styles look good when the view has the "hide empty fields" checkbox checked and unchecked.
- [x] unchecked
- [x] checked

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/utilities/_lists.scss
```diff
@@ -3,6 +3,12 @@
     div div:not([class]) {
       @include hb-category-item;
     }
+
+    & > a:not([class]),
+    & > div > a:not([class]) {
+      // TODO this isn't perfect, link hover style gets applied to the psuedo element
```
silverli: what do you think about this @kaseybon ? I couldn't think of a perfect workaround

kaseybon (approved):

PR #640, opened 2020-06-26
STN-406: IE 11 Text Overflow in Rows

kaseybon:
# READY FOR REVIEW

## Summary
This pull request fixes a bug in IE 11 where text and images in a single horizontal and vertical card in a row would overflow the width of the container and sometimes the width of the page. This was an accessibility concern because the image overflowing the side of the page was inaccessible.

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. - [x] All tests should be passing
2. - [x] All content on the [Row Wells](https://economics-dev.stanford.edu/qa/row-wells) page should be contained within the width of the page.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.row.scss
```diff
@@ -76,6 +76,10 @@
   }
 }
 
+.counted-items-1.field-hs-row-components {
```
silverli, line 81: I have one question about this scss style -- why do you add the classes on the same line, rather than adding an &.field-hs-row-components nested below? I was just curious and I want us to sync up on a standard style strategy

silverli, line 81: I would like to write a comment here too about how if there is just one item in the row, there is no need for flex

silverli (changes requested):
Looking great! one or two small comments

docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.row.scss
kaseybon: I didn't nest it since it's such a simple rule and it didn't feel necessary to break up the styles. If there were more styles or more things nested under the `.counted-items-1` class then I would have nested like you mentioned. Both ways of writing the styles are correct and give the same results, it's just a matter or preference.

docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.row.scss
```diff
@@ -76,6 +76,12 @@
   }
 }
 
+// When there is more than one item in a row it uses flexbox however when there
```
silverli, line 79: 🎉

silverli (approved):
Looks great!

PR #645, opened 2020-07-03
HSD8-825: Added event_series to event importer

joegl:
# READY FOR REVIEW

## Summary
https://stanfordits.atlassian.net/browse/HSD8-825
[I created a new test export view](https://economics.stanford.edu/events/past-events/export-series-events) with the eventSeriesTitle and eventSeriesBody fields. [Edit view link.](https://economics.stanford.edu/admin/structure/views/view/hs_economics_events_export/edit/views_data_export_4)
Added configuration to import the Event Series.

## Notes
**NOTE:** The view and importer only include the first Event Series referenced by the Event node. The D8 sites only allow one reference and I haven’t seen an event with more than one, even though the field is set for unlimited references.
**NOTE:** I did not include the Event Series image in the importer

## Potential Issues
**ISSUE?:** The hs_events_importer module can be enabled while the hs_event_series module is disabled. Is including the event series data in the events module importer an issue?
**ISSUE?:** When I first pulled the repo and before I changed anything, I got an error when doing a config:import `Unexpected error during import with operation update for migrate_plus.migration.hs_events_image_importer: The entity does not have an ID.`

## Need Review By (Date)
7/7

## Urgency
Medium

## Steps to Test
1. Pull local stack and code
2. Import configuration changes `drush config:import -y`
3. Set the event importer migration url (admin/structure/migrate/events) to https://economics.stanford.edu/events/past-events/export-series-events
4. Run an import `drush migrate-import hs_events_importer --limit=25`
5. Check event series content was created and events are properly linked

## PR Checklist

### DIFF
--- a/config/default/migrate_plus.migration.hs_events_importer.yml
+++ b/config/default/migrate_plus.migration.hs_events_importer.yml
@@ -340,4 +340,5 @@ destination:
     - field_hs_event_speaker
     - field_hs_event_type
     - field_hs_event_video
+    - field_hs_event_series
 migration_dependencies: {  }
--- a/config/default/migrate_plus.migration.hs_events_importer.yml
+++ b/config/default/migrate_plus.migration.hs_events_importer.yml
@@ -117,10 +117,6 @@ source:
       name: event_series_body
       label: 'Event Series Body'
       selector: eventSeriesBody
-    -
-      name: event_series_id
-      label: 'Event Series ID'
-      selector: eventSeriesID
   ids:
     event_id:
       type: string
--- a/config/default/migrate_plus.migration.hs_events_importer.yml
+++ b/config/default/migrate_plus.migration.hs_events_importer.yml
@@ -340,5 +340,4 @@ destination:
     - field_hs_event_speaker
     - field_hs_event_type
     - field_hs_event_video
-    - field_hs_event_series
 migration_dependencies: {  }
--- a/config/default/migrate_plus.migration.hs_events_importer.yml
+++ b/config/default/migrate_plus.migration.hs_events_importer.yml
@@ -109,6 +109,18 @@ source:
       name: image
       label: 'Image URL'
       selector: imageUrl
+    -
+      name: event_series_title
+      label: 'Event Series Title'
+      selector: eventSeriesTitle
+    -
+      name: event_series_body
+      label: 'Event Series Body'
+      selector: eventSeriesBody
+    -
+      name: event_series_id
+      label: 'Event Series ID'
+      selector: eventSeriesID
   ids:
     event_id:
       type: string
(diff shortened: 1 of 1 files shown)


pookmish:
Nice! this looks great. exactly like i envisioned it would be. 

However, this is going to conflict with their existing events that were imported. What happens with the migration mapping is that if there is no value to populate the field, it will erase anything that is in the field. So all of their existing sites have imported events, and they probably have added an event series to those sites where desired. The data coming from events.stanford.edu doesn't contain those event series data, so all those previously modified events, will be cleared of their event series info.

The easiest way to solve this is with the `overwrite_properties` in the configuration. By default all fields/properties are synced 1 to 1. By using the `overwrite_properties` you are telling migrate module to sync 1-to-1 specific fields/properties. Other fields/properties can be changed. Then what happens is the content will be imported the first time and that field/property will be ignored in future imports.

We are currently doing this with CAPx importer for the status field. It allows the status of a person to change and not be synced. I think we should take that knowledge and add it to the migration config. see [the capx importer migration config](https://github.com/SU-HSDO/suhumsci/blob/develop/config/default/migrate_plus.migration.hs_capx.yml#L183) for an example of that.

joegl:
> Nice! this looks great. exactly like i envisioned it would be.
> 
> However, this is going to conflict with their existing events that were imported. What happens with the migration mapping is that if there is no value to populate the field, it will erase anything that is in the field. So all of their existing sites have imported events, and they probably have added an event series to those sites where desired. The data coming from events.stanford.edu doesn't contain those event series data, so all those previously modified events, will be cleared of their event series info.
> 
> The easiest way to solve this is with the `overwrite_properties` in the configuration. By default all fields/properties are synced 1 to 1. By using the `overwrite_properties` you are telling migrate module to sync 1-to-1 specific fields/properties. Other fields/properties can be changed. Then what happens is the content will be imported the first time and that field/property will be ignored in future imports.
> 
> We are currently doing this with CAPx importer for the status field. It allows the status of a person to change and not be synced. I think we should take that knowledge and add it to the migration config. see [the capx importer migration config](https://github.com/SU-HSDO/suhumsci/blob/develop/config/default/migrate_plus.migration.hs_capx.yml#L183) for an example of that.

Just to clarify and make sure I understand:
- The current export from events.stanford.edu doesn't include event series
- Sites already have imported events without event series, and set the event series manually.
- Importers will set values for already pulled events unless deliberately instructed not to

The above combined with new importer means any import would wipe out the manually created event series currently set on the events. But by adding the event_series field to the overwrite_conditions, it will only set the event series value the first time, and ignore it in the future. Does that sound right?

pookmish:
Yep. That's exactly what the scenario is :+1:

joegl:
I added the field to the overwrite_properties. I'm a little confused, because I've found conflicting information on what the overwrite_properties do.

pookmish:
Oh now that i notice the `overwrite_properties` was already there, you had it right the first time. 
- https://git.drupalcode.org/project/drupal/-/blob/9.0.x/core/modules/migrate/src/Plugin/migrate/destination/EntityContentBase.php#L28
- https://git.drupalcode.org/project/drupal/-/blob/9.0.x/core/modules/migrate/src/Plugin/migrate/destination/EntityContentBase.php#L271

pookmish (approved):

joegl:
Thanks for finishing this!

PR #655, opened 2020-07-20
HSD8-853 Update default config settings

pookmish:


pookmish:
Reviewed and approved by H&S

PR #683, opened 2020-08-11
WIP feat(STN-473):  investigate masonry - 2 proposed solutions

silverli:
# READY FOR REVIEW

## Summary
I have made the case for 2 different solutions in this PR:

**Solution 1 - CSS only**
- uses modern CSS grid techniques to accomplish masonry
- however, an extreme con of this is that the rows render vertically in order
- also, this may not be accessible as most Western accessibility techiniques prefer a left to right tab order
- the extreme pro of this approach, is that it works in all browsers, and is great for performance since no JS is used

**Screenshot of SOLN 1**
Screen Shot 2020-08-11 at 6 52 34 PM

**Solution 2 - JS solution + CSS**
- uses Vanilla JS to accomplish masonry layout like was done in the prior, legacy su_humsci theme
- however, I didn't figure out how to lazyload the images like was happening in the legacy theme, but I have opted for some performance benefits by using the imagesloaded module
- TODO: get lazy load working (it uses a view pager setting) and then we can copy from Mike's JS here: /Users/lindsaysilver/Sites/suhumsci/docroot/themes/humsci/su_humsci_theme/patterns/organisms/masonry/js/masonry.js (line 37)
- TODO: Does not work well with exposed filters
- TODO: .no-js CSS needs more work
- TODO: On resize, this is buggy. I don't see this issue in the legacy theme here: https://archaeology.stanford.edu/research/publications I am not sure what is different between the two implementations

**Screenshots of Colorful:** (raised cards and not)
Screen Shot 2020-08-11 at 6 46 58 PM

Screen Shot 2020-08-11 at 6 46 32 PM

**Screenshots of Traditional (no background well, and with a well)**
Screen Shot 2020-08-11 at 6 49 23 PM
Screen Shot 2020-08-11 at 6 49 41 PM

## Steps to Test Solution 1 and 2
1. Go to a view that uses the vertical card pattern.
2. Edit that view.
3. Change the view if it's using a Grid pattern as it's "Format", to the Unordered List setting.
Screen Shot 2020-08-11 at 6 43 17 PM

4. Add two classes to the view wrapper: 'hb-masonry' and 'hb-raised-cards' (raised-cards isn't essential though!):
Screen Shot 2020-08-11 at 6 43 25 PM

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/objects/_layouts.masonry.scss
```diff
@@ -0,0 +1,65 @@
+// citing use of this codepen https://codepen.io/chriscoyier/pen/NeRNBO
+
+// SOLN 1 - CSS ONLY
```
silverli, line 3: this is soln 1. comment out the masonry JS to see this in action and the CSS below

PR #726, opened 2020-09-18
STN-544: add file and pdf icon links to hb and legacy theme

rkelbel:
# READY FOR REVIEW

## Summary
This card adds pdf and file link styling to add icons to links selected within the WYSIWYG/CKEditor when editing Drupal pages. These styles utilize the Font Awesome icons, so they should match those that would have been used with the Font Awesome library previously. This also gives the option to remove the icon from the link, by removing the link style if needed, as this was something that was difficult using the Font Awesome icon plugin on CKEditor previously.

Jira Card: [STN-544](https://sparkbox.atlassian.net/browse/STN-544)

## Need Review By (Date)
9/18

## Urgency
medium

## Steps to Test
1. Run npm test
2. Import changes to the CKEditor config files.
3. Create a basic page, (content > basic page) and add "PDF Link" and a "File Link" to the page within the WYSIWYG editor link style options. The text will need to be added as a link/anchor before the PDF and File link styles are available from the style dropdown.
image
4. Confirm the link icons/styles show up in the WYSIWYG while content editing and the option in text styles exist. These links and icons should update color with the theme color pairings (these links are using the secondary color to the pairings).
5. These icons should be able to be swapped/modified if you choose a different link style, additionally, make sure the icon goes away if you delete the link.
image
6. Confirm the styling shows up correctly on the page view after saving the page.
image
7. Confirm these options are all available with all editor entry options: (Basic HTML, Basic HTML w/o Media, Full HTML, & Minimal HTML). 
8. Additional testing options would be to add links to cards, wells, and within a spotlight.
9. Test that same page in the different appearance themes: Colorful, Traditional, and the Legacy theme.

Note: An "outline" version of the icon has been added as a class modifier but is not added to the link style dropdown, it has been added for possible future use upon request.

## PR Checklist


config/default/filter.format.minimal_html.yml
```diff
@@ -17,7 +17,7 @@ filters:
     status: true
     weight: -10
     settings:
-      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <hr> <p class> <span class> <table class> <a href hreflang data-entity-substitution data-entity-type data-entity-uuid title class>'
+      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <hr> <p class=""> <span class=""> <table class=""> <a href hreflang data-entity-substitution data-entity-type data-entity-uuid title class=" hs-button hs-button--big hs-secondary-button hs-more-link hs-external-link hs-file-link hs-pdf-link hs-private-link">'
```
heatherdesigns, line 20: I'm curious. What does this do?

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
```diff
@@ -350,3 +350,27 @@ button.hs-more-link,
   margin-bottom: hb-calculate-rems(20px);
   @include hb-more-link;
 }
+
+a.hs-pdf-link {
+  @include hb-pdf-link;
+
+  &--outline {
```
heatherdesigns, line 358: Since the outline version is not fully implemented, can you add a comment here as to what this is for? I scrolled down and see that you added comments in the other scss files. :smile:

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.links.scss
```diff
@@ -250,3 +250,57 @@
...
+    &:focus {
+      @include hb-link-background-image(true, 0.9em, left, $hb-colorful-default, $hc-colorful-pairings);
+    }
+  }
+}
+
+//File Icon Link
+@mixin hb-file-link {
+  @include hb-link--inline;
+  @include hb-pairing-color ('color', 'secondary');
```
heatherdesigns: Remove space at the end of `hb-pairing-color`.

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.links.scss
```diff
@@ -250,3 +250,57 @@
     transition: hb-transition(transform);
   }
 }
+
+//PDF Icon Link
+@mixin hb-pdf-link {
+  @include hb-link--inline;
+  @include hb-pairing-color ('color', 'secondary');
```
heatherdesigns: Remove space the end of `hb-pairing-color`.

docroot/themes/humsci/su_humsci_theme/scss/utilities/mixins/_links.scss
```diff
@@ -0,0 +1,39 @@
+@charset "UTF-8";
+
+//
+// Links mixins for utility use
+//
+
+//PDF Icon Link
+@mixin hs-pdf-link {
+  &::before {
+    content: "\f1c1";
```
heatherdesigns, line 10: Well this was quite a bit easier than adding the SVGs. 😆 Great work!!!!

heatherdesigns (commented):
This looks fantastic! Great work! 

A couple questions and something to ask / double check with the H&S team.

In the local footer dark variant, the file and pdf icons are the same color as the external and more link icons. I think they may want them to be the same color as the other icons from font awesome which is `secondary-hightlight`. See the the facebook and linkedin icons. 
dark footer icons

It looks like the text editor isn't rendering the correct colors as the theme color pairing. I looked at our sprint branch and this issue is there as well. There isn't anything you need to do about it in your work as it looks to be pre-existing. I'm not sure it has always been like this. Maybe? 🤔  Perhaps we should ask if it is an issue and if a bug card should be made? This screenshot was taken in the editor of the Colorful theme and mountain color pairing which doesn't use red.
editor doesnt reflect color pairing

And lastly, a couple things are I'm just unfamiliar and curious about. 😆 
What is the config/default/filter.format.minimal_html.yml file do?
And I saw the compiled CSS files for su_humsci_theme. I've never made changes to this theme and am curious as to what you did to compile the CSS and if there more I should know about it. Thanks!!!!

config/default/filter.format.minimal_html.yml
rkelbel: So with the minimal html formatting, the WYSIWYG will actually strip out any html tags that aren't allowed to be added.  Essentially protecting the WYSIWYG from html entry that you do not want.  Adding it to the the "allowed html" allows those classes or tags to be allowed and therefore not stripped out by the WYSIWYG upon saving.  

This is especially important if classes or tags were added using something like "Full HTML" but then a different user that doesn't have "FULL HTML" access and is set to Minimal or something similar. It will keep the WYSIWYG editor from stripping those previously added html classes or tags out of it upon that user saving the node.

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
rkelbel: Nice catch, I meant to go back and do that here as well and forgot to. 😄

docroot/themes/humsci/su_humsci_theme/scss/utilities/mixins/_links.scss
rkelbel: Thank you! 🙇‍♂️

heatherdesigns (approved):
Excellent!!

happy racoon

PR #729, opened 2020-09-24
STN-563 Refactor Javascript for individual theme use

rkelbel:
# READY FOR REVIEW

## Summary
Previously we have been compiling all Javascript into a single file (scripts.js) and then including that file into theme pages. However, we would like to start delivering individualized javascript for animations, etc. to each theme.  This work refactors our current JS to build individual files for each theme, while also importing all of our current js into those individual javascript index files delivery of those shared functions.

### Approach:
Current source JS libraries have been moved into a shared folder directory and we created two new source libraries for each theme in its own individual theme named folders. (Airy is currently receiving the colorful scripts, as mentioned in noted in the JIRA card, but the same treatment could be done for the airy theme in the future if it was needed.)
image

What you will also see changed is the `webpack.config.js` file, this was updated so it would build these new index.js files and place them in the correct `dist` folders for Drupal to use. These folders will be similar to the way our CSS is compiled, `[theme_folder]/js/index.js`. 

Additionally, you will see Drupal library and info .yml files have been updated to import the new index.js files correctly from their individual locations and no longer relying on the file `scripts.js` which should no longer exist or build when running `npm run build`.

Last but not least, we are using `import` option to import the shared js index.js into each of the theme javascript files, which takes care of those functions still being available in each theme type. You will also notice that we did not add an individual airy.js file this is because it's library file is actually using colorful.js and currently not needed

JIRA CARD: [STN-563](https://sparkbox.atlassian.net/browse/STN-563)

## Need Review By (Date)
9/28

## Urgency
medium

## Steps to Test
1. Run `npm test`
2. Clear Drupal cache
3. Run `npm run build`
4. After verifying the files in the commit, using the summary approach above for some guidance and information, you should be able to test to make sure both the build and the javascript library is working for each theme.
- To first test the build system, let's make changes to the src files.
- If you'd like to test that the scripts are loading in your themes correctly, you can add a console log to the individual javascript src theme file. Example path:`humsci_basic/src/js/traditional/traditional.js` 
I prefer a message something like this: `console.log('Hi, this is your colorful theme speaking, I wanted to let you know I am working.')` 😄 
- Once you've added that console log to the file, run `npm run build` or `npm run build:js`, this should rebuild each index file if you've altered each one.
5. Once you've added the script and run the build, reload the site and open dev tools, then click the console tab, your console should have the message printing that you added in your theme js file. (If you don't see the changes after refresh, try clearing the drupal cache.)
6. Also, confirm that the drop-down in the main menu and the standard items that use javascript on the site are still working.
(**Note:** You may see some js errors in the console from the index.js file, however, this javascript error pre-exists and unrelated to these changes. It is also present in the staging and dev environment. The error is "then" related within the `hb-card functions`.)
7. After you've confirmed the javascript is functionality, repeat the steps and change the themes between 'colorful', 'traditional' and 'airy'. Airy should give you the same message as the colorful theme as it uses the same src file.
8. Provided these steps have been successful and there are new broken functions, we should be all set with this work. 😃 

## PR Checklist


docroot/themes/humsci/humsci_basic/humsci_basic.info.yml
```diff
@@ -4,8 +4,8 @@ description: A base for visually distinctive HumSci themes.
 package: 'Humanities & Sciences'
 core: 8.x
 core_version_requirement: ^8
-libraries:
-  - humsci_basic/base
+libraries: 
+  # Libraries are loaded via sub-themes
```
heatherdesigns, line 8: Nice! I like how you added a clarifying comment rather than just deleting this.

heatherdesigns (commented):
I was able to confirm that the JS for theme Colorful and Traditional themes is working! :tada:

Unfortunately I think something is wrong with my database because when I switch to the Airy or Basic theme I get a page with an error. I checked the sprint branch and I have the issue there as well. 😔

Even though the `Airy` theme isn't used at the moment, I feel like we should go ahead and set it up. It's hard to know what the future is for `Airy` but I think there is a good chance that when it is implemented, there could be a different team in place. I think it's good to have as much foundation set as possible. Then they don't need to figure out what needs to be done or why the JS for Airy isn't working.

docroot/themes/humsci/humsci_basic/humsci_basic.info.yml
rkelbel: Thank you! 😄

kaseybon:
I pulled down the branch and all the themes are working for me, Airy looks just like Colorful and all of the JS is working. I don't think it's a huge priority to setup an Airy directory yet. When we start working on that theme we will be setup Airy JS pretty easily since the foundational work has been done here.

heatherdesigns (approved):

PR #748, opened 2020-10-28
STN-552: Bug - Views more link padding missing

rkelbel:
# READY FOR REVIEW

## Summary
This bug was found after commits took place on 9/10, the "more link" on views no longer has padding between the button and the element above.

After investigation, I found that this was due to some grid work we did where the padding below the grid rows was being removed at the tablet and above screen sizes.  This is still desired work, however, the missing padding on the bottom of grid items allows the more link to stick directly to the bottom of a hb-grid element rather than having some space.

This card adds a margin to the top of the "more link" itself when it is added to a grid view. There is margin on the cards on smaller devices, the media query for tablet screen size was added to adjust for that change to keep the spacing consistent. The spacing was chosen to align to be consistent with how the view-rows margin is on horizontal card views with "more link" attached.

## Need Review By (Date)
10/28

## Urgency
medium

## Steps to Test
1. npm run test
2. To add a "more link" to a view for testing, pick a view that has a card grid available for editing. I chose, http://sparkbox-sandbox.suhumsci.loc/people/people-grid.
3. Click "edit view" on the contextual link
4. Once on the View edit page, locate the "Pager" section of the view and edit the following options.
image
- Change "More link:" to yes and add text for the link text.
- Change the Pager to "display a specified number of items" instead of using a pager, I chose 6 items as my specific number for the grid.
- Change "link display" to custom url, I added google as my url.
- Save the View
5. Once the view has been altered and saved you should see the more link button appear at the bottom of the view, with space above it. Inspecting the page to see where the grid and the grid-item cards end as well as the more link itself, you should see the margin above apply to give separation.
6. You can also add this to pages within a paragraph for testing, for this I used the "Default News" - "HP 3 Across News" view display in the paragraph to check, you may need to follow the steps above and add the more link with custom url to the view to test the link though.
7. I added this within a Row so I could also test it within a well, which should work with no differences from the previous view test.
image
8. The margin should adapt to stay consistent on both themes and between device sizes.

Note: More link spacing should be consistent with horizontal views that have the more link added as well.

## PR Checklist


heatherdesigns (commented):
This works great when the `.more-link` is a direct sibling to `.hb-grid`. :tada: :tada: :tada:

I also tested when it is used with a view that has pagination. We should make sure padding displays for that instance as well.

more link button with pagination

As the code is written right now, the padding is placed on the button or anchor tag, if we place it on the `div` with the `more-link` class that wraps the button or anchor tag, the `margin-top` we add will collapse nicely with the `margin-bottom` of any sibling element above it. Thus ensuring that we don't add too much margin. Let me know if that makes sense. If not, I'm happy to pair and a visual explanation.

Perhaps something like this:
```
.more-link {
  .views-element-container & {
    margin-top: hb-calculate-rems(24px);

    @include grid-media-min('md') {
      margin-top: hb-calculate-rems(48px);
    }
  }
}
```

rkelbel:
> .more-link {
>   .views-element-container & {
>     margin-top: hb-calculate-rems(24px);
> 
>     @include grid-media-min('md') {
>       margin-top: hb-calculate-rems(48px);
>     }
>   }
> }

This is a very elegant solution, nice suggestion Heather!
Updating and pushing this iteration.

heatherdesigns (approved):

PR #749, opened 2020-10-28
STN-581 : Fix highlighted label to observe inline and block styles

heatherdesigns:
# READY FOR REVIEW

## Summary
This PR fixes a regression where the highlighted label was no longer inline after updates to classes, perhaps when the hb-text-reset was introduced. The regression can be seen here: https://economics.stanford.edu/people/graduate-students 

It was also noted in a backlog grooming meeting that the `inline fields` option for fields in views has never worked for labels. This PR fixes that functionality so that it works. When a user chooses to use the `inline fields` functionality, Drupal replaces `div` tags with `span` tags. By removing the `display: block` property from the `hb-field-label` mixin,  we allow the content to flow beside the label as intended by Drupal.

## Need Review By (Date)
EOD 10/29/2020

## Urgency
medium

## Steps to Test
1. In your CLI, run `npm run test` and confirm our tests continue to pass.
2. To test the inline fields functionality, go to http://sparkbox-sandbox.suhumsci.loc/people
- [x] Confirm the email, phone, office, office hours, and website labels sit above their associated content. 
    above
3. Find and click on the contextual link to edit the view. Under `Format` find `Show:` and click on `Settings`. This should open a modal for the `People list link (grouped): Row style options`.
4. Under `Inline fields` check the box for email, then click Apply. Save the view and go back to the people page.
- [x] Confirm the email label now sits inline (beside) its associated content.
    inline
5. To test the addition of `hb-highlighted-label` works with the `inline-fields` functionality, click on the contextual link to edit the view once again.
6. Under `Fields` click on `Content: Email (Email)`
7. Under `Style Settings` check `Customize label HTML`, then check `Create a class` and enter `hb-highlighted-label` in the CSS Class input field. Click Apply, Save, and go back and view the People page.
- [x] Confirm the addition of `hb-highlighted-label` only changes the text styles of the label and does not change the alignment of the label.


rkelbel (approved):
Awesome work, this was a great dive into how Drupal Views label functionality works but also the fix to allow highlighted label to work as intended.

ALT

PR #755, opened 2020-11-10
STN-602: Table heading link refinement

rkelbel:
# READY FOR REVIEW

## Summary
Hover states for linked headings on tables did not have correct hover or focus states, which were not passing accessibility.
These linked headings did already have styling written but was not applying due to specificity, I have updated that class and it should now work as it should.

Example heading issue: https://creees-prod.stanford.edu/academicscapstone-thesis/past-theses
Jira Card: [STN-602](https://sparkbox.atlassian.net/browse/STN-602)

## Need Review By (Date)
11/16

## Urgency
medium

## Steps to Test
1. npm run test
2. A good page to use for testing is: http://sparkbox-sandbox.suhumsci.loc/publications/publication-table
3. Edit the view using the contextual link in the upper right hand of the view, above the filter across the right from the title, "Default Courses".
4. Go to Format > Settings
5. In the settings check any of the checkboxes to make the heading titles Sortable, then Apply, and then save the view.
6. The table view should now have links in those heading titles for sorting.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.tables.scss
```diff
@@ -26,7 +26,8 @@
   top: 0;
   @include hb-global-color('border-color', 'white');
 
-  a {
+  a,
+  a:not([class]) {
```
heatherdesigns, line 30: I can't think of a reason why `a` and `a:not([class])` styles for table headings would be any different, so let's refactor lines 38 - 41 into one declaration for table heading links.  So that we don't have `a:not([class])` declared twice (line 30 and 38).

And I think we should flip the colors on the links. We there isn't a link on a table heading, it is white. Links should also be white, then change to `secondary-highlight` on hover. This will make the color on all table headings consistent. Links are indicated by the text underline. And the hover state darkens slightly. Let me know if that makes sense.

heatherdesigns (approved):
Looks great!! :tada: :tada: :tada:

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.tables.scss
rkelbel: I have refactored how these were laid out, that was a great idea and a good design change for the color consistency.

heatherdesigns (approved):
Fantastic! :tada: :tada: :tada:

PR #781, opened 2021-01-04
STN-605: Refinement of Text Area / WYSIWYG Block Quote Styles

heatherdesigns:
# [STN-605](https://sparkbox.atlassian.net/browse/STN-605) READY FOR REVIEW

## Summary
Refine block quote styling in the Text Area to compliment the testimonial designs.

Adds a utility class to change the block quote text to black.

Includes styles for when hb-dark-pattern is used.

Designs: [Figma](https://www.figma.com/file/RiJgL9DltLHzig4ynM34iU/H-and-S-Testimonial-Block-Quote?node-id=0%3A1)
These are the designs for the testimonial block. Block quote examples are located to the right of the `H is the Winner! 🎉` section.

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. In your CLI, run `npm run test` and confirm our tests pass.
2. To test the updated block quote styles, you can sync with staging and go to the [QA Block Quote Styling](https://sparkbox-sandbox-stage.stanford.edu/qa-block-quote-styling) page. Or create a new flexible page in your local environment and fill it will block quotes. Ideally the block quote should be tested with a variety of content surrounding it.
3. Confirm that the block quote styles are applied in the following 3 scenarios:
- [x] In the text area / WYSIWYG, when text is highlighted and the block quote button is clicked, the block quote style is applied. This is how I expect 99% of users will implement block quotes.
- [ ] In the text area / WYSIWYG, click on the `Source` button. Add a block quote element with text inside.
    ```
    <blockquote>Testing of the blockquote element only. There is nothing else inside of here! I added this via the source code in the WYSIWYG.</blockquote>
    ```
- [ ] Test the `hb-blockquote` class. Again, the in text area / WYSIWYG, click on the `Source` button and add a div with the class `hb-blockquote` with some text inside.
    ```
    <div class="hb-blockquote">A div&nbsp;with the hb-blockquote utility class. This was added via the source code. A paragraph with the hb-blockquote utility class. This was added via the source code. A div&nbsp;with the hb-blockquote utility class. This was added via the source code.</div>
    ```
4. A new utility class of `hb-blockquote-black-text` has been added. This will change the text color in the block quote from secondary to black.
    - [x] In Layout Builder, add the `hb-blockquote-black-text` class to the block which contains the block quotes. Confirm the text color changes to black.
5. Block quote in footer. Right now the only place that a block quote may encounter a dark background would be in the footer. If you have synced with staging, you should already have a block quote in the footer area.
    - [x] Scroll down to the footer area and confirm that the block quote is visible on a dark background. For Traditional, everything (text, border, icon) should be white. For Colorful, everything expect the icon should be white.


rkelbel:
This works as expected as far as applying, however, I noticed that the quotes around the actual text only come up if it's wrapped in a `<p>` tag, is this the accurate display/functionality?

When a user enters text and highlights the text to hit the blockquote button the WYSIWYG automatically throws a `<p> `tag around it. This is not the case with the other entry options.

Let me know if this is what we expect or if we need to surround any text within the `<blockquote> `with the quotes that like we have demonstrated in the design.

For example below, the top quote doesn't have `<p>` tags the bottom one does:
image

heatherdesigns:
@rkelbel 
> This works as expected as far as applying, however, I noticed that the quotes around the actual text only come up if it's wrapped in a <p> tag, is this the accurate display/functionality?

Excellent catch! Sorry I forgot to notate this in the PR. Because there isn't an element inside of the blockquote or div element to target, I can't add the quotes. But if the user wants the quotes, they can still add them to the text in the text area / WYSIWYG.

Looking at [staging](https://sparkbox-sandbox-stage.stanford.edu/), you can see where our current implementation that uses just the blockquote element also does not have quotes around the text.

blockquote element

rkelbel:
> @rkelbel
> 
> > This works as expected as far as applying, however, I noticed that the quotes around the actual text only come up if it's wrapped in a  tag, is this the accurate display/functionality?
> 
> Excellent catch! Sorry I forgot to notate this in the PR. Because there isn't an element inside of the blockquote or div element to target, I can't add the quotes. But if the user wants the quotes, they can still add them to the text in the text area / WYSIWYG.
> 
> Looking at [staging](https://sparkbox-sandbox-stage.stanford.edu/), you can see where our current implementation that uses just the blockquote element also does not have quotes around the text.
> 
> blockquote element

Makes sense, good to bring to H&S's attention I think, they may already realize it but something we should likely clarify with them just to be sure.

I'm good to approve! Nice work!!

rkelbel (approved):
Awesome work Heather!

image

PR #801, opened 2021-02-02
Feat(STN-597): update figcaption to new caption/credit styling in WYSIWYG

rkelbel:
# READY FOR REVIEW

## Summary
This updates the old `figcaption` styling to match our new universal caption/credit styling as a gradient overlay.

"There are 2 caption fields for text area images. One is in the media library, the other is added via the WYSIWYG. This ticket’s scope is limited to displaying the caption entered in the WYSIWYG (not the one stored in the media library). "

[JIRA Card: 597](https://sparkbox.atlassian.net/browse/STN-597)

## Need Review By (Date)
2/3/21

## Urgency
Medium

## Steps to Test
1. `npm run test`
2. To test this new styling you'll need to add a page using the image property caption spot, this work does not style the caption/credit that is coming from the media module caption entry area. This option will show up when you add an image through the WYSIWYG or you can click the "Edit Media" button on an existing image you've already placed.
image
3. The figcaption should now resemble the same look and feel of the caption work we've done with the postcards and earlier elements. Where the text is white and over a dark gradient overlay at the bottom of the images.
4. Changing the alignment of the image should not change the caption's function, it should stay with the image.
5. Changing the image style such as "Small Scaled" shouldn't affect the caption application either, it should scale and move with the image size.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
```diff
@@ -104,7 +104,11 @@
     }
 
     figcaption {
-      @include hb-caption($italic: true);
+      @include hb-caption-credit;
+      padding: hb-calculate-rems(16px) hb-calculate-rems(7px) hb-calculate-rems(8px) 15%; // Using a percentage to the left to control the width of the captions
```
heatherdesigns: ❤️ comment!

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
```diff
@@ -104,7 +104,11 @@
...
+      text-align: right;
+      line-height: 120%;
+      height: auto;
```
heatherdesigns: It looks like the mixin of `hb-caption-credit` is only used here and in `docroot/themes/humsci/humsci_basic/src/scss/utilities/_caption-credit.scss`. text-align, line-height, and height all have the some values in both places. It may make sense to place these properties into the mixin. Test it and see if it works.

heatherdesigns (commented):
Looks and functions fantastic! I tested a variety of image alignments and sizes and all of them look great.

thats-really-nice

I left a couple comments in the code and a few thoughts below.

We have a character limit for captions on images in paragraphs and views. Is it possible to add a limit the number of characters in the WYSIWYG image caption as well? I wasn't sure if that would be expected or if it is even possible.

number of characters

Can we update the comments in `docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss` for `hb-caption` and `hb-caption-credit` so that they are reflective of the updated usage? 

comment

We will still need mixin of `hb-caption` to support the `hs-credits` and `hs-caption` legacy classes. Let's update the comment for the `hb-caption-credit` mixin to reflect its new usage on both views, paragraph types, and in the WYSIWYG.

And I think you have a rouge commit. It probably stuck around after a rebase.

extra commit

rkelbel:
> We will still need mixin of `hb-caption` to support the `hs-credits` and `hs-caption` legacy classes. Let's update the comment for the `hb-caption-credit` mixin to reflect its new usage on both views, paragraph types, and in the WYSIWYG.

I have removed the comment, per our discussion on Slack it pretty much applies to all the captions with this work so no need to really have the comment call it out anymore.

> We have a character limit for captions on images in paragraphs and views. Is it possible to add a limit the number of characters in the WYSIWYG image caption as well? I wasn't sure if that would be expected or if it is even possible.

This isn't a setting that is available in Drupal's admin settings that I could find or am aware of. However we could use a theme hook to do this limitation, but as we also discussed via Slack. We will bring it up in stand-up and leave that limiting work as refinement in a later card if we want it as it is out of scope for this card.

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
rkelbel: This was a good callout, it works and I didn't see anything break in testing.

rkelbel:
> And I think you have a rouge commit. It probably stuck around after a rebase.
You were correct, I did a new rebase and that is now gone, should be all good there.


Additionally, I added a small bug fix I found while testing the captions, I had missed `<ol>` in the previous caption style work so I added it in.

docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss
```diff
@@ -104,7 +104,8 @@
     }
 
     figcaption {
-      @include hb-caption($italic: true);
+      @include hb-caption-credit;
+      padding: hb-calculate-rems(16px) hb-calculate-rems(7px) hb-calculate-rems(8px) 15%; // Using a percentage to the left to control the width of the captions
```
heatherdesigns: Ack! 🙈  
I should have looked closer at the padding. It has the same values here and in the `_caption-credit.scss` file. We could probably move this into the `hb-caption-credit` mixin as well.

And I noticed that there's a media query that changes the padding for larger screen sizes. Let's add that as well. The screenshot below is a side-by-side of a WYSIWYG caption on the left and a postcard image caption on the right. You can see that the padding is different. For consistency sake, let's make both the same.

padding media query

heatherdesigns: Here's the postcards page I was using for reference: http://sparkbox-sandbox.suhumsci.loc/qa/qa-postcards. You may need to add a caption to a postcard image.

heatherdesigns (commented):
Almost there! One more request + the compiled CSS and this is good to go!

heatherdesigns (approved):
bravo

PR #816, opened 2021-02-22
STN-650:  Views padding refinement

rkelbel:
# READY FOR REVIEW

## Summary
If there are multiple views stacked on a page with layout builder, there is no spacing between the title and the bottom of the previous view.

[JIRA Card: STN-650](https://sparkbox.atlassian.net/browse/STN-650)

TEST-RESEARCH-AREA-Latin-Department-of-Classics

 Live Examples: 
See examples of how spacers were necessary and added on https://classics-prod.stanford.edu/admin/structure/types/manage/hs_research/display/default/layout 

## Need Review By (Date)
2/22/21

## Urgency
medium

## Steps to Test
1. npm run test
2. To check this card's work you will need to create a page that has multiple views on the page added through the Layout Builder components area. 
Something like this:
image
3. I recommend also adding some views using paragraphs in the page edit area as well, so you can compare what views spacing is like with paragraphs vs. the Layout Builder component direct block component additions.
4. Once you have that page created, you should see that the Layout Builder component Views you've added have margin at the bottom for spacing that is consistent with the same margin used between Views in the paragraphs additions. This margin should be consistent with the usage of each theme type.
image (1)
5. The views should have no margin on the top of the first one or the bottom of the very last one and if only one view is added to the component area no extra margin should add at all.
6. Now also check these View component layouts with the Dark Inversion setting in each theme. The spacing should be consistent.
7. Additionally check using the` hb-dark-pattern` utility class on the individual views, adding the class to the CSS: area of the View itself. (Edit View > Advanced > Other > CSS Class)
8. The last way to verify `hb-dark-pattern` would be to add it to the Layout Builder Attributes block class (Layout > Manage Attributes > Block Attributes > Class), I updated this class in this card to add the same light padding around the block to give separation from the edge of the view, this also allows H&S to wrap the Title of the View as well if they would wish to wrap it that way.
image

In this image below you will notice the difference between how the View Title is wrapped and not wrapped by the hb-dark-pattern background. The top application is applied to the View CSS class, the lower one is applied to the Layout Builder Attributes Block Class. Just worth noting how they are applied and how they should look different in terms of the Title.
image

9. The following utility classes have been tested with this work and all apply as they were previously, feel free to check the following classes for verification.
hs-well :check_mark: 

- hb-title
- hb-heading-1 
- hb-heading-2 
- hb-heading-3 
- hb-heading-4 
- hb-heading-5 
- hb-heading-6 
- hb-divider
- hb-heading-link-color

_Notes: 
- hb-heading classes can be added to the View using the Layout Builder > Manage Attributes > Block Title Attribute > Class, it will override the Views Block title. 
- I also did notice there is some hierarchy issues applying the hb-heading 3-6, to the title field in Views, which are pre-existing and not caused by this card work. Also 100% it's a concern._

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/_main.scss
```diff
@@ -164,6 +164,7 @@
   'components/unpublished',
   'components/media',
   'components/private-page',
+  'components/views-layout',
```
heatherdesigns: I think this work may be better suited placed under `objects` with other layout type styles. We have `_layouts.rows.scss`, let's make a `_layouts.views.scss` file for this.

move placement

docroot/themes/humsci/humsci_basic/src/scss/components/_views-layout.scss
```diff
@@ -0,0 +1,10 @@
+//These are styles for views blocks specifically added through Layout Builder.
+.layout__region .block-views {
+  &:not(:last-child) {
+    margin-bottom: hb-spacing-width();
+  }
+
+  &.hb-dark-pattern {//Allows hb-dark-pattern to be added with Layout Builder Attributes block class
```
heatherdesigns: Add space after opening bracket please. :smile:

heatherdesigns (commented):
This looks great! Although I think we may want to move this work into the `objects` directory rather than the `components` directory.

heatherdesigns (approved):

PR #822, opened 2021-03-03
STN-703  Hotfix caption credit bug

rkelbel:
# READY FOR REVIEW

## Summary
As a logged out user the caption/credits styling appears in the middle of the screen.

JIRA Card: 
https://sparkbox.atlassian.net/browse/STN-703

Prod Example: 
https://sparkbox-sandbox-prod.stanford.edu/news/test-news-0

Approach:
What I figured out from examining this issue is this is happening from the Media Image fields on node pages.
We don't see the issue when we are logged in because there is a "contextual region" class that is applied for the editor links and that container has `position: relative` applied to it, thus containing the caption/credit div. When logged out that container disappears and so does the position.

As a temporary fix, in this work, I've added `position: relative `to the field classes that are affected. The intention is to remove this work after a more elegant solution has been created using a twig template for the caption-credit media field only, allowing us to add a position: relative wrapper to that application and not apply it to the other view modes the media images can utilize.

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. npm run test
2. Create a test page that uses one of the media image fields for uploading or using images from the media library, such as: event, event series, news, project ct, publication, or person.
image

3. Add an image to that test page that has a caption/credit added to it via the media caption/credit field for the image.
image

4. Ensure that the content type is displaying the caption/credit view mode of the media image via the Manage Display option for the content type.
image

5. Verify that the caption credit is appearing correctly as an overlay to the image in both the logged in and logged out state using Incognito browsing for logged out.

Before:
image

After:
image

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/components/_media.scss
```diff
@@ -17,3 +17,16 @@
...
+// Hotfix for caption/credit positioning bug with the media image field
+.field-hs-event-image,
+.field-hs-news-image,
+.field-hs-event-series-image,
+.field-hs-person-square-img,
+.field-hs-publication-image,
+.field-hs-accordion-image, //This still goes full width of the accordion, but keeps it contained to the image area.
+.field-hs-spotlight-image,
+.field-hs-hero-image,
+.custm-new-project-image {
```
pookmish: custm fields are not part of the code base. it's might require an asset injector if that is an issue.

pookmish (changes requested):

docroot/themes/humsci/humsci_basic/src/scss/components/_media.scss
rkelbel: I will remove that line, we are supplying an asset injector workaround to Alex as well so that shouldn't be an issue.

heatherdesigns (approved):

pookmish (approved):

PR #836, opened 2021-03-19
Feat(STN-690): Bug fix for raised vertical card width

rkelbel:
# READY FOR REVIEW

## Summary
`hb-raised-cards-uniform-height` impacts both vertical linked cards and vertical cards in unexpected ways.
On vertical cards that contained inline items such as links or a vertical linked card, the display:flex applied to the card container from the utility class made it so the width of the cards was not full width of the space within a row.

Jira Card: https://sparkbox.atlassian.net/browse/STN-690

## Need Review By (Date)
3/18/21

## Urgency
medium

## Steps to Test
1. npm run test
2. Create a flexible page with a row of postcards with anchor links and a row of vertical linked cards where you only add the title and link. You can create additional rows with different types of content but those first two additions are important because the links are inline elements that will not naturally expand 100% to the container within `display:flex`.
image
(I added the orange background to `.hb-cards` for troubleshooting.)

- As you can see in the screenshot the cards with full paragraph content or an image, there is no issue with them going full width. However, the cards that are linked cards or only contain a short anchor link, which is similar to the examples shared in the Jira card as well.

3. Verify that this fix applies `width:100%` only to the div following the postcard container classes only and it does not trickle down any further into the elements.
The cards should look like this with the application to the card wrapping div.
image

4. Verify this fix on both Traditional Colorful and that it doesn't affect horizontal cards negatively or other card types.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/utilities/_raised-cards.scss
```diff
@@ -6,6 +6,10 @@
   &--uniform-height .field-hs-row-components .hb-grid__item,
   &--uniform-height .hb-grid .hb-grid__item {
     display: flex;
+
+    & div:only-child {
+      width: 100%;
```
heatherdesigns, line 11: This works great!

heatherdesigns: Also check out the code on lines 22 - 24 for `.hb-raised-cards--uniform-height .hb-card`. With the changes above, we may not need to set the width at 100% here anymore.

heatherdesigns (commented):
Good work squashing this bug!! I left a recommendation. Test it out and let me know what you think.

cute high five

docroot/themes/humsci/humsci_basic/src/scss/utilities/_raised-cards.scss
rkelbel: I think this is valid, when I remove it and test it I do not see any regression. I'll go ahead and remove it and please follow up on your testing as well. Great suggestion!

heatherdesigns (approved):
Nice work!!! 

clipping chick

PR #853, opened 2021-04-19
STN-674: Callout Box Component

rkelbel:
# READY FOR REVIEW

## Summary
Create a Callout Box component, the ability to float a text area content within an immersive story.

West D7 Example: https://west.stanford.edu/news/blogs/and-the-west-blog/2020/fight-plague-local-governments-lean-trust-could-they-lose-it

HumSci D8 example: https://news.stanford.edu/2021/01/27/diseases-history-intertwined/

JIRA Card: https://sparkbox.atlassian.net/browse/STN-674

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. npm run test
2. Create a Flexible Page test page.
3. In the components paragraphs area you should see Callout Box in the list of components to choose from, add one.
4. The Callout Box only has two fields, the text area and the alignment selection, by default it will be selected to the Right.
5. Create a callout box with the default right alignment and one with the default left alignment.
6. On small/mobile sized screens the Callout Box should be 100% width and stack like the other components above it.
7. On medium to desktop sized screens the Callout Box should float within the text area below it and be a 40% width as requested in the card it will not be 100% on desktop size even if it is floating above a non-text area component.
8. Place the callout box above and below any other components and it should not float or interact with the other component types, unless it is a text area below it.

## PR Checklist


heatherdesigns (commented):
The Callout Box component is looking fantastic! It looks great across all breakpoints, stacking nicely on smaller screen sizes.

Here's a screenshot of my flexible page so you can see the combination of components I tested. 
callout box

I noticed a couple items:
1) We should probably prevent it from being used within rows as the text area. In the screenshot above the callout box has the default alignment of `right`. It doesn't work as expected because the row automatically places the next component in a column beside the first component.
row

2) Callout Box Number Three in the screenshot has links and a button. I cannot click on any of them.

3) Note that this is completely my opinion and I'm curious on Tori's feedback. I think it feels like there is too much spacing below the callout box. It's really noticeable on Callout Box Number Two in the screenshot. It's noticeable in Callout Box Number One as well, but because of the content, it isn't as bad. I will defer to H&S's opinion on the spacing.

rkelbel:
> I noticed a couple items:
> 
> 1. We should probably prevent it from being used within rows as the text area. In the screenshot above the callout box has the default alignment of `right`. It doesn't work as expected because the row automatically places the next component in a column beside the first component.
> 1. Callout Box Number Three in the screenshot has links and a button. I cannot click on any of them.
> 2. Note that this is completely my opinion and I'm curious on Tori's feedback. I think it feels like there is too much spacing below the callout box. It's really noticeable on Callout Box Number Two in the screenshot. It's noticeable in Callout Box Number One as well, but because of the content, it isn't as bad. I will defer to H&S's opinion on the spacing.

1. I agree, I thought I had removed it from the Row configuration, but apparently, I haven't. When I tested it, it wasn't working, not to mention there really isn't a use case for a callout in a row, you'd just use a text area instead. I will get this removed from config.

2. Interesting, the floats really make the text area component below it smother the callout element, I'll investigate this, great catch!

3. I will have to recreate some of your components you made with those text styles, I have a feeling it's a compounding space of the elements having margin and the callout having margin below it itself because it should only have 2.4rem max below the actual callout-box div, any padding added within it add more white space.  I'll recreate some of your blocks and try and find a good solution. Thank you!!

docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.callout-box.scss
```diff
@@ -0,0 +1,26 @@
...
+
+    &--align-left {
+      float: left;
+      margin-right: hb-spacing-width('xs');
+    }
+  }
+}
+
+// Float should only apply if text areas other siblings should clear the float element
+.ptype-hs-callout-box {
```
heatherdesigns, line 24: I don't think we need these styles. I removed them and it didn't change anything. I think it's because the float is applied to a single div with the class of `hb-callout-box` which is inside of `ptype-hs-callout-box`. So if there was a second div inside of `ptype-hs-callout-box` it would need to be cleared.

heatherdesigns: Never mind. :smile: thank you for checking!!!!

heatherdesigns (approved):
Fantastic work!!!!

PR #892, opened 2021-05-28
STN-777: Color band sidebar refinement

rkelbel:
# READY FOR REVIEW

## Summary
We want the Color Band to be visible full width in the sidebar left/right sidebar regions unless a block extends over the element.

This work also adds a utility class called `hb-sidebar-block` which applies the existing styles that sidebar group blocks already receive.  This allows the same styles to be added onto blocks within the layout builder for the sidebar for things such as custom blocks, which are edge cases.

JIRA Card: https://sparkbox.atlassian.net/browse/STN-777

## Need Review By (Date)
6/4

## Urgency
medium

## Steps to Test
1. npm run test
2. To test this item, you'll need to add a Color Band on a page that has menu sidebars or you'll want to add sidebar elements to a page with color bands you are testing manually. A page I used for testing was http://sparkbox-sandbox.suhumsci.loc/qa/qa-embeds.
3. You'll want to test against menu blocks on each sidebar, as well as adding other block types using examples within the JIRA Card such as "Group Blocks: Content", then adding custom blocks within it that are listed such as "about Stanford", etc.
4. We've also added a utility class in this work which gives the same padding and the border that is around the group blocks. This can be tested by adding a custom-created block to the sidebar, use the "+ Create Custom Block" button that is featured at the top of the Layout Builder block sidebar menu. Then using "manage attributes" on the block add the `hb-sidebar-block` class on that block. This should resemble the same look as any of the group blocks that are added by themselves.
image

- [x] Verify `hb-well` works with color band when applied to a sidebar group block via Layout Builder.
- [x] Verify these tests work with each theme and color pairings.
- [x] Verify the utility class has also been added to the Group Block Utility Class section of our documentation.

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.general.scss
```diff
@@ -304,4 +304,10 @@
 
   padding: hb-calculate-rems(20px) hb-calculate-rems(18px);
   border: 0 none;
+
+  .hb-three-column__sidebar-1 &,
+  .hb-three-column__sidebar-2 & {
+    position: relative;
+    z-index: 1;
```
heatherdesigns: Let's use a variable from `humsci_basic/src/scss/settings/_variables.zindex.scss` for the value of `1`.

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.layout.scss
```diff
@@ -98,10 +98,22 @@
...
 
   // The override below changes the sidebar styling to default within the layout builder admin area.
   .layout-builder & {
     z-index: auto;
     background-color: initial;
   }
 }
+
+// This mixin adds a padding and border outline for utility usage.
+@mixin hb-sidebar-block {
```
heatherdesigns, line 117: I love a good mixin! Nice work!! :tada:

heatherdesigns (changes requested):
Fantastic work on the color band refinement! This is looking sooooo much better! :tada:

I was able to apply the `hb-well` class to a block in the sidebar as seen on the [Upcoming Events and Seminars page](https://economics.stanford.edu/seminars-events/upcoming-seminars-and-events) on the Economics site.

It looks like the background-color of `hb-well` is being overwritten by the background-color set for sidebar utility class.

hb-well

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.general.scss
rkelbel: This has been fixed, also found a second place we needed to update this and did that at the same time.

rkelbel:
> I was able to apply the `hb-well` class to a block in the sidebar as seen on the [Upcoming Events and Seminars page](https://economics.stanford.edu/seminars-events/upcoming-seminars-and-events) on the Economics site.
> 
> It looks like the background-color of `hb-well` is being overwritten by the background-color set for sidebar utility class.

So I found that since those classes were moved to a mixin, the hierarchy in how they are applied have changed. I tried a few options but I would lose the background color on the nav sidebars if I removed it and moved it to the block utility class.

So I added the well mixin to the sidebar to catch this application.

Let me know if you have additional ideas here, but right now it should be working correctly.

heatherdesigns (approved):

PR #929, opened 2021-07-26
STN-793:  Bug fix aligned images in wysiwyg

rkelbel:
# READY FOR REVIEW

## Summary
When inserting an image in the Text Editor, and the image is set to align left or right, the CKEditor does not closely reflect the correct styles. (Note: this only effects the editor preview, not the final page.

JIRA Card: https://sparkbox.atlassian.net/browse/STN-793

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. npm run test
2. `lando drush @sparkbox_sandbox.local cr` - this work adds a new media image field template and needs a cache clear to pick it up.
2. Open a page with images in WYSIWYG or create a new one.
3. Add images in through the WYSIWYG media button, when adding those images, use different scales and sizes.
4. Align the images left, right and center, with text below the image that aligns to the left or right of the image if you choose an alignment for the image. These should float correctly next to each other unless the image is centered.
5. The images should align correctly left and right and should also take up the correct amount of space to more closely reflect what the alignment with text will look like when saved.

## PR Checklist


docroot/themes/humsci/humsci_basic/templates/field/media--image.html.twig
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This Caption Credit template overrides the standard media.html.twig file in Stable theme.
+ */
+#}
+{%
+  set classes = [
+    'hb-media-image',
+  ]
+%}
+
```
heatherdesigns, line 11: Fantastic work finding a solution to target images here! 🏆

heatherdesigns (commented):
I know you spent a TON of time looking for the best solution. I want to recognize and thank you for your efforts!! :tada: :tada: :tada:

The text editor experience when adding images has been greatly improved!

I did notice that there's still quite a bit of space around an image that is right aligned when a caption is added.

### Screenshot without a caption:

regular

### Screenshot with a caption:

with caption

good job

heatherdesigns (approved):

rkelbel:
> I know you spent a TON of time looking for the best solution. I want to recognize and thank you for your efforts!! 🎉 🎉 🎉
> 
> The text editor experience when adding images has been greatly improved!
> 
> I did notice that there's still quite a bit of space around an image that is right aligned when a caption is added.

This should be addressed, caption images in media embeds have the alignment applied to the figure instead of the article like it does for all of the other media embeds. Added some styling that applies to the figure using the caption and drupal-media element so this should be fixed.

PR #933, opened 2021-08-03
HSD8-1091: Added year conversion to dateTime widget for years before 1969.

joegl:
# READY FOR REVIEW

## Summary
- Fixed absolute years not working in DateTimeYearOnly widget
- This had to do with the fact they were using 1949 specifically -- the DateTime object uses timestamp conversion so any year before `december 31 1969` won't work, and needs to be formatted as `1949-06-01` instead of just `1949`. Years after 1969 work fine.

## Urgency
Medium

## Steps to Test
1. Pull/setup locally
2. Open `/admin/structure/types/manage/hs_person/form-display` and set the Cohort field Min setting before 1970
3. Open `/node/add/hs_person`, go to Student Info tab, and check Cohort field min is set correctly

## PR Checklist


docroot/modules/humsci/hs_field_helpers/src/Plugin/Field/FieldWidget/DateTimeYearOnly.php
```diff
@@ -115,6 +115,8 @@ protected function getOptions() {
    *   The year of the datetime.
    */
   protected function getYear($dateTime) {
+    // rewrite YYYY as YYYY-06-01 for year before 1969
+    if((int)$dateTime > 0 && (int)$dateTime < 1970) $dateTime = $dateTime . '-06-01';
```
pookmish: Interesting issue. Do you have a link to a stack overflow on this or something?

pookmish:
```suggestion
    if ((int) $dateTime > 0 && (int) $dateTime < 1970) $dateTime = $dateTime . '-06-01';
```
BTW don't let @jbickar  see this oneliner. I bet he'd have a stroke. :rofl:

docroot/modules/humsci/hs_field_helpers/src/Plugin/Field/FieldWidget/DateTimeYearOnly.php
```diff
@@ -115,6 +115,8 @@ protected function getOptions() {
    *   The year of the datetime.
    */
   protected function getYear($dateTime) {
+    // rewrite YYYY as YYYY-06-01 for year before 1969
```
pookmish:
```suggestion
    // Rewrite YYYY as YYYY-06-01 for year before 1969.
```

pookmish (changes requested):

docroot/modules/humsci/hs_field_helpers/src/Plugin/Field/FieldWidget/DateTimeYearOnly.php
pookmish: It seems to me like this format might be needed for all years. When i attempt to put `2000` for the minimum, i only get the options starting from 2021

joegl: Interesting. I tested it with years in the 70's and it seemed to work but nothing after 2000. Maybe we just reformat every year from `YYYY` to `YYYY-06-01`?

pookmish: yeah i think that'd be easiest

joegl: So this is fun. It works with any `YYYY` date that's between 1970 and 1999. Otherwise, it needs to be converted to `YYYY-MM-DD` or `YYYY-06-01` in this case.

Here it is with 1999:

joegl: 😄

I will have to get used to `if (` vs `if(`

joegl: I just dropped the year condition altogether and am converting all YYYY to YYYY-MM-DD now.

pookmish (approved):

PR #957, opened 2021-09-07
Fix(STN-863): Full-width Color Band extends beyond the content region on mobile.

pebutler3:
## Issue
The color band component is can be displayed full-width, and it is
currently relying on overflow-x: hidden on the body to prevent overflow
issues, but after testing on mobile devices you can see that this simply
doesn't work.

# READY FOR REVIEW

## Summary
Add overflow-x: hidden; to `html` selector and this seems to resolve the
issue.

## Urgency
_Medium_

## Steps to Test
1. Create a Flexible Page and add a Full Width Color Band & Save.
2. Utilizing either physical devices or Browserstack, view your new page and confirm that you are no longer able to scroll the browser to the right and see the Color Band hanging out beyond the rest of the content.
3. Check in Colorful/Traditional themes.
4. Switch to different mobile browser and repeat.
5. For bonus points, inspect the device's browser and remove the `overflow-x: hidden;` from the `html` selector to observe the bug happening without its addition to the stylesheets.

## Browser testing
- [x] Check the standard desktop browsers to ensure there isn't any regressions in functionality.
- [x] Mobile (Where the bug is specifically called out):
    - [x] iOS - Safari
    - [x] Android - Chrome, Firefox

## PR Checklist


rkelbel (approved):
This works and looks great Paul!

nice

PR #964, opened 2021-09-16
Fix(STN-877): config updates private collections

pebutler3:
# READY FOR REVIEW

## Summary
Update field configurations to achieve:
* Color band is no longer available for Private Collections
* Items Per Row is a required field
* Items Per Row defaults to 1

## Urgency
low

## Steps to Test
1. Create a private page
2. Add a new collection and confirm that Items Per Row defaults to 1 and is now required.

## PR Checklist


rkelbel (approved):
This works exactly as it should, nice work Paul!

Some things to note for Release Notes Card:
We will probably want to make these notes in the Release Notes card for this sprint if anyone has been using the private collection at all, which I'm not sure if they have.

It would be good to note that any existing Private Collections in prior existence will be required to select an "items per row" if they have not done so already since it is now required.

And they won't be able to save the page without doing so, nor will they be able to save a page they have a color band added to that is within a private collection, that will need removed before saving.

PR #983, opened 2021-10-19
fix(STN-875): default images not loading

rkelbel:
# READY FOR REVIEW

## Summary
Default image for people and events not appearing in views for site owners/contributors when logged in, but they do appear for anonymous viewers, developer and administrator roles.

JIRA Card: https://sparkbox.atlassian.net/browse/STN-875

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. npm run test
2. The JIRA Card in the description has a lot of staging and other links available to dive deeper into the problem to see the errors.
3. The easiest way to see the issue is to masquerade as a content editor role and then go to the People Grid page: http://sparkbox-sandbox.suhumsci.loc/people-grid and see if all of the images load for you.  (There will be images that are broken, those do not count in this as an issue.)
4. Default images are especially an issue with loading, as they should be the default image that loads when a staff member doesn't have an image to it's page profile. You can update the default images here: http://sparkbox-sandbox.suhumsci.loc/admin/structure/field-default-values , if you are only testing the People grid, the only default image you need to worry about updating is the People option.
5. To Masquerade as different users you can do that using the Masquerade module, go to http://sparkbox-sandbox.suhumsci.loc/admin/users/masquerade (you can always get to this from the menu Manage > Users > Masquerade).
- In the masquerade autocomplete search box you can imitate any registered user on the site, a good user to choose to test for Content Editors is the user "Howard".
6. After you've enabled your masquerade profile, you can visit the People Grid page and investigate the page to make sure the images are all still loading.  The Howard profile will remain active on all page you browse until you "un-masquerade". (Which is available as a link in the toolbar next to the username/user icon on the top.)

If you want to verify the change I made, you can also re-add the wildcard to the css and then investigate those same pages with the user roles and see how it loads.

## Test Verification:
- [x] Test all card types with images, with content editor, admin and logged out states to make sure the wildcard hasn't broken anything.
- [x] Test this on both themes, with different types of cards, Vertical Linked, Vertical, Horizontal, etc.
- [x] Test with images and without.
- [x] Test inside Collections and other Paragraph components that may be able to contain a Postcard with an image.

## PR Checklist


pebutler3:
Way to go @rkelbel this is quite a find -- it makes me wonder why it was there initially. We'll have to be pretty thorough with the backstop tests.

PR #1001, opened 2021-11-12
Feat(STN-855): Lead Font Animation

pebutler3:
# READY FOR REVIEW

## Summary
Add Lead Font Animations for Traditional & Colorful themes. Implemented `reducedMotionPrefered` and `hasAnimationEnhancements` mixins.

[STN-855](https://sparkbox.atlassian.net/browse/STN-855)

## Need Review By (Date)
asap

## Urgency
medium

**Notes:**
The description of the animations in the original card are a confusing to follow if you read all the way through. However, I have distilled this down to a few scenarios.
1. Traditional theme animations are `opacity` changes only.
    - If the Lead Font text is a sibling of Splash Font, it has an animation delay.
2. Colorful theme animations have opacity + transforms
    - If the Lead Font text is a sibling of Splash Font, there is NO animation delay.

## Steps to Test
1. `npm run test`
2. Create a content page with a Splash Font and a Lead Font in the same Text Area.
3. Create a separate Text Area with just a Lead Font.
4. Ensure Animation Enhancements are enabled on your active theme.
5. Observe animation behaviors
6. Switch themes and repeat
7. Confirm enabling Prefers Reduced Motion, still prevents the animations from running.

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
+++ b/docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
@@ -421,33 +421,47 @@
 // Splash Font Animation
 @mixin hb-font-splash-animation {
   @include hb-traditional {
-    &.animate {
-      .hb-has-animation-enhancements & {
-        animation-timing-function: ease-in-out;
-        animation-duration: 1700ms;
-        animation-fill-mode: forwards;
-        animation-name: traditionalSplashFont;
-      }
-    }
+    @include hasAnimationEnhancements(ease-in-out, 1700ms, traditionalSplashFont);
   }
 
   @include hb-themes(('colorful', 'airy')) {
-    &.animate {
-      .hb-has-animation-enhancements & {
-        animation-timing-function: ease-in-out;
-        animation-duration: 2000ms;
-        animation-fill-mode: forwards;
-        animation-name: colorfulSplashFont;
-      }
-    }
+    @include hasAnimationEnhancements(ease-in-out, 2000ms, colorfulSplashFont);
   }
 
-  // Removes animation styles when Reduced Motion is enabled
+  @include reducedMotionPrefered;
+}
+
+// Lead Font Animation
+@mixin hb-font-lead-animation {
+  @include hb-traditional {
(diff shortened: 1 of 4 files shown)


docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
```diff
@@ -451,3 +454,14 @@
...
+@mixin hasAnimationEnhancements($timingFunction, $duration, $animationName) {
+  &.animate {
+    .hb-has-animation-enhancements & {
+      animation-timing-function: $timingFunction;
+      animation-duration: $duration;
+      animation-fill-mode: forwards;
+      animation-name: $animationName;
+    }
+  }
+}
```
rkelbel, line 467: Nice mixin. 👏🏻

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
```diff
@@ -421,28 +421,31 @@
...
+  @include hb-themes(('colorful', 'airy')) {
+    @include hasAnimationEnhancements(ease-in-out, 2000ms, colorfulLeadFont);
   }
 
-  // Removes animation styles when Reduced Motion is enabled
+  @include reducedMotionPrefered;
+}
+
+// Removes animation styles when Reduced Motion is enabled
+@mixin reducedMotionPrefered {
```
rkelbel, line 448: 👏🏻

rkelbel:
This work looks good, it's functioning as expected for the most part.  I did see an occasional cached text and then a reload animation, but it was not reproducible on every refresh.

The one thing I wanted to point out is actually some verbiage in the JIRA card and the actual styling and animation expectation. The lead font and splash font adapt to each other as they are supposed to when they are in the same text area together.  However, in the card, we mention "same text area" and "same page" that being said, when these are on the same page together but not the same text area they react as they would as if they were single elements added and not like they do when in the same text area. _"Animation properties are listed below for when the lead font is on a page with the splash font."_
- I think it's totally fine that this is how they are acting and should be the expectation, but since we also have the phrase "on the same page" in the JIRA ticket, I think it's a good idea for us to surface this behavior with H&S to make sure they understand the behavior and they have the correct expectations.

PR #1022, opened 2021-12-23
Feat(STN-883):  Vertical timeline pattern for views

rkelbel:
# READY FOR REVIEW

## Summary
As a Developer, I would like to use the Vertical Timeline pattern in a view and have the styles appear like the pattern used in a paragraph type.

This work aims to address some specific issues that do not currently work as we would like without impacting the current functionality of the Vertical Timeline as a Paragraph.

- **Current issue**: _" The Variants for Closed and Open should function as expected. If necessary, use default behavior (expanded). "_
  - This issue was addressed utilizing a class on the view to modify the functionality.
- **Current issue:** _" Currently the pattern appears without the vertical connecting lines. "_
  -  This issue was addressed using CSS only.

Jira Card: https://sparkbox.atlassian.net/browse/STN-883

## Need Review By (Date)
1/3

## Urgency
medium

## Steps to Test
1. npm run test
2. To check the styling of the Vertical Timeline in a View you can use [this page](https://sparkbox-sandbox.suhumsci.loc/example-vertical-timeline-view) if you have synced with staging recently.
3. You can compare the styling against[ this page](https://sparkbox-sandbox.suhumsci.loc/qa-vertical-timeline-raised-cards) that is also available to sync from staging, this is the paragraph display of the Vertical Timeline.
4. Although the Pattern Variant setting within the view does not work to adjust the collapse and expansion of the timeline items, we have a utility workaround for this functionality.
- If you go to the View page for the Timeline View to [edit it here](https://sparkbox-sandbox.suhumsci.loc/admin/structure/views/view/custm_custom_econ_events/edit/block_11).
- Then add the CSS class `hb-timeline__collapsed` to the View.
image
- Save the View.
- This allows the existing JS for Vertical Timeline to function and allows the Timeline Items to be collapsed by default by correctly adjusting the Aria Expand attributes.
- This should meet the requirements set in the JIRA card for default functionality and allow the use of collapse if needed.

## Verification Checklist:
- [x] The Paragraph usage of Vertical Timeline styling is unaffected, functionality and styling are the same as previous.
- [x] The Vertical Timeline in Views has the ability to be collapsed on load by using the steps to add the `hb-timeline__collapsed` class in the View.
- [x] Using `hb-raised-cards` or `hb-raised-cards--uniform-height` doesn't alter the styling of the Vertical Timeline in a View.
- [x] This works as it should in both Traditional and Colorful themes.
- [x] This works properly with all color pairings, with the connecting lines and dots changing to the correct color pairing.
- [x] Verify the documentation has been updated to track the usage of the `hb-timeline__collapsed` [utility class doc update](https://github.com/SU-HSDO/suhumsci/blob/9b258c3da5cc605b75211c028fc3f142b8096130/docroot/themes/humsci/humsci_basic/docs/utility-classes.md)

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.timeline-item.scss
```diff
@@ -23,6 +23,33 @@
   }
 }
 
+// This adds the connected lines and dots to the Timeline Item within the Views pattern usage.
```
kaseybon: Instead of duplicating these styles we should modify line 2 to include these classes:
```
.ptype-hs-timeline-item,
.views-view-pattern  .hb-timeline-item { … }
```

docroot/themes/humsci/humsci_basic/docs/utility-classes.md
```diff
@@ -134,6 +134,13 @@ For best practice, you might just want to use one responsive class if you'd like
 | hb-card-image-10-of-12 | ✅|
 | hb-card-image-11-of-12 | ✅|
 | hb-card-image-12-of-12 | ✅|
+
+#### Vertical Timeline within Views
```
kaseybon: This could be combined with the table of classes on line 107-109. I don't think the vertical timeline needs its own separate section from other view specific classes. I'm also wondering if this needs to be view specific, couldn't this class be used on a component page as well?

kaseybon (changes requested):
Styles look great and working as expected but we have some duplication in the CSS that needs to be addressed. I also added some suggestions to the documentation to hopefully simplify our docs.

docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.timeline-item.scss
rkelbel: Good call out, I think I meant to refactor this after I got it working but forgot. Thank you!

docroot/themes/humsci/humsci_basic/docs/utility-classes.md
rkelbel: This class is already being applied within the component itself when you select the collapsed dropdown from paragraphs options, the reason it works as a View class is that the variant doesn't work inside of Views. Using this class allows us to apply that variant as a utility, while it could be added in a component section it is not needed as the component does it itself.

I will move it up with the other lines though, that makes sense.

kaseybon (approved):

PR #1023, opened 2021-12-23
Fix(STN-937):  Fix menu active links using front token with hb-dark-pattern

rkelbel:
# READY FOR REVIEW

## Summary
When using the Dark Inversion version of the masthead, and also using the the Drupal token for the home page of `<front>` for the Home menu link, the menu link was appearing wrong when the link was on that page as active.

image

Upon investigation it was found that when using a `node/1` path for home the link was getting the `hb-main-nav-link--active-trail` class added to the link, which was adding a background color to the active link class.

When using the Drupal token, a path is not followed and that class is not generated by the menu on the link itself. Therefore only changing the color of the text for the Dark Inversion of the menu but not giving it the background that would make the active link accessible.

_This work adds a background color to all active links in the menu within the Dark Inversion on the first level, there are also classes that cover the lvl2, etc, these remained in the styles because the specificity was there because of previous background color bugs on the lower levels. (There is likely room for refactoring here, but would also affect the non-inverted masthead links as well. Which may be more scope and testing than is meant for this card.)_

JIRA Card: https://sparkbox.atlassian.net/browse/STN-937

## Need Review By (Date)
1/3

## Urgency
medium

## Steps to Test
1. npm run test
2. Ensure that your Theme Settings [here](https://sparkbox-sandbox.suhumsci.loc/admin/appearance/settings) are set to use the Dark Masthead for whichever theme you are checking first, both will need to be tested.
2. Edit the Main Menu link for Home [here](https://sparkbox-sandbox.suhumsci.loc/admin/structure/menu/item/1/edit), change the link to `<front>`.
3. Save the link/menu.
4. Go to the Home page and resize your browser to utilize the mobile menu.
5. The Home link should be active and look similar to this:
image

## Verification Checklist
- [x] Home link now is accessible when active with the light background when using Dark Masthead for the menu in mobile.
- [ ] Home and other Active links within the desktop menu using the Dark Masthead are unaffected.
- [x] Active Links in the lower levels are still styled accurately within the Dark Masthead mobile menu.
- [ ] Home and other Active links within the desktop menu using the Light Masthead are unaffected.
- [x] This styling is accurate on both Traditional and Colorful themes.
- [ ] This styling changes accurately and remains accessible between color pairings on each theme.

## PR Checklist


kaseybon:
Active styles for the homepage on desktop are missing.

There are also some differences between the home link and interior pages when active. Interior pages include a light border on the left that updates to the secondary color on hover.

**Active Interior Link**
Screen Shot 2022-01-06 at 10 35 04 AM

**Active Interior Link on Hover**
Screen Shot 2022-01-06 at 10 35 10 AM


**Active Home Link**
Screen Shot 2022-01-06 at 10 35 18 AM

rkelbel:
> Active styles for the homepage on desktop are missing.
> 
> There are also some differences between the home link and interior pages when active. Interior pages include a light border on the left that updates to the secondary color on hover. Active Interior Link...

This is something using that `active-trail` class that Drupal applies as well, the border does work when the home page is using a path and not the `<front>` token.

I'm not sure how to approach adding this to that link using the same `.is-active` class I fixed the accessibility background with as it would apply it to every 'is-active' link in the hierarchy of a menu tree.

kaseybon:
Wherever we are targeting the `.hb-main-nav__link--active-tail` class we can also target the data attribute that is added to links using the `<front>` token: `[data-drupal-link-system-path="<front>"]`.

We could also look at the template file where we are adding the active trail class and see if there is anything we can add to the logic to also add the class to links using the token.

rkelbel:
> Wherever we are targeting the `.hb-main-nav__link--active-tail` class we can also target the data attribute that is added to links using the `<front>` token: `[data-drupal-link-system-path="<front>"]`.
> 
> We could also look at the template file where we are adding the active trail class and see if there is anything we can add to the logic to also add the class to links using the token.

I've updated the styling to use the data link as you mentioned, should be good to go now.

kaseybon (approved):

PR #1038, opened 2022-01-24
Fix(STN-838): Background in Exposed Views Choice Filter in well

rkelbel:
# READY FOR REVIEW

## Summary
The hs-well color is impacting the exposed filter using both multiselect and dropdown select fields type in a view.

This occurs in both scenarios: 
- views used in a collection with a hb-well background
- using the utility class hs-well on the view

Link example:https://ealc-prod.stanford.edu/people/alumni

JIRA Card: https://sparkbox.atlassian.net/browse/STN-838

## Work Done in this Card
Upon further review the background color was an issue with both multi-select and single select dropdown exposed filters in Views with a `hb-well` behind them.

While reviewing the CSS that was being applied to this card I noticed that we were also applying a white background to this when the dark inversion was turned on or `hb-dark-pattern` utility class was added to the View classes.

This work adds a white background to that basic exposed filter multi or single selection and removes unneeded classes that was addressing dark inversion only as it is no longer needed.

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. npm run test
2. Find or create a View with an exposed Filter selection. A good one to use is http://sparkbox-sandbox.suhumsci.loc/news.
3. Add the `hb-raised-cards` and `hb-well` class to the component region of that page using the Layout tab. You can set those classes using the "configure" option on the "Main Region" field block. 
4. To test to make sure the dark inversion option still functions accessibly, you would add the `hb-dark-pattern` utility class to the classes on the View itself.
image
5. To turn on the dark inversion for the whole theme, you would need to go to the Appearance > [theme] settings > and check the dark inversion checkbox.

## Confirmation checklist:

- [ ] The background for multi-select Views exposed filter is correct without `hb-well` on the page.
- [ ] The background for single select Views exposed filter is correct without `hb-well` on the page.
- [ ] The background for multi-select Views exposed filter is correct with `hb-well` on the page.
- [ ] The background for single select Views exposed filter is correct with `hb-well` on the page.
- [ ] The background for multi-select Views exposed filter is correct with `hb-dark-pattern` on the view.
- [ ] The background for single select Views exposed filter is correct with `hb-dark-pattern` on the page.
- [ ] The background for multi-select Views exposed filter is correct with `hb-well` on the page, on all Traditional color pairings
- [ ] The background for single select Views exposed filter is correct with `hb-well` on all Colorful color pairings.

## PR Checklist


kaseybon:
@rkelbel by moving the background up to the container element we lose the rounded corners for dropdowns. We can probably add a border radius here to match the `.chosen-choices` element.
 News___QA_Site

rkelbel:
> @rkelbel by moving the background up to the container element we lose the rounded corners for dropdowns. We can probably add a border radius here to match the `.chosen-choices` element. News___QA_Site

Border-radius should be added, great catch, hard to see on well but very present on dark pattern.

kaseybon (approved):

PR #1051, opened 2022-02-15
Stn 653  spotlight color dropdown

lunarfusion:
# READY FOR REVIEW

## Summary
Added option on the Spotlight paragraph type to select a global gray background color (#D9D7D2) or the 
 theme's default Spotlight background color.

View of select list options:
Screen Shot 2022-02-16 at 9 16 26 AM

View of results in Colorful/Cardinal:
Screen Shot 2022-02-16 at 9 55 02 AM

View of results in Colorful/Mountain:
Screen Shot 2022-02-16 at 9 55 21 AM

Jira Card: https://sparkbox.atlassian.net/jira/software/c/projects/STN/boards/80?modal=detail&selectedIssue=STN-653

## Need Review By (Date)
asap

## Urgency
medium

## Steps to Test
1. npm run test
2. Create a new page and add a Spotlight paragraph type to that page.
3. Under the "Background Color" select list, select "Default background" and verify that the theme's default spotlight background appears.
4. Under the "Background Color" select list, select "Medium Gray background" and verify that the global medium gray background #D9D7D2 appears.

## PR Checklist


rkelbel (commented):
Monet, first of all, this is awesome work! Congrats on the first SBX PR request! 🎉 💥 😄 

This is work is very close but I did have some things for feedback that we will want to address or discuss before merging this work.

rkelbel (changes requested):
There seems to be some configuration missing from this work that fully adds the configuration for the template to add the class name from the new field consistently.

Namely, When I create a Short Spotlight slide, the class is not populating for me on my spotlight slide.  It is however populating correctly for the Tall spotlight.

This is due to the configuration between "Default" (Tall) and "Preview" (Short) Displays not being the same in their field display for the background select list field.

Preview (Short):
image

Default (Tall):
image

You can see between those two screenshots that the field is not in the Background area on short and it is on tall. This likely happened because you and I missed the Preview configuration display file when we paired to export the configuration.
If you do another export you can find this missing file and export it by itself and discard any of the other config that spits out. The file you will be looking for is: `core.entity_view_display.paragraph.hs_spotlight.preview`.
Once you export that file, we should be able to import the short display variant and see the class properly populating.

rkelbel (changes requested):
Some additional config changes we will want to make, is to make the Label "hidden" for the field on both the Preview and Default displays for Short and Tall. Although this field is not actually showing up, I think we should cover our bases by removing a possible label from the equation.

image

Additionally, I believe the request from Tori was for the list options to reflect similar listings as Color Band.  Currently, we have the background gray option listed as "Medium Gray background" in the select list.  
I think this could be simplified to just say "Gray" as there is not as of now an additional gray selection. 
We can also simplify "Default background" to "Default". Similar to Color Bands options: https://sparkbox-sandbox.suhumsci.loc/admin/structure/paragraphs_type/hs_clr_bnd/fields/paragraph.hs_clr_bnd.field_hs_clr_bnd_bgc/storage

image

If you would like to verify this with Tori prior to updating this in configuration please do so, I just thought I would include this feedback for consideration and consistency.

rkelbel (changes requested):
The last config change we should make is also something H&S seems to prefer on fields like this.
We may want to make this field "Required" for the background color choice, setting the field to required removes "None" from the select list as an option. Although we have a Default color set, that would ensure "None" cannot be selected.

Although I do not see this as an Acceptance Criteria from H&S on the card, it is probably worth confirming before committing all of the configurations for this work.

image

config/default/core.entity_form_display.paragraph.hs_spotlight.default.yml
```diff
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
+    - field.field.paragraph.hs_spotlight.field_hs_spotlight_bg
```
rkelbel, line 6: Perfect Field machine-id naming! 👍🏻

docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.spotlight.scss
```diff
@@ -2,6 +2,10 @@
   padding: hb-spacing-width('md') 0;
   @include hb-pairing-color('background-color', 'spotlight');
 
+  &--background-gray {
+    @include hb-global-color('background-color', 'gray-medium');
+  }
+
```
rkelbel, line 8: Nice SASS! 💥

rkelbel (approved):

rkelbel (approved):
This looks awesome @lunarfusion ! Nice work! 🎉 
high five

PR #1056, opened 2022-02-18
STN-478: Expand all accordion functionality

kaseybon:
# READY FOR REVIEW

## Summary
This pull request adds a toggle all feature for accordion buttons. The ability to turn on this toggle is an optional field on each accordion item. Enabling the toggle on at least on accordion will add a class to the accordion that will trigger the Javascript to create a toggle button and insert it before the first accordion on the page.

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. Run `npm test`
2. Import the accordion config
3. On a page with multiple accordions enable the toggle all button by editing an accordion by turning on the expand/collapse toggle

- [x] Confirm that the toggle all button is keyboard accessible.
- [x] Confirm that the toggle will expand/collapse all accordions on the page and that the button text updates appropriately.
- [x] Confirm that a page that does not have an accordion but does have the `hb-accordion_toggle-all` on the component region does not have console errors.
- [x] Confirm that there are no console errors on pages that do not have an accordion or the `hb-accordion_toggle-all` class.
- [x] Confirm that accordions on pages without the toggle enabled do not have a toggle all button.

## PR Checklist


rkelbel:
@kaseybon So this works fantastically and looks great, however, I noticed there was something with the functionality that was different than we had discussed with H&S.  I believe H&S was wanting that each place that this accordion option is enabled that the Toggle All button will appear above that accordion.  That doesn't seem to be the case, when multiple accordions have this enabled only the one Toggle button appears at the top of the whole list of accordions.

Ideally, H&S wants this to be something that appears wherever the client deems necessary in their page layout.
Let me know if you have any questions on this.

Both of these Accordions are checked to enable the toggle, but only the one button is present.
Also confirmed in code that the div with the correct class appears within each accordion element.
image

PR #1062, opened 2022-03-01
Fix: spotlight bg hook added and remove region disable field

rkelbel:
# READY FOR REVIEW

## Summary
This work changes the Spotlight background class field application from Twig variable to a hook inside the module.
No functionality should change swapping between these methods.

However, the "Background" region inside the Paragraph should know be removed and the background field should now be set to disabled if you look at the Manage Display tab for the Spotlight Paragraph.  Checking both Preview and Default display modes for Tall and Short, they should both be the same.

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. npm run test
2. Import the new configuration `lando drush @sparkbox_sandbox.local config:import --partial`
3. Clear cache `lando drush @sparkbox_sandbox.local cr`
4. View a page with different accordion types setup and different colors.
5. Check default option has not changed
6. Verify it works on both Traditional and Colorful.

## PR Checklist

### DIFF
--- a/docroot/modules/humsci/hs_layouts/hs_layouts.module
+++ b/docroot/modules/humsci/hs_layouts/hs_layouts.module
@@ -5,6 +5,8 @@
  * hs_layouts.module
  */
 
+use Drupal\Component\Utility\Html;
+
 /**
  * Implements hook_plugin_filter_TYPE__CONSUMER_alter().
  */
@@ -39,5 +41,17 @@ function hs_layouts_preprocess_container(&$variables) {
   if (!empty($variables['element']['layout-section'])) {
     // Add styles to the layout builder admin configuration.
     $variables['#attached']['library'][] = 'hs_layouts/layout_builder_admin';
+  }
+}
+
+/**
+ * Implements hook_preprocess_preprocess_pattern_spotlight().
+ */
+function hs_layouts_preprocess_pattern_spotlight(&$variables) {
+  $paragraph = $variables['context']->getProperty('entity');
+  if ($paragraph->hasField('field_hs_spotlight_bg') && !$paragraph->field_hs_spotlight_bg->isEmpty()) {
+    $bgValue = $paragraph->field_hs_spotlight_bg->first()->getValue()['value'];
+
+    $variables['spotlight_bg_class'] = Html::cleanCssIdentifier($bgValue);
   }
 }
--- a/docroot/modules/humsci/hs_layouts/hs_layouts.module
+++ b/docroot/modules/humsci/hs_layouts/hs_layouts.module
@@ -49,9 +49,8 @@ function hs_layouts_preprocess_container(&$variables) {
  */
 function hs_layouts_preprocess_pattern_spotlight(&$variables) {
   $paragraph = $variables['context']->getProperty('entity');
-  if ($paragraph->hasField('field_hs_spotlight_bg') && !$paragraph->field_hs_spotlight_bg->isEmpty()) {
-    $bgValue = $paragraph->field_hs_spotlight_bg->first()->getValue()['value'];
-
-    $variables['spotlight_bg_class'] = Html::cleanCssIdentifier($bgValue);
+  if ($paragraph->hasField('field_hs_spotlight_bg') && $paragraph->get('field_hs_spotlight_bg')->count()) {
+    $bgValue = $paragraph->get('field_hs_spotlight_bg')->getString();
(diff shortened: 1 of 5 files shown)


docroot/modules/humsci/hs_layouts/patterns/spotlight/spotlight.html.twig
```diff
@@ -1,15 +1,12 @@
...
 {% if variant %}
   {% set variant_class = "hb-spotlight--image-#{variant}" %}
   {% set attributes = attributes.addClass(variant_class) %}
 {% endif %}
 
-{% set bg_class = background.field_hs_spotlight_bg['0']['#markup']|clean_class %}
-
-
 {% spaceless %}
-  <div{{ attributes.addClass(bg_class) }}>
```
rkelbel, line 12: This was removed as Line 1 was doing the same thing.

docroot/modules/humsci/hs_layouts/patterns/spotlight/spotlight.ui_patterns.yml
```diff
@@ -21,10 +21,6 @@ spotlight:
       label: "Overlay Text"
       type: mixed
       preview: "<h2>Etiam sollicitudin</h2><p>ipsum eu pulvinar rutrum, tellus ipsum laoreet sapien, quis venenatis ante odio sit amet eros. Lorem ipsum dolor sit amet</p><a class=\"decanter-button\">augue augue mollis justo.</a>"
-    background:
-      label: "Background"
-      preview: "Background"
-      type: mixed
```
rkelbel, line 27: I removed the Background region per request from Mike.

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -41,3 +43,15 @@ function hs_layouts_preprocess_container(&$variables) {
     $variables['#attached']['library'][] = 'hs_layouts/layout_builder_admin';
   }
 }
+
+/**
+ * Implements hook_preprocess_preprocess_pattern_spotlight().
+ */
+function hs_layouts_preprocess_pattern_spotlight(&$variables) {
+  $paragraph = $variables['context']->getProperty('entity');
```
rkelbel, line 51: The getProperty method had to be located within the ui_patterns module which is contrib, but was being utilized by the PatternContext inside of the hs_layout modules use of patterns. This field is protected by default.

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -41,3 +43,15 @@ function hs_layouts_preprocess_container(&$variables) {
...
+  if ($paragraph->hasField('field_hs_spotlight_bg') && !$paragraph->field_hs_spotlight_bg->isEmpty()) {
+    $bgValue = $paragraph->field_hs_spotlight_bg->first()->getValue()['value'];
+
+    $variables['spotlight_bg_class'] = Html::cleanCssIdentifier($bgValue);
```
pookmish:
```suggestion
  if ($paragraph->hasField('field_hs_spotlight_bg') && $paragraph->get('field_hs_spotlight_bg')->count()) {
    $bgValue = $paragraph->get('field_hs_spotlight_bg')->getString();
    $variables['attributes']->addClass(Html::cleanCssIdentifier($bgValue));
```

pookmish: I'm not sure if its `$variables['attributes']` or `$variables['#attributes']` with the `#`.

Using this way, you can remove the class from the twig template because it adds it to the attributes already.

pookmish (changes requested):

docroot/modules/humsci/hs_layouts/hs_layouts.module
rkelbel: Interesting. Thanks for this Mike, I'll take a shot with this.

kaseybon: @pookmish I can see it added to the attributes when I dump the variables to the page but it isn't getting showing up in the template. By the way, it did need the `#`.

pookmish: apparently `attributes` was an object, not an array. I've modified my suggestion and it looks to work on my side

kaseybon: Working on my side as well. Pushing a commit now.

pookmish (approved):

PR #1084, opened 2022-04-06
HSD8-1252: Cesta publications importer

joegl:
# READY FOR REVIEW
## Summary
Created a CSV publications importer for CESTA.

## Steps to Test
1. Import the migration configuration locally on a copy of the CESTA site.
    - `/admin/config/development/configuration/single/import`
1. Download the "CESTA Publications Template - 2022-04-05 - downloaded and adjusted" file from the ticket
    - https://stanfordits.atlassian.net/browse/HSD8-1252
1. Import the CSV file.
    - `/admin/structure/migrate/manage/default/migrations/custm_cesta_publications/csv-upload`
    - There will be some encoding errors – 27 of 92 will fail due to an iconv() notice:
      - `Notice: iconv(): Detected an illegal character in input string in Drupal\stanford_migrate\Form\StanfordMigrateCsvImportForm->fixLineBreaks() (line 290 of modules/custom/stanford_migrate/src/Form/StanfordMigrateCsvImportForm.php).`
      - This can be fixed with cleaning up the bad characters.
1. Open some of the newly created publication nodes and verify the data.
    - **NOTE:** My local was caching the /admin/content view. I had to rebuild my cache to see the imported Publications.
1. Review instructions for user below.
    
## Instructions for use
1. Go to the CSV upload page for the migration
    1. `/admin/structure/migrate/manage/default/migrations/custm_cesta_publications/csv-upload`
1. Download the example CSV file
1. Paste your data into the file
1. Make sure every row in the file has a unique ID number
    1. The ID number is used to align data for subsequent imports. If you need to change the data in the sheet and re-import, make sure the ID's align. Otherwise it will create a new publication or update the wrong publication.
1. Clean up the data in the file
    1. The Author's and Co-Author's data should be in the form of "First Last" and can have multiple comma delimited `,` values.
    1. The Publication Type data should only have 1 value.
    1. The Year data should only contain the year – full dates aren't accepted.
    1. The DH Method or Approach data can have multiple comma delimited `,` values.
    1. Remove or change any special characters.
    1. Globally, no values are being ignored; if the data should not be imported, leave it blank. E.g., values like "n/a" or "n/a for DH method" will be imported and need to be changed to blank values.

## PR Checklist


migrate_plus.migration.custm_cesta_publications.yml
```diff
@@ -0,0 +1,160 @@
...
+    -
+      plugin: callback
+      callable: trim
+    -
+      plugin: substr
+      start: 0
+      length: 4
+  field_hs_publication_year:
+    -
+      plugin: concat
```
pookmish, line 95: you'll want to have a `skip_on_empty` here before the `concat`

migrate_plus.migration.custm_cesta_publications.yml
```diff
@@ -0,0 +1,160 @@
...
+      delimiter: '-'
+      source:
+        - '@pub_year'
+        - constants/pub_md
+  field_hs_publication_author:
+    -
+      plugin: concat
```
pookmish, line 106: you'll want to have a `skip_on_empty` here before the `concat`

migrate_plus.migration.custm_cesta_publications.yml
```diff
@@ -0,0 +1,160 @@
...
+      callable: trim
+    -
+      plugin: entity_generate
+      entity_type: taxonomy_term
+      bundle_key: vid
+      bundle: hs_publication_type
+      value_key: name
+      ignore_case: true
+  field_hs_publication_link/uri: url
+  field_hs_publication_link/title: url
```
pookmish: It might be better to have some actual readable text here.

migrate_plus.migration.custm_cesta_publications.yml
joegl: @pookmish What should I put as the `source` ?

joegl: Does this look right?

```
  field_hs_publication_year:
    -
      plugin: skip_on_empty
      method: process
      source: '@pub_year'
    -
      plugin: concat
      delimiter: '-'
      source:
        - '@pub_year'
        - constants/pub_md
```

migrate_plus.migration.custm_cesta_publications.yml
joegl: Since I'm combining two source fields, how would I incorporate the skip?

pookmish: You can do 2 `skip_on_empty` items.

pookmish: Since this is a one time csv import, it might be less of a concern.

joegl: Yea I think the issue is if one field is blank, we'd still want the value from the other field. E.g., if there are no Co-Authors, we don't want to skip the Author. Unless I'm misunderstanding. I'd assume something like this would require values for both fields or else it would skip:

```
    -
      plugin: skip_on_empty
      method: process
      source: author
    -
      plugin: skip_on_empty
      method: process
      source: coAuthors
```

pookmish: oh ok... maybe in a pervious step, have a `default_value`? idk. We're probably over-engineering it for this.

joegl: I'm going to settle for a skip on the authors field but not co-authors. I think it makes sense that way.

joegl:
@pookmish I implemented feedback. 

There are a couple "issues" hanging:
1. Encoding can cause some issues with import (smartquotes/iconv issue)
2. The "Invalid header order" requiring a new spreadsheet to be created
3. How to handle skip on the author fields.

Since this is a one-time/single-use import I think I'd rather focus on assisting with the data formatting in the spreadsheet pre-import than try to resolve the current problems in the migration file itself. Although I am going to take note of the `iconv()` issue we talked about.

joegl:
This has been imported into the site.

PR #1103, opened 2022-05-02
shs-2: fix alignment of media asset edit button in wysiwyg

ryanhagerty:
# READY FOR REVIEW

## Summary
This fixes a bug introduced by CK Editor styles involving the 'Edit' button on media in the WYSIWYG editor.

## Urgency
_'low'_

## Screenshot
Screen Shot 2022-05-04 at 1 40 12 PM

## Steps to Test
1. Pull down the branch, navigate to the `humsci_basic` theme folder, and run `npm run watch` after installing the required modules.
2. In Drupal, go to a 'Flexible Page'. Add an image via the WYSIWYG. Hover over the image and verify the edit button is centered instead of off-centered and weird.

## Questions
- The `watch`/`build` command outputs changes for Airy/Traditional/Colorful. Should these be included or is that handled by a build process for the appropriate theme?
- The solution to this bug could have also been handled by modifying the custom module styles that controls the edit button. Is that approach ever taken? If so, can we get access to those repos?
- Are labels heavily-used here or optional?


mickey4k:
What did you have to do to be able to add an image to the WYSIWYG?  If I click the image button and upload an image, the required image styles don't get created and it never makes it to the WYSIWYG.  If I try to add an existing image, none of the images in the library load, even with Stage File Proxy enabled and configured.

ryanhagerty:
@mickey4k 

Yeah, there's def something wonky going on with files. Here's how I got around it:

I didn't upload a new image, I chose a pre-existing one. To get the files that already exist, I had to run an extra `lando` command, as the initial set up apparently doesn't grab the files. 

`lando blt drupal:sync:files --site=sparkbox_sandbox`
then
`lando drush @sparkbox_sandbox.local cr`

Try that and let me know if it works.

mickey4k (approved):
Works!  :tada: 
Looks like the merge conflicts are going to be fun to work out.  Good luck!  [hiding]

PR #1139, opened 2022-06-22
Shs 50 twig menu poc

ryanhagerty:
# READY FOR REVIEW

## Summary
This is the new mega menu V2.

## Steps to Test
[ ] Go to the music dev site (https://music-dev.stanford.edu/). 
[ ] There is a new module there called `Stanford HumSci Mega Menu Config`. This will be already be installed manually on music so that the team can look at the work, but it should be off by default (check another site to confirm such as: `hs-sandbox-dev` or `islamicstudies-dev`. 
[ ] Once the module is installed, or if you're on the music site, confirm that there's a new button under 'Appearance' called `HumSci Mega Menu Settings`. Going there will turn on and off the mega menu. Feel free to turn it on / off and save to test.
[ ] With it on, confirm the mega menu is there and not seriously broken (but several people have already checked it out).

## Screenshots
Screen Shot 2022-07-13 at 1 07 00 PM


ryanhagerty:
@pookmish Head's up - this PR combines a BE and FE into one due to a shortened timeline. LMK if you need anything, thanks.

docroot/modules/humsci/hs_megamenu/config/install/hs_megamenu.settings.yml
```diff
@@ -0,0 +1,2 @@
+hs_megamenu:
+  use_hs_megamenu: 0
```
pookmish: If this is needed, I'd rather move it into a config pages like on `/admin/config/site-options`. You can simply add a field and then any config ignore or config readonly won't need to adjusted to accommodate.

pookmish (changes requested):
after adding any field to the main menu items, the template changes to use: `modules/contrib/menu_item_extras/templates/menu--extras.html.twig`
At this time this entire functionality in this PR does nothing and none of the template changes nor the styles would have any affect. The display after the template changes is this: 
image

docroot/modules/humsci/hs_megamenu/config/install/hs_megamenu.settings.yml
codechefmarc: @pookmish - for the setting, I wasn't aware you were using config_pages module, so we could certainly move that setting in there if that is best practice for Stanford. Let me know if this a blocker for release.

As for the menu item extras issue, I was told that it was added as a preemptive measure ahead of the work to be done on a separate task for adding extra things to the new megamenu. So, if we can, let's disable that module for now and when work starts on that piece, we can definitely test adding items and making sure the menu doesn't break.

pookmish: There's no sense in merging something if it's does nothing. The template changes and the styles will have no affect on the megamenu. Currently it makes it look like a mega menu only. It does nothing beyond that and once more work begins with the mega menu in terms of fields etc, all of this will be technical debt

codechefmarc:
@pookmish - I've moved the config setting for turning on and off the mega menu from the custom module and into config_pages module under Site Options. Let me know if this works for you. Thanks!

docroot/modules/humsci/hs_megamenu/hs_megamenu.module
```diff
@@ -0,0 +1,22 @@
...
+ * Target main menu and add a twig variable to show/hide the new mega menu.
+ */
+function hs_megamenu_preprocess_menu(&$variables) {
+
+  if ($variables['menu_name'] == 'main') {
+    $config_pages = \Drupal::service('config_pages.loader');
+    $useMegaMenu = $config_pages->getValue('hs_site_options', 'field_en_mega_menu');
+    if ($useMegaMenu) {
+      $variables['use_hs_megamenu'] = $useMegaMenu[0]['value'];
+    }
```
pookmish: Simplify it.
```suggestion
     $variables['use_hs_megamenu'] = (bool) $config_pages->getValue('hs_site_options', 'field_en_mega_menu', 0, 'value');
```

pookmish (changes requested):

codechefmarc:
@pookmish - Refactored.

PR #1141, opened 2022-07-01
Security fix for composer >= 2.2.0

generalredneck:
# READY FOR REVIEW

## Summary
See https://getcomposer.org/allow-plugins
as of last night, allow-plugins is required to have a list of plugins that are accepted each with a true false value. Simply making this value "true" is no longer supported and cause composer install to fail with the following error
image

## Need Review By (Date)
ASAP

## Urgency
HIGH

## Steps to Test
1. run composer install
2. Validate that it finishes successfully.

## PR Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -283,7 +283,16 @@
     "config": {
         "sort-packages": true,
         "discard-changes": true,
-        "allow-plugins": true
+        "allow-plugins": {
+            "composer/installers": true,
+            "simplesamlphp/composer-module-installer": true,
+            "davidbarratt/custom-installer": true,
+            "drupal/console-extend-plugin": true,
+            "acquia/blt": true,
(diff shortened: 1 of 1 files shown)


pookmish:
https://github.com/composer/composer/issues/10907#issuecomment-1172186852
Update composer and this will be resolved.

generalredneck:
@pookmish Fair nuff. I made the assumption that the documentation was outdated because of this line. Sorry I didn't go look to see if it was an active bug in Composer itself. See where the [documentation](https://getcomposer.org/allow-plugins)  says
 
> You can also set the config option itself to false to disallow all plugins, or true to allow all plugins to run (NOT recommended).

Thank you for the quick response and we will resolve this on a per machine basis.

PR #1208, opened 2022-10-19
SHS-91: Codeception: Test video embeds and captions

codechefmarc:
# READY FOR REVIEW

## Summary
Adds a test to verify that video embeds can be added with a caption and that captions show up on the public view.

## Urgency
low

## Steps to Test
1. Run codeception tests either via CI or locally
2. Verify that the testVideoEmbed test passes.

## PR Checklist


callinmullaney (approved):
@marcb4k Tests are passing on local and in CircleCI. The steps match the ticket requirements as well. We might need to follow up on this test to actually save the node and verify the video displays but this isn't part of the existing requirements. Also, that additional save step needs to consider other things like text-formatting which is out of scope for this ticket anyways.

Nice work!

PR #1216, opened 2022-11-02
shs-98: move around relative position for photo caption

ryanhagerty:
## Summary
This removes the `relative` position in a stylesheet for the photo caption and instead places it as a class name on the templated level.

## Steps to Test
1. Go to a page with a photo caption. Confirm the caption element now has the `hb-position-relative` class.
2. Confirm the photo captions behave as before.

## Screenshots
Screen Shot 2022-11-02 at 9 22 47 AM

## Notes 
The ticket description (https://fourkitchens.atlassian.net/browse/SHS-98) was a little confusing for me as it's steeped in some historical context. I found a couple different photo captions and they both picked up the template, however I'm unsure as to **all** the use cases where the caption is employed.


callinmullaney (approved):
:+1:

PR #1218, opened 2022-11-02
shs-122: update accordion state to use correct mixin

ryanhagerty:
## Summary
Fixes an issue with border colors on the alternative accordion presentation. The original mixin does not respect color pairings, only a pairing for the theme's default color theme.

## Steps to Test
1. Go to the accordions page (`/accordions`). Edit the layout, under the Main Region, add the `hb-accordion-secondary` class to the Field Formatter Class area.
2. Visit the accordion page and verify that it's using the correct layout (border color of the theme's secondary color, hover is the background primary color etc)
3. Go to the theme settings (`/admin/appearance/settings/humsci_colorful`) under Color Pairing change the color to a different color. Go back to the accordions page and verify it changed to the appropriate color. Do this a few times to be sure.
4. Repeat these steps for the Traditional site.

## Screenshots
Screen Shot 2022-11-02 at 2 28 49 PM


callinmullaney (changes requested):
@ryanhagerty Code changes here are definitely on the right path to simplify the styling a bit more. I am seeing some differences in this branch compared to the ticket requirements. Let me know if you have any questions/concerns with these. I did an in-browser mockup of what the changes should look like.
1. The active state for should return the accordion label and icon back to the text color with the standard background-color for the block (transparent seems to work as well)
2. The active state matches the border styles and switches them to primary color. Note: there is no separation border between the `<summary>` and accordion content
3. The active state when the user hovers over the summary to close the accordion should only change the text color and color of the icon to `secondary` 

(screenshot shows open accordion, closed accordion, and open accordion with hover over the `<summary>`)
Screenshot from 2022-11-07 12-14-20

ryanhagerty:
@callinmullaney 

For this one looks like just some false positives.

Percy colorful:
https://percy.io/da667d23/suhumsci/builds/22749911/changed/1268880054?browser=edge&browser_ids=18%2C25%2C30%2C31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=768&widths=576%2C768%2C1200

Percy traditional:
https://percy.io/da667d23/suhumsci/builds/22749927/changed/1268882139?browser=chrome&browser_ids=18%2C25%2C30%2C31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1200&widths=576%2C768%2C1200

ryanhagerty:
@callinmullaney Thanks for checking it out. Updated everything. I'm confused about what it should actually look like now, but trust you have better context, so I think it should be good.

callinmullaney (approved):
Code looks good, feedback was implemented, and no demo feedback from the client. I checked out the percy tests and confirmed the false positives. Seems that webkit has some slight differences with how it renders fonts and borders? The flags in the test were so small it's not worth diving into further. I went ahead and approved them as this would be an unnecessary step for the client.

Merging into the release. Nice work!

PR #1228, opened 2022-11-29
shs-87: set up base reduced motion option

ryanhagerty:
## Summary
This adds `prefers-reduced-motion: reduce` to all elements.

## Steps to Test
1. Go to a Colorful page with animations. Here is a good example: `components/hero-images-main-region`. Verify animations play (transitions, sliders, etc.)
2. Depending on what OS you use, turn on Reduced Motion in your Accessibility preferences. For Macs, open Accessibility -> Display. A screenshot is included below.
3. Visit the same page from above. Verify animations/transitions do not play but everything is still functional.
4. Do The same for Traditional.
5. Poke around a few other pages and verify functionality still works. I tried about 20 pages, and everything seems fine. While this touches a lot of things, it's low risk due to the < 0.1% of people who would have this setting on. 

## Screenshots
Screen Shot 2022-11-29 at 3 40 55 PM

## Notes
- Will need to change base branch after sprint branch is set up
- The ticket directly mentioned doing this on the slider, but it makes more sense to apply to *all* transitions and animations.


su-sws-jira:

Links:
- https://pr1228-xl4l51hazwlqlhfthsf5pbfukgymwmrc.tugboatqa.com
- https://hs-colorful-xl4l51hazwlqlhfthsf5pbfukgymwmrc.tugboatqa.com
- https://hs-traditional-xl4l51hazwlqlhfthsf5pbfukgymwmrc.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/63866f5632d30df7ac8990af

callinmullaney (changes requested):
@ryanhagerty The changes here is likely the best approach to getting this theme to respect the user's OS accessibility setting for reduced motion. Ideally, this should be done at the implementation level instead of forcing this override BUT doing some research it appears there are 103 references to animation styles in the theme. At that point I didn't even check the transition styles. Adding this media query at the implementation level would results in a couple hundred additional lines and we don't need to bloat things any further.

I only found one implementation instance where these styles likely wouldn't apply and think you should account for this as well. https://github.com/SU-HSDO/suhumsci/blob/develop/docroot/themes/humsci/humsci_basic/src/scss/utilities/_wysiwyg-text-area.scss#L178

ryanhagerty:
@callinmullaney Made the change. I'm unfortunately not able to test the new addition as I'm having some local lando issues with this site right now 🙃

callinmullaney (approved):
Does the trick! Thanks for making the change.

PR #1231, opened 2022-12-06
Shs 125 some accessibility improvements

ryanhagerty:
## Summary
This addresses some of the accessibility violations for SHS-125. The ticket has a lot of bullet points that were mainly split out into their own tickets.

This PR fixes two items:
* It revamps the original menu to use buttons instead of links when there is no link
* It adds focus outline to form elements

## Steps to Test
1. Visit a site with the original menu. Verify it still works across all screen sizes. Verify that visually nothing has changed. Also check that the navigation works with keyboards and with VoiceOver
2. Tab through an element with a form field (the search works) and verify that it now has an outline

## Screenshots
Screen Shot 2022-12-06 at 8 52 00 AM

Screen Shot 2022-12-06 at 8 52 18 AM

## Notes
We'll probably want to include this in the next release and not the one going out tomorrow.


docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.forms.scss
```diff
@@ -12,6 +12,9 @@
   }
 }
 
+/* @TODO: SHS-138
```
callinmullaney, line 15: :clap: nice!

callinmullaney (approved):
@ryanhagerty This PR definitely improved upon the keyboard navigation for the menus and the code changes (with the add TODO) are all spot-on from my perspective. I also tested the mega menu on the music site and it functions as you would expect. 

The only thing I can see might cause feedback from the client or a bit of rework is the `primary` color being used as the outline. In most cases this seems to be the ideal choice but (hs_colorful for example) some headers are already a shade of the primary and the contrast is hard to see. At any rate, this is good to merge into our release :tada:

PR #1247, opened 2023-01-17
Shs 104 megamenu animation

ryanhagerty:
## Summary
This adds transitions effects to the new megamenu. 

## Steps to Test
1. Visit a site with the new mega menu enabled. 
2. At desktop, hover over the top level menu items. The underline should animate.
3. Click a menu item with a dropdown. The child menu container should animate. 

**Notes**
* This is pointing to the release branch to avoid a bunch of files below. When there's a new sprint branch, we can change that.
* I did this ticket out of order. Other tickets weren't ready, and I felt I could get this one done with the info it had.
* This would be _slightly_ better if I transitioned the translate Y axis, but that would involve a lot of refactoring to 'hide' it behind everything else.


callinmullaney (dismissed):
@ryanhagerty From a code perspective this looks good and probably the same way I would have tackled it. I have a couple designer/UX opinions but take these with a grain of salt and I defer to Tori's eye on this:
1. Animation seems a bit fast compared to the other sites linked in the ticket. They have a slightly smoother/slower animation that just "feels" better.
2. The margin gaps on `megamenu__item` make the roll-over of multiple items feel a little disjointed. Rolling off an item feels too immediate like there should be a 5-10px padding around each item. I noticed this when moving my cursor to the arrow icon (which is right on the edge) and the hover bar was a bit flickery (new word!). 

Approving code but lets let the team way in and see if they have additional changes we can roll into this PR.

ryanhagerty:
@callinmullaney This should be good to go if we want to get a release branch going.

callinmullaney (approved):
code and help text addition looks good. Approve.

callinmullaney:
@pookmish No release branch as this is the only ticket we're pushing through from last week. I pulled the release branch and recompiled sass. Ready to merge when you have a moment.

PR #1256, opened 2023-02-07
SHS-128: Properly size images

ryanhagerty:
## Summary
This updates many image styles for components and views-based components.

## Need Review By (Date)
- 5/10

## Urgency
medium

## Steps to Test
1. The exhaustive way of testing would be to create a flexible page and add every component with an image and every view component with an image and test the breakpoints.
2. If that's not possible, I would pick some higher level components like slideshows, heroes or views and test those across breakpoints.

## Notes
- There was a generic responsive image style that I refined. This only scaled the width, so I kept it without cropping since the context of this image can get applied in many places. I also kept the image styles at the 'breakpoints' rather than the actual image size (due to container padding).
- The Hero Layered Slider - We talked about this briefly, but this is the one that is controlled by a module that forces a certain view. I made the best compromise I could for better image styling.
- The postcard has three versions. I did styles for all versions
- There were a few views I couldn't trigger displaying, even when providing an argument based on the view (if there was one).
- I erred on the side of components may change in the future rather than reusing styles for the sake of reusing. 
- One component allowed upscaling, which I continued to allow because I don't know the context on why that decision was made (every other upscaling was with CSS).
- This does not address collections. I'm not sure this is possible - maybe with some heavy handed preprocessing? The image has no context that it's in a collection. I could make a view I guess, but then the collection also has many possible column options, so there's that as well. Ultimately, I don't think the juice is worth the squeeze. 
- I would like to give Stanford as much time as possible to test this. I'm not sure the normal release schedule allows for that. We may have to commandeer a server to discuss with them. 

## Potential bugs discovered along the way
- The Hero Layered Slider image has its height set by CSS based on the monitors viewport height. This could create some funky images if someone has a larger monitor / vertical arrangement.
- Testimonial - if you click 'none' as the alignment when adding, the whole quote is borked.
- on `/events/colloquium-series` if the text isn't long enough the first event wraps around the image
- On photo albums at mobile the image doesn't take up the full width. That's probably a design decision, but every other component takes up the full width when an image is present.


callinmullaney:
Percy tests:
1. https://percy.io/da667d23/suhumsci/builds/25504926
2. https://percy.io/da667d23/suhumsci/builds/25504947

callinmullaney:
@pookmish This PR is ready for you to take a look. Sending this over to you a bit earlier and separate from our normal weekly release as it contains a lot of configuration changes. 

We've (4K, Tori, Joey, Marion) have been testing this for awhile on Acquia's dev environment and the end results are positive (see tests below). I also incorporated some WebP and ImageMagik compression changes that were originally part of the closed Sprint 21 branch. Let me know if you have any thoughts/questions. 

Tests: 
    Dev (image files = 408kb) - https://www.webpagetest.org/result/230407_AiDc8S_6YH/
    Prod (image files = 535kb) - https://www.webpagetest.org/result/230407_BiDcEC_6WV/ 
    Dev (image files = 73kb) - https://www.webpagetest.org/result/230407_AiDcVY_6YQ/
    Prod (image files = 342kb) - https://www.webpagetest.org/result/230407_AiDcQ2_6Z0/

callinmullaney (approved):
Approving my review for functional and code. Functional review was also performed by Tori and team on the Acquia dev environment.

config/default/webp.settings.yml
```diff
@@ -1,3 +1,3 @@
 _core:
   default_config_hash: 0ube66wBYymHixNiMD-uyY8Tjo8alYV3NII6dv5iBJ8
-quality: 85
+quality: 80
```
callinmullaney, line 3: Note: These compression changes (although minor) contribute to the performance gains achieved in the linked performance tests. These were ported over from a previously rejected PR.

config/default/imagemagick.settings.yml
```diff
@@ -1,6 +1,6 @@
 _core:
   default_config_hash: cp-t_XKrU3l5dl9BIOiM-favydmIDTQhL11eF-1MjRc
-quality: 85
+quality: 80
```
callinmullaney, line 3: Note: These compression changes (although minor) contribute to the performance gains achieved in the linked performance tests. These were ported over from a previously rejected PR.

callinmullaney:
@pookmish I updated this PR with the latest changes from the release branch. Tests seem to mostly be passing accept `run-deploy-branch`. CI seems to be noting uncommitted changes to `docroot/.htaccess` and I'm not exactly sure where this is coming from as we're only addressing image configuration. Any ideas on what could be causing this automated test failure?

callinmullaney:
@pookmish I merged in the latest from develop and re-ran the CI tests. All are passing now.

pookmish (approved):
This is a big change, but just changing image displays. Keep a close eye on how it affects the sites when it goes out.

PR #1260, opened 2023-02-14
shs-109: refactor page scroll animation

ryanhagerty:
## Summary
Refactors the page scroll animation JS

## Steps to Test
1. On both Colorful and Traditional, go to `/components/hero-images-main-region` and the homepage and ensure that the heroes that used to animate still animate correctly.

**Notes**
I kept the preexisting array of items watch for animation. When reducing it to the parent component, I discovered that different animations fire based off image, text, etc. so most of the classes all relate to two parent elements (the overlay and gradient heroes). I also kept the ones I wasn't sure (`hs-font-lead`, `hs-font-splash`) because I can see they animate things in the code, but I don't know where they exist.


callinmullaney (approved):
@ryanhagerty this is significantly refactored and as far as I can test it appears to be working in an identical manner as before (albeit much simpler). 
I did some digging and the two classes you mentioned are utility classes. They have to be manually applied in layout builder and appear to affect callout boxes, specific instances of caption-credits (although this appears to be more of a font-override than animation), or specific instances of `.hs-font-splash ~ .hs-font-lead` in the traditional theme. That last instance I was not able to test/replicate but nothing in your code changes would appear to affect it any different than before.

callinmullaney:
I'm approving and putting on hold for now. I'll merge this into our release branch later in the week.

PR #1286, opened 2023-04-03
Ckeditor5 class stylesheet

ryanhagerty:
## Summary
This allows for class styles to represented in the CKEditor5 WYSIWYG editor. 

## Steps to Test
1. After pulling down the branch and importing config, make sure to disable the `insert` module. It is incompatible with CK5 and will result in a WSOD.
2. Create a new flexible page. Click the 'Source' button, and copy/paste the code below so that you have all the associated classes and don't have to create any.
3. Click the 'Source' button again to view the rendered styles in the WYSIWYG. Compare the styles you see to this page: https://hs-colorful-stage.stanford.edu/test-page-ck5 to make sure they line up (edit the page to get to the CK4 WYSIWYG - this page gets wiped every time there's a release BTW).
4. Save the page and make sure the styles are accurate.
5. Do the same for traditional.
6. Make sure there's no 'leaking' styles that affect the admin theme (screenshot below).

## Notes
* This branch is forked off the main `ckeditor5` branch
* Below is a screenshot of what the admin theme would look like if we used the blanket stylesheet like we did with CK4. Because CK5 is **not** in an iFrame, we need to isolate the styles, which requires us to create a separate stylesheet in Grunt just for the CK5 editor.
* Another issue with the styles not being isolated is that the `:root` pixel size between the main theme and admin theme are different. This means **all** rem definitions in the `humsci_basic` are not compatible with the admin theme. This is why the separate `imports.scss` stylesheet recreates some of the associated styles with `px`.
* There are bugs with CK5 that are not associated with this PR. The two I noticed: the editor `textarea` will be small until you start adding content; the editor's buttons behave like a sticky nav. This can sometimes cause funky behavior.
* The old version of CK4 added dotted borders to tables. I did not recreate this style because it was an artifact of CK4 overriding SHS `table` class styles. I made the editor styles look like the final user-facing styles.
* I added commands for building the CK5 stylesheets. I included this command in the overall `watch` function. I did this because it's technically the 'right' thing to do so that everything's always automatically included, but personally I think it should be removed from the watch command. It really slows down CSS compiling and the amount of times `imports.scss` will need changing will likely be minimal. 

## Screenshots
What the admin theme would look like if we just used the blanket stylesheet like in CK4:
Screenshot 2023-04-03 at 12 54 37 PM

## Code
```
<p>
    <a class="hs-button" href="https://www.google.com">Test Primary Button</a>
</p>
<p>
    <a class="hs-button--big" href="https://www.google.com">Test Big Primary Button</a>
</p>
<p>
    <a class="hs-secondary-button" href="https://www.google.com">Test Secondary Button</a>
</p>
<p>
    <a class="hs-more-link" href="https://www.google.com">Test More Link</a>
</p>
<p>
    <a class="hs-external-link" href="https://www.google.com">Test External Link</a>
</p>
<p>
    <a class="hs-file-link" href="https://www.google.com">Test File Link</a>
</p>
<p>
    <a class="hs-pdf-link" href="https://www.google.com">Test PDF Link</a>
</p>
<p>
    <a class="hs-private-link" href="https://www.google.com">Test Private Link</a>
</p>
<p>
    &nbsp;
</p>
<p class="hs-font-lead">
    Testing lead font
</p>
<p class="hs-font-splash">
    Testing splash font
</p>
<p class="hs-short-line-length">
    Test short line length
</p>
<p class="hs-breakout-box">
    Test breakout box
</p>
<p class="hs-caption">
    Test caption
</p>
<p class="hs-credits">
    Test credits
</p>
<p class="hs-emphasized-text">
    Test emphasized text
</p>
<table class="hs-table--borderless">
    <tbody>
        <tr>
            <td>
                table here 1
            </td>
            <td>
                column
            </td>
        </tr>
        <tr>
            <td>
                table here 2
            </td>
            <td>
                oh look another column
            </td>
        </tr>
        <tr>
            <td>
                afsdfsda
            </td>
            <td>
                and another
            </td>
        </tr>
    </tbody>
</table>
<table>
    <tbody>
        <tr>
            <td>
                table here 1
            </td>
            <td>
                column
            </td>
        </tr>
        <tr>
            <td>
                table here 2
            </td>
            <td>
                oh look another column
            </td>
        </tr>
        <tr>
            <td>
                afsdfsda
            </td>
            <td>
                and another
            </td>
        </tr>
    </tbody>
</table>
```


ryanhagerty:
@callinmullaney 

Regarding the anchor plugin: I think I'm going to need a BE create this module for a couple reasons.

- CKEditor5 requires plugins to be structured in a certain way ([see docs here](https://api.drupal.org/api/drupal/core%21modules%21ckeditor5%21ckeditor5.api.php/group/ckeditor5_architecture/10.0.x#plugins)). So I don't think the structure it needs is compatible with that module gist you sent over. 

Maybe we could make the plugin module a dependency of the larger proposed HS Admin module?

- I got the module to install without errors (yea), but it does not display in the admin side of CK5 config. Looking at other CK5 modules, [like this for example](https://git.drupalcode.org/project/ckeditor_mentions), I think more PHP is needed to actually implement the plugin rather than drag and drop a JS library (like assigning a button for content authors, config options, etc.)

- The plugin is structured from a modern JS perspective rather than a bundled JS file like jQuery for example. So if you [checkout the plugin](https://github.com/bvedad/ckeditor5-anchor), there is no `dist`, there is no `build` command, etc. It's several JS and CSS files. I think I could experiment with this part to get it working, but I unfortunately can't get to that point.

- I would love to see how the BE portion of this ends up constructed so I can try to learn for future reference.

PR #1299, opened 2023-04-24
shs-4874: color band limit width when sidebar present

ryanhagerty:
## Summary
This limits the full width color bands width when a sidebar is present

## Steps to Test
1. Go to the color bands page `/components/color-bands`. It has a mix of full width and limited width color bands. Observe that the full width doesn't fully extend because a sidebar is present.
2. Create a new flexible page and add a mix of full width and limited width color bands. Save. Observe that the full width version extends all the way because there is no sidebar.

## Screenshots
Screenshot 2023-04-24 at 11 29 16 AM

Screenshot 2023-04-24 at 11 29 06 AM

## Notes
- The ticket didn't mention this, but here's a scenario I have a question about: when a sidebar column is present, it's only present as a 'sidebar' at desktop sizes. Do we want to have the sidebar 'full-width' at lower breakpoints?
- The selector is `.hb-three-column--one-sidebar`. I have never seen a page with more than one sidebar. I'm guessing this is a naming convention and there's not pages with two or three columns floating about?


callinmullaney (approved):
:+1: pretty easy fix. Nice work!

PR #1301, opened 2023-05-01
fix: hotfix for stray query parameters attached to url

ryanhagerty:
## Summary
This fixes an issue where stray query parameters like `?check_logged_in=1` would trigger search functionality.

## Steps to Test
1. Go to the colorful homepage with this appended `/?check_logged_in=1`. Check the console to ensure you don't see an error related to `cannot read properties of undefined (reading 'toLowerCase')`
2. Reload the homepage without the `/?check_logged_in=1` and ensure there's no console errors.
3. Test the search expand functionality to make sure it still works. If you still have content with 'pepperoni pizza' in timelines and accordions, search for that.
4. Click on those pages and make sure the appropriate content expands. 
5. Check the console to make sure there's no errors related to `cannot read properties of undefined (reading 'toLowerCase')`


callinmullaney (approved):
@ryanhagerty Re-tested the functionality for Accordions and Timelines to verify they are still opening if the query parameter search matches what is inside. I also test non-search query parameters and I'm not seeing the JS error that is present on the reported economics site. Nice work!

pookmish (approved):

PR #1313, opened 2023-06-15
HSD8-1475: Pull all departments for capx importer profiles

joegl:
# READY FOR REVIEW

## Summary
- Increased the item limit from 1 to 10 on the Person department field.
- Modified the importer to pull more than 1 department.

## Need Review By (Date)
_['10/30', 'asap', etc.]_

## Urgency

## Steps to Test
1. I'm not sure how you want to test this, but you'll want to test an import of the `migrate_plus.migration.hs_capx` migration on a capx profile with multiple departments. I had been using the `BIOLOGY:BIO-FACULTY` workgroup and [Barbara Block](https://profiles.stanford.edu/barbara-block) as an example (she's in Oceans and Biology).

## PR Checklist

### DIFF
--- a/config/default/field.storage.node.field_hs_person_department.yml
+++ b/config/default/field.storage.node.field_hs_person_department.yml
@@ -13,7 +13,7 @@ settings:
   target_type: taxonomy_term
 module: core
 locked: false
-cardinality: 1
+cardinality: 10
 translatable: true
 indexes: {  }
 persist_with_no_fields: false
--- a/config/default/field.storage.node.field_hs_person_department.yml
+++ b/config/default/field.storage.node.field_hs_person_department.yml
@@ -13,7 +13,7 @@ settings:
   target_type: taxonomy_term
 module: core
 locked: false
-cardinality: 10
+cardinality: -1
 translatable: true
 indexes: {  }
 persist_with_no_fields: false
--- a/config/default/migrate_plus.migration.hs_capx.yml
+++ b/config/default/migrate_plus.migration.hs_capx.yml
@@ -295,7 +295,7 @@ process:
         value: label/text
     -
       plugin: flatten
-  field_hs_person_department:
+  department:
     -
       plugin: sub_process
       source: department
@@ -303,6 +303,12 @@ process:
         value: department
     -
       plugin: flatten
+  field_hs_person_department:
+    -
+      plugin: null_coalesce
+      source:
+        - '@department'
+        - org_title
     -
       plugin: skip_on_empty
       method: process
--- a/config/default/migrate_plus.migration.hs_capx.yml
+++ b/config/default/migrate_plus.migration.hs_capx.yml
@@ -306,15 +306,25 @@ process:
         value: department
     -
       plugin: flatten
+  orgTitle:
+    -
+      plugin: skip_on_empty
+      method: process
+      source: org_title
+    -
+      plugin: explode
+      delimiter: '|/'
   field_hs_person_department:
     -
       plugin: null_coalesce
       source:
         - '@department'
-        - org_title
+        - '@orgTitle'
     -
       plugin: skip_on_empty
       method: process
+    -
+      plugin: multiple_values
     -
       plugin: entity_generate
       entity_type: taxonomy_term
--- a/config/default/migrate_plus.migration.hs_capx.yml
+++ b/config/default/migrate_plus.migration.hs_capx.yml
@@ -108,8 +108,8 @@ source:
       selector: education
     -
       name: department
-      label: Departement
-      selector: academicOffices/0/department
+      label: 'Department'
+      selector: academicOffices
     -
       name: cv_document
       label: 'CV Document'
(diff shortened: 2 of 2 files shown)


joegl:
@pookmish This is close. I have it pulling multiple departments. The issue I'm having is there used to be a `null_coalesce` step which as far as I can tell would use the `org_title` value if the `department` value was empty/null, and I have not been able to rebuild that into the migration.

I originally tried to just change the selector from `selector: academicOffices/0/department` to `selector: academicOffices/department` because from my memory that would build an array of all the `department` values in `academicOffices`, but I was never getting more than 1 value that way. 

Also for reference so you don't have to go digging, this is what the JSON array looks like (I X'd out personal info):
```
"academicOffices": [
        {
          "address": "Pick up",
          "affiliationType": "University - Faculty",
          "city": "Stanford",
          "department": "Biology",
          "displayAddress": {
            "html": "Pick up",
            "text": "Pick up"
          },
          "officeName": "office",
          "phoneNumbers": [
            "XXXXXX",
            "XXXXXX"
          ],
          "position": "Professor",
          "state": "XXXXXX",
          "type": "academic",
          "zip": "XXXXXX"
        },
        {
          "affiliationType": "University - Faculty",
          "department": "Oceans",
          "position": "Professor"
        }
      ],
```

config/default/migrate_plus.migration.hs_capx.yml
```diff
@@ -108,8 +108,8 @@ source:
       selector: education
     -
       name: department
-      label: Departement
-      selector: academicOffices/0/department
+      label: Department
+      selector: academicOffices
```
pookmish, line 112: What happens if you do `academicOffices/*/department` without changing the process methods?

joegl: Definitely will try that. I think I got used to the simpleXML selectors, and this is a JSON import. I didn't know about the asterisk/wildcard (is there a list of these somewhere?).

joegl:
@pookmish It also looks like tests aren't going to pass here until that blt/UserConfig issue is resolved.

config/default/migrate_plus.migration.hs_capx.yml
joegl: @pookmish I tried using `academicOffices/*/department` as the selector with the original process methods, but didn't have any luck.

joegl: Yea it doesn't look like the default JSON `data_parser_plugin` for `migrate_plus` includes any special selectors. There is an open issue to add a JSONPath parser plugin to allow for that: https://www.drupal.org/project/migrate_plus/issues/3007709

pookmish: oof, 5 years old. ouch. Well that was worth a shot. guess we'll have to do the subprocess approach.

joegl: Yea I'm a little surprised the JSONPath support hasn't got more traction.

joegl: I was able to re-implement the `null_coalesce`.

joegl:
> @pookmish It also looks like tests aren't going to pass here until that blt/UserConfig issue is resolved.

I merged your fix in here, so it should all be set.

config/default/field.storage.node.field_hs_person_department.yml
```diff
@@ -13,7 +13,7 @@ settings:
   target_type: taxonomy_term
 module: core
 locked: false
-cardinality: 1
+cardinality: 10
```
pookmish: we should just make this unlimited -1

config/default/migrate_plus.migration.hs_capx.yml
```diff
@@ -297,9 +297,12 @@ process:
       plugin: flatten
   department:
     -
```
pookmish, line 299: Should keep the `skip_on_empty` to prevent any errors.

pookmish (changes requested):
This doesn't seem to populate all departments. Currently it still only imports a single department term

joegl:
> This doesn't seem to populate all departments. Currently it still only imports a single department term

Did you make sure to import the configuration that increases the field limit from 1 to 10? Maybe that's not getting imported by default. I can pull this down fresh another time and test it.

pookmish:
> Did you make sure to import the configuration that increases the field limit from 1 to 10? Maybe that's not getting imported by default. I can pull this down fresh another time and test it.

yes. The taxonomy isn't even getting created for the multiple values.

config/default/migrate_plus.migration.hs_capx.yml
joegl: I actually was wondering about this as it relates to the `null_coalesce` process.

~~From my understanding, `null_coalesce` will run through the sources value specified in order and pick the first one that's not `NULL`. So in this migration, I assumed it was meant to use the `org_title` value as the Department if the `department` value was `NULL`. However, because there's a `skip_on_empty` above the `null_coalesce` wouldn't it never fire/work? Because if the department value is empty, it will skip it and never get to the `null_coalesce` process.~~

Ignore this. I was reading the migration in the wrong order.

PR #1321, opened 2023-07-06
HSD8-1473: Add News RSS bundle

joegl:
# READY FOR REVIEW

## Summary
- Added `hs_entities` module with new `hs_entity` Entity Type
- Added News RSS bundle to `hs_entity` Entity Type
- Removed News RSS ECK bundle and Importers collections ECK Entity type
- Added update hooks to facilitate all of the above and migrate relevant data from the old to the new

## Need Review By (Date)
_['10/30', 'asap', etc.]_

## Urgency

## Steps to Test
1. Checkout the latest release.
    - `git checkout 9.7.1-release`
1. Install the default site locally
    - `drush si su_humsci_profile -y`
1. Create a few News RSS ECK items
    - Create here: `/admin/structure/migrate/news-rss`
1. Checkout this branch
    - `git checkout HSD8-1473--add-news-rss-bundle`
1. Run database updates
    - `drush updatedb -y`
1. Verify
    - All the News RSS items you created still exist and have been migrated to the new Entity type (`/admin/structure/migrate/news-rss`).
    - The News RSS ECK Bundle has been removed.
    - The Importers Collections ECK Entity type has been removed (`/admin/structure/eck`).

## PR Checklist


pookmish:
Gotta have those mews.

docroot/modules/humsci/hs_entities/src/Entity/HsImporterType.php
```diff
@@ -0,0 +1,68 @@
...
+namespace Drupal\hs_entities\Entity;
+
+use Drupal\Core\Config\Entity\ConfigEntityBundleBase;
+
+/**
+ * Defines the HumSci Importer type configuration entity.
+ *
+ * @ConfigEntityType(
+ *   id = "hs_importer_type",
+ *   label = @Translation("HumSci Importer type"),
```
pookmish: this should be agnostic of importers. it will be used on content type fields for events, news, publications, etc.

docroot/modules/humsci/hs_entities/src/Entity/HsImporter.php
```diff
@@ -0,0 +1,57 @@
...
+
+use Drupal\Core\Entity\ContentEntityBase;
+use Drupal\hs_entities\HsImporterInterface;
+
+/**
+ * Defines the humsci importer entity class.
+ *
+ * @ContentEntityType(
+ *   id = "hs_importer",
+ *   label = @Translation("HumSci Importer"),
```
pookmish: This should be agnostic like the other entity type

pookmish (changes requested):
I did a quick glance and noticed a pretty big change needed. let me know when you have that part refactored and I'll review closer

docroot/modules/humsci/hs_entities/src/Entity/HsImporterType.php
joegl: I was curious what your thoughts were gonna be on this.

Right now there are 4 different ECK Entity types (Importer collection, Publication collection, etc.,). When you create an Entity Type, you have to specify what base fields are included (like label, status, created, changed, etc.,), whether it's revisionable or translatable, whether it has bundles or not, etc., I originally was going to create one Entity type for all the ECK bundles, but because there are 4 different ECK Entity types that have different base fields enabled/disabled I decided to just do a 1-to-1 migration from the ECK Entity type to a hard-coded Entity type. The full plan then is to create 4 different Entity types to match the existing Collections, and then re-create the ECK bundles as they are now within those new Entity types.

joegl: To add, I'm fine with creating all these bundles under 1 Entity Type. I'm just curious what your approach would be to that.

joegl: Also follow up just for organization:

4 ECK Entity types exist but there's only 5 bundles. 

Importer Collections
- Bundles: News RSS
- Base fields: None.

Course Collections
- Bundles: Instructor
- Base fields: Label (Title), Author, Created, Changed

Event Collections
- Bundles: Conference Session, Speaker
- Base fields: Label (Title)

Publications Collections
- Bundles: Publication Author
- Base fields: Label (Title)

One issue I encountered when trying to do these as a single Entity Type was that when I added the "label" field to the News RSS bundle (that doesn't use it), I had trouble saving them because the label was required.

pookmish: All valid points. 
I don't see much concern over the Author, Created, & Changed labels. Those just get auto populated and just hang out in the DB.
For the Label property, it's only missing on the news, so we can use some mechanism to generate or just populate that label on those. For example, we use [Automatic Entity Label](https://www.drupal.org/project/auto_entitylabel) on people nodes. I wonder if that could be easily adopted to this custom entity type. If so, we could just populate it with a field value or even the current time or something. And it wouldn't change your update hook much either, just adding another property on them.

I think this way, it will reduce how much custom code we have to maintain. In the event something is deprecated during Druapl 10, we would have less to manage by going with a single entity type.

joegl: Yea using one Entity Type sounds ideal. I do think the Automatic Entity Label solution should work for the News label field. And like you said, the other base fields are mostly in the background. I'll pivot on this and get those change in soon.

joegl: Alright this should be done now.

pookmish (approved):

PR #1331, opened 2023-08-15
feat(shs-4993): adjust link color on warbler color scheme

mndonx:
# READY FOR REVIEW

## Summary
Switches the link colors in Warbler color scheme of Traditional theme from tertiary to secondary. 

## Urgency
_medium_

## Steps to Test
1. View site in traditional theme
2. Switch color scheme to "Warbler"
3. Note that text links are using #005b6c and hover states are using #616a5d


mndonx:
@callinmullaney Seems like the tugboat failed, but if I recall, changes are checked locally so I'll go ahead and tag you. I didn't think it was wise to swap the secondary and tertiary colors, so I made a link exception for the Warbler color scheme. Let me know if that solution makes sense to you.

callinmullaney (changes requested):
@mndonx This one needs some changes. I'll help explain some of the theme complexity and how to fix:
1. Traditional, Colorful, and Airy are the only "Theme" the platform has. Within each theme (via the appearance->theme_settings in Drupal) there are color schemes that can be chosen.
2. The mixin `hb-themes` allows you to target styles for a specific set of themes. Otherwise, if you only need to target a specific theme there is `@include hb-traditional` `@include hb-colorful` and `@include hb-airy`
3. Targeting a specific theme color scheme to override isn't very common but check out the following file to see how this was done previously. The style structure should roughly be the same. `https://github.com/SU-HSDO/suhumsci/blob/develop/docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.icons.scss#L63`

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.links.scss
```diff
@@ -125,6 +125,15 @@
     }
   }
 
+  @include hb-themes(('traditional', 'warbler')) {
```
callinmullaney: Use `@include hb-traditional` given this is the only "Theme" we want to target

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.links.scss
```diff
@@ -125,6 +125,15 @@
...
+    @include hb-pairing-color('color', 'secondary');
+
+    &:hover,
+    &:focus {
+      @include hb-pairing-color('color', 'tertiary-darken-20');
```
callinmullaney: I believe this line should be `@include hb-pairing-color('color', 'tertiary');` according to the ticket AC. Double check this though.

mndonx:
```suggestion
      @include hb-pairing-color('color', 'tertiary');
```

mndonx:
@callinmullaney Ahhh, I was thinking "airy" was a color scheme. I see that it is just listing two different themes. Got it.

callinmullaney (approved):
@mndonx Looks good to me! Thanks for making those updates. The styles now seem to be isolated to the specific color scheme for Traditional. Merging into our release.

PR #1305, opened 2023-05-04
Release: Editor Experience Improvements

callinmullaney:
# READY FOR REVIEW

## Summary
- coming soon

## Need Review By (Date)
on hold

## Urgency
low

## Steps to Test
1. TBD

## PR Checklist
- [PR Checklist](https://gist.github.com/sherakama/0ba17601381e3adbe0cad566ad4d80a5)
- [Humsci Basic PR Checklist](https://github.com/SU-HSDO/suhumsci/blob/develop/docs/HumsciBasicPRChecklist.md)

PR #1335, opened 2023-08-18
SHS-5003: Sort paragraphs in Paragraphs Browser by weight

hectorlj:
# READY FOR REVIEW

## Summary
Add patch that sorts configured paragraph types by weight on the configuration form
Merge only after #1305 is deployed

## Need Review By (Date)
_['10/30', 'asap', etc.]_

## Urgency

## Steps to Test
2. Add a paragraph to an entityreference revision field (Full Width Region or Main Region) where paragraph options were configured by exclusion,
2. Check the paragraphs browser, I see the paragraph options sorted by weight

## PR Checklist


callinmullaney (changes requested):
@hectorlj This meets ticket AC but I think there is another change that is needed here that is somewhat implied. 

Can you update the sorting for the Collection paragraphs field (`/admin/structure/paragraphs_type/hs_collection/fields/paragraph.hs_collection.field_hs_collection_items`) so it matches the Flexible Page order(`/admin/structure/types/manage/hs_basic_page/fields/node.hs_basic_page.field_hs_page_components`)? LUCKILY, this type of change can be exported into yml and imported in each site because `config_ignore` doesn't apply. I figure if we don't make this change now it's going to come back to you in the near future.

callinmullaney (approved):
@hectorlj Thanks for making those weight changes on collections. Looks good to merge :tada:

PR #1340, opened 2023-08-25
fix(shs-5007): fixing redundant links on images

mndonx:
# READY FOR REVIEW

## Summary
Event, News, People, and Publication cards (aka hb-cards) can have multiple elements that all link to the same place -- in particular, the image, the title, and the optional "more" arrow links are all the same. This causes an accessibility violation of ‘redundant links’. Let's use `tabindex` and `aria` attributes on the `<a href>`  element to remove the image and arrow links from the accessibility tree.

```
<a href=https://stanford.edu/ tabindex="-1" aria-hidden="true">
   <img src="image.jpg" alt="Stanford University">
</a>

[... snip ...]

<a href=https://stanford.edu/>
   Stanford University
</a>
```

## Need Review By (Date)
8/26

## Urgency
medium

## Steps to Test
1. All news, events, people and publications with images should have the images wrapped in a link tag that has `tabindex="-1" aria-hidden="true"` attributes
2. "Read More" links now have title text to elaborate on where the link is taking a user ("Read more" doesn't tell the person using keyboard navigation where they are going.)

## PR Checklist


mndonx:
@callinmullaney This is a first pass at this issue. My guess is that any configuration changes I made here will need to be done on ALL the themes, is that correct? I basically removed all the links from all the various card views (so they could be reapplied using the custom attributes.) Let me know what you think about that solution and if I need to go make and make those config changes elsewhere.

cienvaras:
@callinmullaney Some comments for when you review the PR:

- Changing views config will probably won't work, because we can only control the default views, not the overriden ones. If the image field already includes the link we might need to do some parsing and remove or alter the existing link.
- The templates that @mndonx updated are patterns and are used in other parts of the site, not only in views. I actually had an error when I visited `/training-page` in the traditional site, because the component was used in a context where the `nid` was not available.

Please let me know what you think. I can explore other possible solutions on Friday.

mndonx:
@cienvaras @callinmullaney Yes, this was my fear :sweat_smile: -- I'm still learning what touches what. What do you all think about using javascript to alter things as needed? That might be less disruptive. (Normally I wouldn't suggest doing it this way, but I'm not sure there is a better solution.)

cienvaras:
@mndonx  I think I have a solution using a views pre render hook. It basically do the same you're doing in the config, disable the link from the image fields, but on the fly when the view is displayed. I'll do a couple of extra tests and push it to this branch, so we can test it along your changes.

mndonx:
Oh, that's great -- thank you @cienvaras !

docroot/modules/humsci/hs_layouts/patterns/date-stacked-horizontal-card/date-stacked-horizontal-card.html.twig
```diff
@@ -98,14 +98,14 @@
...
-      {% if title %}
-        {% set link_text = 'Read More about '|t ~ title['#row']._entity.title.value %}
-      {% else %}
-        {% set link_text = 'Read More'|t %}
-      {% endif %}
       <div class="hb-card__decorative-link">
-        <a href="{{ url }}" class="hb-card__link">{{ link_text }}</a>
+        {% if url %}
+          {# Set detailed button text that describes where the link is going for accessibility. #}
+          {% set link_text = 'Read More about '|t ~ title['#row'].node_field_data_title %}
```
mndonx, line 104: I believe you may need to check for `title` before setting the `link_text` to use that variable (and that would apply to the other instances here.)

cienvaras:
Closing this PR, need to use a different approach for solving this issue.

PR #1345, opened 2023-09-01
SHS-5011: Testimonial component should use blockquote/cite markup

cienvaras:
# READY FOR REVIEW

## Summary
- Use `<blockquote>` instead of `<div>` in testimonial component

## Need Review By (Date)
- 09/01

## Urgency
- medium

## Steps to Test
- Create a Flexible page and add some testimonials (or visit `/testimonials` if you're in traditional or colorful sites)
- Confirm that there are not visual changes in the components
- Inspect the source code and confirm that the testimonial quote is in a `<blockquote>` tag.

## PR Checklist


callinmullaney (changes requested):
@cienvaras Functional testing is passing as stated and this appears to be the correct place to make this kind of change. However, I looked into this a bit and I actually think there might be additional changes needed here to make this more semantic and accessible. Feel free to push back on this given it's technically out of scope for the ticket's AC
Using https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element for reference:
1. The exclusion of the `cite` attribute in this PR seems appropriate for the component. These are intended as quotes from people and are not quotes from an external source. 
2. Given these are people-centric quotes it might actually be better to update more markup in this component so `<figure>` wraps the `<blockquote>` and the attribution markup (`.hb-testimonial__author-credits`) is converted to a `<figcaption>`

cienvaras:
@callinmullaney Thanks for the feedback! You're right the `figure` and `figcaption` are a more semantic way to wrap the content. I added the tags and updated the styles. There's a minor difference between the "Top" and "Botton" variables, because in the "Botton" testimonial the link is not inside the author block, but I don't thing that's a problem.

callinmullaney (approved):
@cienvaras Looks good to me!

PR #1354, opened 2023-09-14
fix(shs-5060): fix H2s around titles in bookshelf/publication cards

mndonx:
# READY FOR REVIEW

## Summary
This PR adds a check in the vertical card template and the horizontal card template for if a heading tag has been set in the view, or if it needs to be manually added in the template.  

## Steps to Test
1. Visit publications default view in a local instance, such as: https://hs-traditional.suhumsci.loc/default-views/publications
2. Note that publications no longer has the duplicated heading tags (`<h2></h2><h3>Title</h3>`). 
3. Verify that other vertical or horizontal cards are not affected (they EITHER have their custom view's heading tag or the `<h2>` tag coming from the template. 

## PR Checklist


mndonx:
@cienvaras This is ready for review. I only added it to vertical-card and horizontal-card -- would there be a case to do it on the vertical link card as well? Any others? 

Thanks!!

cienvaras:
@mndonx Works great! There are just a couple of things that can be improved:

- Can you please move the preprocess hooks to `docroot/modules/humsci/hs_layouts/hs_layouts.module` please? It's better to keep all the patterns related code together in the same place.
- Let's move the heading detection code to a new function `_hs_layouts_field_has_title` that takes the field as an argument and returns a boolean, and call this function from the preprocess hooks. This way we can reuse the same code for all the hooks.

> I only added it to vertical-card and horizontal-card -- would there be a case to do it on the vertical link card as well? Any others?
That's good question. Those two patterns are the ones being used in the examples provided, but there might be other patterns being used in other views.  Lets add `vertical-link-card` for now, I think these are the ones that make sense for this content type.

mndonx:
@cienvaras I've moved the code, and tried to make a function that is reusable. Please let me know if I made it more complicated than necessary! I also include the vertical-link-card. 

Thank you!!

cienvaras:
@mndonx Thanks, works great! I just did a minor adjustment to avoid having to pass the `$variables` array to the function. Let me know if you have any questions on the change.

cienvaras (approved):

mndonx:
@cienvaras Thank you! I wasn't quite sure about that.

PR #1357, opened 2023-09-20
SHS-5061: Upgrade text format on postcard text area

hectorlj:
# READY FOR REVIEW

## Summary
Replaces Postcard component Body field format `Minimal HTML` with `Minimal HTML with Styles`  

## Steps to Test
1. Check that `Minimal HTML with Styles` format exists and has correct permissions `/admin/config/content/formats`
2. Ensure configuration for `Minimal HTML with Styles` matches `Minimal HTML`, but with the `Styles` button added to the toolbar `/admin/config/content/formats/manage/minimal_html_with_styles`
3. Check that ` Minimal HTML with Styles` option is the only one checked in `/admin/structure/paragraphs_type/hs_postcard/fields/paragraph.hs_postcard.field_hs_postcard_body`
4. Create content (e.g., flexible page) and embedded the Postcard component
5. The wysiwyg editor in the `Body` field should have the ability to use `Styles`

## Notes
- _Do not merge until design work has been completed_

## PR Checklist


cienvaras (changes requested):
@hectorlj Looks good, but there's a problem with the update hooks. When I ran a db sync, the hooks were executed with no errors, but the changes didn't get applied. I think that the problem is that the hooks are being executed before config is imported, so the text format and the permission don't exist yet. You need to load and apply the new config items in the hook before applying the updates.

cienvaras (changes requested):
@hectorlj I checked again and the permissions and allowed formats are correctly updated. However, I think there's a missing step. The format for existing postcard paragraphs should be updated. Right now, if I edit an existing postcard with content in the Body field, I get this:

Screenshot 2023-10-10 at 10 49 00

It's necessary to set the right format, to prevent confusing the end users.

cienvaras:
@hectorlj I works as expected, however there are errors when the update functions are applied along the Drupal 10 updates. Probably something in the order, as the Drupal 10 hooks alter some user related stuff. Also, D10 has support in core for defining the allowed formats for a field, and the allowed formats module has an update hook for dealing with this.

We can look into possible fixes for this next week.

hectorlj:
@cienvaras Fixed the permissions update hook. The D10 allowed formats are updated in a post update, which will run after `hs_paragraph_types_update_9402`, so that update hook should be fine

cienvaras (approved):
@hectorlj Works great, thanks!

PR #1360, opened 2023-09-26
fix(shs-5063): adjustments to link/icon color in dark mode

mndonx:
# READY FOR REVIEW

## Summary
- Updates PDF, Lock, and File icons in local footer in Colorful theme in dark mode in all color schemes to white / gray medium on hover
- Updates links on Vertical Linked Cards to be white with gray medium hover for Colorful theme
- Updates links on Vertical Linked Card to be black with gray dark hover for Traditional theme

## Urgency
medium

## Steps to Test
1. View the site in Colorful theme. 
2. Ensure the local footer is set to dark mode
3. In all color schemes, make sure the lock, pdf and file icons show in the local footer as white with gray hover. 
4. Visit a page with Vertical Links Cards and add a link to one, like here: http://hs-colorful.suhumsci.loc/components/collections-cards
5. Ensure that links are white with gray hover*
6. Switch to view site in Traditional theme. 
7. Visit a page with Vertical Links Cards and add a link to one, like here: http://hs-traditional.suhumsci.loc/components/collections-cards
8. Ensure that links are black with dark gray hover*

* There is a major issue where the links are just not clickable in this kind of card. I'm not sure it should be an option here. There is no way to click since the whole card is covered in an absolutely positioned link.


mndonx:
@cienvaras Please see my note. I'm not sure about styling links that are not clickable -- it seems like that's a bigger issue. Also, should those vertical linked cards be able to have lock/file/pdf icons added to links? I was not able to figure out how to do that.

cienvaras (changes requested):
@mndonx It seems that there are some cases where colors are not being applied correctly:

- Colors of the footer icons don't match the text color in Traditional, for example, in warbler:

  traditional-warbler

- Postcard link text color is being overridden in Traditional/Warbler:

  card-traditional-warbler

- Lock color on hover is being overridden based on the color scheme for most cases. For example, in Colorful/Ocean:

  colorful-ocean

- Postcard links hover color is not being applied in Colorful
- Postcard links hover color in Traditional doesn't match the color in the ticket. It's `#343639` instead of `#413e39`. Probably an error in the ticket, because this is the value for Gray Medium in Traditional, but it's probably good to check with Tori.

There might be other issues for other Theme / Color scheme combinations, but the examples outline the most common type of errors.

Let me know if you have any questions!

cienvaras:
@mndonx 

> I'm not sure about styling links that are not clickable -- it seems like that's a bigger issue. 

I'm pretty sure we have another ticket for this, I'll check ClickUp.

> Also, should those vertical linked cards be able to have lock/file/pdf icons added to links? I was not able to figure out how to do that.

Yes, but that will probably be other ticket, because we need to do some BE for this before.

mndonx:
@cienvaras I think this is ready for review again. Just two things:

- You mentioned the gray being wrong. Can you point me to where this is happening? I set it to a global gray-medium variable. 

- The hover on the link within the card text isn't going to happen (because of the bug with the link overlayed on top.) I'll be sure to mention that bug again here because it seems weird to try to style it when it shouldn't currently be used at all.

cienvaras (approved):
@mndonx Looks good! It seems that all the issues are solved now. I'm approving and releasing this PR, however there are a couple of things that might need to be adjusted.

I found some examples of inconsistent link colors in the colorful theme. For example:

  Ocean colorscheme:  
  Screenshot 2023-09-28 at 18 58 16

  Mountain colorscheme:
  Screenshot 2023-09-28 at 19 08 58

As you can see, there is a combination of multiple link colors, but the ones with icons have the right styles. However, the ticket is not completely clear about if all links should have the same color or if only the links with icons. It would be great if ask Tori for clarification on this, but I still want to release the current changes to stage so they can test the different combinations. We have enough time to do fixes before the code gets to production.

> You mentioned the gray being wrong. Can you point me to where this is happening? I set it to a global gray-medium variable.

Yes, in the ticket it mentions that, for Traditional, the value should be "Gray Dark" or `#413E39`, but that hex value is the value for "Gray Dark" in Colorful, not Traditional. I'm pretty sure it's an error in the ticket, but we can ask Tori to be on the safe side.

> The hover on the link within the card text isn't going to happen (because of the bug with the link overlayed on top.) I'll be sure to mention that bug again here because it seems weird to try to style it when it shouldn't currently be used at all.

Yes, that's right. We have a ticket for fixing the overlay links so we'll eventually need to add these hover styles. We should create a follow up ticket to do this and also set the right colors for the icons, once we have a way to insert icons in the postcard's text area.

mndonx:
@cienvaras 

> As you can see, there is a combination of multiple link colors, but the ones with icons have the right styles. 

Yep - I went back and forth on this as well, but ultimately it seemed like they were just asking for the icon links to be changed ... BUT they may change their mind when they see it in action. I really don't like all the tedious variations, it becomes really hard to manage! If it does end up getting streamlined, perhaps I can make it a little more generalized for all footer link colors. I'm happy to ask Tori about any and all of these straggling questions tomorrow!

PR #1369, opened 2023-10-05
fix(shs-5067): applying js to colorbox to fix a11y errors

mndonx:
# READY FOR REVIEW

## Summary
This PR adds text to hidden buttons on the colorbox gallery to fix some accessibility issues. 

## Urgency
low

## Steps to Test
1. Visit a photo album component, like https://hs-colorful.suhumsci.loc/components/photo-album
2. Note that WAVE no longer produces 3 errors related to Empty Buttons. 

Screenshot from 2023-10-05 14-40-29

## PR Checklist


docroot/themes/humsci/humsci_basic/src/js/shared/photo-album/colorbox.js
```diff
@@ -0,0 +1,22 @@
+const attachStanfordColorbox = (context) => {
+  const colorboxLinks = 'a.colorbox';
+  const galleryDialogSpan = document.createElement('span');
+  galleryDialogSpan.className = 'sr-only';
+  galleryDialogSpan.textContent = 'Opens gallery dialog';
+  document.querySelector(colorboxLinks).appendChild(galleryDialogSpan);
```
cienvaras: This produces an error if there are no colorbox links in the page, it's necessary to add an `if` to call `appendChild()` only if the result of `querySelector()` is not null or move the code inside the `if` in line 9.

Also, I have some doubts about the intended behavior. `querySelector` only returns the first element in the gallery. Is this the intended behavior?

There's another issue: When I checked the gallery, I noticed that there's an `aria-label` in the `a.colorbox` elements that takes precedence over any inside content in screen readers. It's also malformed, so it might be better to remove it here and use the `span` you're adding.

cienvaras (changes requested):

docroot/themes/humsci/humsci_basic/src/js/shared/photo-album/colorbox.js
cienvaras: btw, I know this is still WIP, just wanted to go ahead and start checking :)

mndonx: Thanks @cienvaras ! I'm struggling to get the locals on both of my computers working again after the D10 update, so I'm afraid I'm delayed getting this back into review. I appreciate the review, and once I'm back in business, I can take a look at these issues.

mndonx: @cienvaras OK, back in business. I've removed the stuff that was borrowing from SWS because I actually don't believe it is needed here to meet the acceptance criteria (and, as you found, is not working as expected.) Now it is *just* adding text in the prev/next/slideshow hidden buttons. Let me know if that is now working as expected!

mndonx: Also, what base branch should I compare this to? I merged in develop but the diff is weird ... it's possible I might need to make a fresh branch, since this really is just changes to like 2 files.

cienvaras: @mndonx Sorry, I forgot to reply sooner. You should always use the most recent release branch (for this sprint `10.0.0-release`). That's the branch where we merge our work and that gets deployed to stage. Then, after testing they merge that into `develop`. In theory we could start our work using `develop` as base, because the release branch will always contain everything in `develop`, but sometimes they add hotfixes into `develop` that are not included in the release branch inmediately.

cienvaras (approved):
Works great, no Wave errors.

PR #1371, opened 2023-10-10
SHS-5059: Skip to main content

cienvaras:
# READY FOR REVIEW

## Summary
- Update `page.html.twig` templates to include the `main-content` id only when layout builder is not enabled
- Add field a logic to include the `main-content` id on the layout builder sections. The id can be added on the main region, around the whole section to include the left sidebar or disabled
- Add validation to the layout customizing forms (both default and overridden) to ensure that the `main-content` id is present in exactly one section
- Create update hooks to update the sections main content field for both default and overridden layouts. For all layouts, the first non full width layout is being configured to display the `main-content` id in the main region. In all other sections in the layout, the main content field is set to `none`.
- Use `<main>` tags for the main content container.

## Need Review By (Date)
11/13

## Urgency
medium

## Steps to Test
1. After the site is installed, visit the default layout settings (Manage display > Manage Layout) for all content types and confirm that the "Main Content begins at" field is present in all section variants. Confirm that, for all content types, the first non full width section has that field set to "Main region" and all others sections has it set to "None".
2. Find a flexible page or private page with an overridden layout. Confirm that the new field is present and that the configurations is the same as the default layouts (first non full width section set to "Main region", all others to "None"). Some examples of nodes with overridden layout in traditional:
  a. https://hs-traditional.suhumsci.loc/mixed-page-well-raised-cards-full-width
  b. https://hs-traditional.suhumsci.loc/mixed-pages/mixed-page-wells-raised-cards-full-width-sidebar
  c. https://hs-traditional.suhumsci.loc/components/collections-cards
  d. https://hs-traditional.suhumsci.loc/private-pages/private-page-collection-cards
3. Edit a default layout and try the following:
  a. Set the main content field in all sections to "None"
  b. Set more than one section with a value different to "None"

    In both cases, you should get a validation error that prevents you to save the form until exactly one section has a non "None" value.

4. Repeat the previous tests on a node with an overridden layouts. You should get the same validation.
5. Test some random pages on the site. All of them should have exactly one `main-content` id on its source, in the position configured on the layout settings for that content type or node. The container with the `main-content` id should use the `<main>` tag.

## PR Checklist


config/default/core.entity_view_display.node.hs_basic_page.default.yml
```diff
@@ -52,6 +53,7 @@ third_party_settings:
           context_mapping: {  }
           section_width: decanter-grid
           region_widths: center
+          main_content: main-region
```
pookmish: it would be much easier to make this a true/false value on the layout region instead of checking for strings. If that's possible.

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -113,3 +114,41 @@ function hs_layouts_preprocess_pattern_vertical_link_card(&$variables) {
...
+ * Implements hook_form_FORM_ID_alter().
+ */
+function hs_layouts_form_entity_view_display_layout_builder_form_alter(&$form) {
+  $form['#validate'][] = '_hs_layouts_unique_main_content_section_validate';
+}
+
+/**
+ * Exactly one section should be set as 'Main Content' in Layout Builder form.
+ */
+function _hs_layouts_unique_main_content_section_validate(&$form, FormStateInterface $form_state) {
```
pookmish, line 89: Is there any way this be done in the validation plugin? It's basically duplicated by the plugin

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -113,3 +114,41 @@ function hs_layouts_preprocess_pattern_vertical_link_card(&$variables) {
...
+  $main_content_found = FALSE;
+  foreach ($form['layout_builder']['#section_storage']->getSections() as $section) {
+    if ($section->getLayoutSettings()['main_content'] !== 'none') {
+      if ($main_content_found) {
+        $form_state->setErrorByName('form', 'Exactly one region should contain the main content id.');
```
pookmish: Translate strings

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -113,3 +114,41 @@ function hs_layouts_preprocess_pattern_vertical_link_card(&$variables) {
...
+        break;
+      }
+      else {
+        $main_content_found = TRUE;
+      }
+    }
+  }
+  if (!$main_content_found) {
+    $form_state->setErrorByName('form', 'Exactly one region should contain the main content id.');
```
pookmish: Translate strings

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -113,3 +114,41 @@ function hs_layouts_preprocess_pattern_vertical_link_card(&$variables) {
...
+ */
+function _hs_layouts_unique_main_content_section_validate(&$form, FormStateInterface $form_state) {
+  $main_content_found = FALSE;
+  foreach ($form['layout_builder']['#section_storage']->getSections() as $section) {
+    if ($section->getLayoutSettings()['main_content'] !== 'none') {
+      if ($main_content_found) {
+        $form_state->setErrorByName('form', 'Exactly one region should contain the main content id.');
+        break;
+      }
+      else {
```
pookmish: No need for an else block here if you return after setting the error.

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($node, Constraint $constraint) {
+    // Layout builder should be enabled and the layout should be overridable for
+    // the current node, otherwise return.
+    $view_display = $this->displayRepository->getViewDisplay('node', $node->bundle());
+    if (!($view_display->isLayoutBuilderEnabled() && $view_display->isOverridable()))
+      return NULL;
```
pookmish: wrap if blocks in curly braces

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
```diff
@@ -0,0 +1,109 @@
...
+
+    // Get node's section storage.
+    $node_context = EntityContext::fromEntity($node);
+    $section_storage = $this->storageManager->load('overrides', [
+      'entity' => $node_context,
+      'view_mode' => new Context(new ContextDefinition('string'), 'default'),
+    ]);
+    // Continue only if a temp section storage exists (layout is being editted).
+    if (!($section_storage && $this->tempstoreRepository->has($section_storage)))
+      return NULL;
```
pookmish: wrap if blocks in curly braces

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -129,6 +129,15 @@ function humsci_basic_preprocess_page(&$vars) {
...
   $vars['site_name'] = $config->get('name');
   $vars['site_slogan'] = $config->get('slogan');
+  
+  if ($route === 'entity.node.canonical') {
+    $node = $vars['node'];
+    // Determine if Layout Builder is enabled for current node. Based on this,
+    // the default main content id will be added/removed.
+    $vars['lb_enabled'] = \Drupal::service('entity_display.repository')
+      ->getViewDisplay('node', $node->bundle())
+      ->isLayoutBuilderEnabled();
```
pookmish: This is pretty heavy. You can just check if `$node->get('layout_builder__layout')->count()` to check for customized layout builder settings. Otherwise it should be expected every content type has layout builder on the default display.

docroot/modules/humsci/hs_layouts/hs_layouts.install
```diff
@@ -0,0 +1,116 @@
...
+function hs_layouts_update_9401() {
+  $storage = \Drupal::entityTypeManager()->getStorage('entity_view_display');
+  $query = $storage->getQuery()
+    ->condition('targetEntityType', 'node')
+    ->condition('mode', 'default');
+  $items = $query->execute();
+  foreach ($items as $item) {
+    $view_display = $storage->load($item);
+    if ($view_display->isLayoutBuilderEnabled()) {
+      $sections = $view_display->getThirdPartySetting('layout_builder', 'sections');
```
pookmish: `$view-display->getSections()` would be more concise here.

docroot/modules/humsci/hs_layouts/hs_layouts.install
```diff
@@ -0,0 +1,116 @@
...
+      if (count($sections)) {
+        $main_section = $sections[0];
+        if (count($sections) > 1) {
```
pookmish: consolidate these two conditions

docroot/modules/humsci/hs_layouts/hs_layouts.install
```diff
@@ -0,0 +1,116 @@
...
+        $view_display->save();
+      }
+    }
+  }
+}
+
+/**
+ * Set main content id in nodes with overridden layout builder config.
+ */
+function hs_layouts_update_9402() {
```
pookmish: This needs to be done in a batch process.

docroot/modules/humsci/hs_layouts/hs_layouts.install
```diff
@@ -0,0 +1,116 @@
...
+  // Get all sections in nodes with overridden layout.
+  $db = \Drupal::database();
+  $result = $db->select('node__layout_builder__layout', 'nl')
+    ->fields('nl', ['entity_id', 'delta', 'layout_builder__layout_section'])
+    ->orderBy('entity_id')
+    ->orderBy('delta')
+    ->execute();
```
pookmish: use entity query and refactor the upgrade process. Do the work on the node and sections, not the database record. This is far too dangerous and complicated. Do basically the same as the previous update hook.

pookmish (changes requested):
Please review and refactor

config/default/core.entity_view_display.node.hs_basic_page.default.yml
cienvaras: There are some reasons for using a single string value instead of multiple booleans:
- Code is simpler with a single value, even if it's a string. Checking / assigning multiple booleans would make the conditionals more complex
- There all three possible values for all sections, regardless of available regions. The values for this field don't necessarily match a region.
- It's easier for admin users to set a single value per section, instead of setting multiple values on each region.

docroot/modules/humsci/hs_layouts/hs_layouts.module
cienvaras: `entity_view_display` view display entities don't have contraints implemented yet. It's [planned for Drupal 11](https://www.drupal.org/project/drupal/issues/2869792) but not available yet. For now, the only possible way we found for validating this was to do it on the form.

docroot/modules/humsci/hs_layouts/hs_layouts.module
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/hs_layouts.module
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/hs_layouts.module
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
cienvaras: Fixed.

docroot/themes/humsci/humsci_basic/humsci_basic.theme
cienvaras: Fixed. Removed the check altogether because I was checking for layout builder enabled for the display, not overrides. Added a check on the templates to check for node pages.

docroot/modules/humsci/hs_layouts/hs_layouts.install
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/hs_layouts.install
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/hs_layouts.install
cienvaras: Fixed.

docroot/modules/humsci/hs_layouts/hs_layouts.install
cienvaras: Fixed. However, there's a problem when running the update hooks in some specific sites. We're getting the following errors:

> [error]  Drupal\purge\Plugin\Purge\Invalidation\Exception\InvalidExpressionException: The URL is invalid. in Drupal\purge\Plugin\Purge\Invalidation\UrlInvalidation->validateExpression() (line 84 of /app/docroot/modules/contrib/purge/src/Plugin/Purge/Invalidation/UrlInvalidation.php). 
> [error]  Drupal\Core\Entity\EntityStorageException: The URL is invalid. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 

I did some debugging and the problem happens only in sites with an underscore on its machine name (`hs_traditional`, `hs_colorful`, etc.) It seems that the purge module has a presave hook that validates the URLs before purging. But the URL used by the drush install commands is the machine name (e.g. `http://hs_traditional/`) and the underscores are not valid characters in URL's.

I'm pretty sure the error also happens in Acquia. Do you know if there's a workaround for this? Updating the database directly doesn't trigger the presave hook, but I agree that it's not the best way to implement this update.

cienvaras:
@pookmish Thanks a lot for the feedback. It's ready for review again. I also left a question about an issue we're having with one of the update hooks, any advice for handling this is welcomed.

config/default/core.entity_view_display.node.hs_basic_page.default.yml
pookmish: ok, i'm fine with that, but lets make the value match the region name. that makes it more dynamic for any possible layouts in the future.

docroot/modules/humsci/hs_layouts/src/Plugin/Layout/HumsciBaseLayout.php
```diff
@@ -0,0 +1,64 @@
...
+
+    $form['main_content'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Main Content begins at'),
+      '#description' => $this->t('Where does the page’s main content begin? Typically this is the Main Region but occasionally is the Left Sidebar. Choose “None” if you’ve already set this value on another section.'),
+      '#default_value' => $this->configuration['main_content'] ?: 'none',
+      '#options' => [
+        'none' => $this->t('None'),
+        'main-region' => $this->t('Main Region'),
+        'left-sidebar' => $this->t('Left Sidebar'),
```
pookmish: get the options from the layout definition and add the `none` option. that will make it less hard coded.

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($node, Constraint $constraint) {
+    // Layout builder should be enabled and the layout should be overridable for
+    // the current node, otherwise return.
+    $view_display = $this->displayRepository->getViewDisplay('node', $node->bundle());
+    if (!($view_display->isLayoutBuilderEnabled() && $view_display->isOverridable())) {
```
pookmish: check if these methods even exist by making sure the object is an instance of `\Drupal\layout_builder\Entity\LayoutEntityDisplayInterface`

pookmish (changes requested):

docroot/modules/humsci/hs_layouts/src/Plugin/Validation/Constraint/UniqueMainContentConstraintValidator.php
cienvaras: Thanks, fixed!

docroot/modules/humsci/hs_layouts/src/Plugin/Layout/HumsciBaseLayout.php
cienvaras: As mentioned in a previous comment, the values for the `main_content` setting don't match exactly the available regions for a layout. The options are the same for all layouts, in order to make things easier for end users. I renamed the values to match the region names in the "3 Column" layout as requested.

cienvaras:
@pookmish Thanks for the feedback. This is ready for review again.

docroot/modules/humsci/hs_layouts/src/Plugin/Layout/HumsciBaseLayout.php
pookmish: Then rename the values in the layout definition and get the options from the plugin definition. This is my requirement.

pookmish:
No, it still requires changes I mentioned in previous reviews.

ahughes3:
@cienvaras tagging you for additional visibility in regards to the latest comment from @pookmish

docroot/modules/humsci/hs_layouts/src/Plugin/Layout/HumsciBaseLayout.php
cienvaras: @pookmish  I think I didn’t explained myself well in the previous comments. The values for this field are not related to the sections in the layout definition. We’ll always have the same values, for all different layouts:
- None
- Main Section
- Left Sidebar

If we get the values from the sections in the layout definition, we’ll have things that don’t make sense. For example, the “3 Column” layout has a “Right Sidebar” that we don’t need and the “3 Column With Image” has 7 different regions that don’t match at all the values we need for the `main_content` field.

Even if the values resemble specific sections in some layouts, there’s not a 1-1 relation in any of the layouts. This field is similar to `section_width` or `region_widths`. with a hardcoded set of values that will be the same in all cases.

cienvaras:
@pookmish I just checked all the feedback and I'm pretty sure all the required fixes have been applied. Please let me know if there's anything that I'm not missing or any other improvement that can be done. Thanks! (cc. @ahughes3)

ahughes3:
@pookmish @joegl adding an additional nudge here to see the latest comment from @cienvaras 

Please let us know what needs to be done in order to get this one merged and rolled out to stage/production

docroot/modules/humsci/hs_layouts/src/Plugin/Layout/HumsciBaseLayout.php
pookmish: This doesn't provide a future proof approach. I understand the situation you are avoiding, but it will be better to make this a dynamic option.

docroot/modules/humsci/hs_layouts/hs_layouts.install
joegl: @cienvaras Is there a specific site you're experiencing this on? I'd like to test it on my local. I'm also not sure if this would actually be an issue on Acquia.

cienvaras: @joegl Yes, in sites that have an underscore in their name: `hs_traditional`, `hs_colorful`, `culture_emotion_lab`, etc.

I did some tests in the acquia dev environment last year and the problem was present, because drush used the `http://[SITENAME]` as the domain in the CI. But I'm not sure if the problem is still present.

joegl: Oh right, those would be good examples to try based on your first report 🤦

PR #1377, opened 2023-10-18
fix(shs-5105): gin form override for input width overlap

mndonx:
# READY FOR REVIEW

## Summary
This PR adds a small override to the Gin theme (coming from the humsci_basic theme) to constrain form text input fields to 100%. 

## Urgency
medium

## Steps to Test
1. Edit the homepage in both hs_colorful and hs_traditional
2. Edit either the "Testimonial" or "Spotlight" component 
3. Bring your browser down to a tablet sized width (before it switches to a single column)
4. Note that the fields are no longer breaking out of their parent container. 

## PR Checklist


mndonx:
@cienvaras Let me know if you think I've put this change in the correct location. This file is in humsci_basic, but is also loaded when in Gin.

cienvaras (changes requested):
@mndonx `imports.scss` contains styles for ckeditor, that's why it's been loaded in Gin. However, it's not the best option for these styles, as they're not ckeditor related.

For adding styles to the admin theme there are two options:

- The custom admin theme (Gin based) that the site is using: `docroot/themes/humsci/su_humsci_gin_admin`
- The `hs_admin` module: `/Users/andres/dev/suhumsci/docroot/modules/humsci/hs_admin`

I'm not completely sure about why there are two places where the admin theme styles can be changed, as this was implemented before I started in the project. In this case, I think that the custom admin theme is the best option for adding these styles.

mndonx:
@cienvaras Aha, I should have guessed there was a custom Gin admin theme, I just couldn't find it. I'll move that bit now. Sorry about that!

mndonx:
@cienvaras Ready for re-review. I believe that this subtheme does not have compiled Sass files, just CSS files in the dist directory.

cienvaras (approved):
Works great, thanks!

PR #1384, opened 2023-10-25
SHS-5093: A11y Level A: Element IDs are not unique, on search results page

hectorlj:
# READY FOR REVIEW

## Summary
- Create unique id for each instance of the default search 

## Steps to Test
Do the following for several sites:
1. Navigate to `/search`
2. Verify search form element ids are unique

## PR Checklist


docroot/modules/humsci/hs_blocks/hs_blocks.module
```diff
@@ -116,3 +116,16 @@ function hs_blocks_preprocess_block__layout_builder(&$variables) {
     $variables['label'] = $variables['configuration']['label'];
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function hs_blocks_preprocess_block__views_exposed_filter_block(&$variables) {
+  if (str_contains($variables['derivative_plugin_id'], "search-search_page")) {
```
cienvaras: Please change the double quotes to single quotes:

```suggestion
  if (str_contains($variables['derivative_plugin_id'], 'search-search_page')) {
```

docroot/modules/humsci/hs_blocks/hs_blocks.module
```diff
@@ -116,3 +116,16 @@ function hs_blocks_preprocess_block__layout_builder(&$variables) {
...
+    // Append the build id to create
+    // unique form id.
+    $build_id = $variables['content']['#build_id'];
+    $variables['content']['#id'] .= '-' . $build_id;
+    $variables['#cache']['max-age'] = 0;
```
cienvaras: Are you sure this is needed? I did a quick test with the cache enabled and I still got different ids for both blocks. I might be doing something wrong in my test, but please check to confirm if it's needed.

cienvaras (changes requested):
Works as expected, but I have a question and also found a minor thing to fix. Please see inline comments.

docroot/modules/humsci/hs_blocks/hs_blocks.module
hectorlj: True, even if they get cached the ids will still be unique. You're right, it's not necessary to update the id every time. I'll remove it 👍🏼

cienvaras (approved):
Works great, thanks!

PR #1390, opened 2023-11-03
Shs 5107 media table overflow

mndonx:
# READY FOR REVIEW

## Summary
This PR fixes the overflow issues on Media admin table views. 
- Removes code that overflowed the box
- Overrides code that was hiding the scrollbar (which is a bad accessibility practice)
- Adds a small gradient to the right of the table to indicate that there is more content to the right if one scrolls.

## Urgency
medium

## Steps to Test
1.  Log in to hs_traditional or hs_colorful locally. 
2. Visit the media listing admin page, like https://hs-colorful.suhumsci.loc/admin/content/media
3. Note that the listing no longer overflows, and the scrollbar is visible at the bottom. There is a small gradient on the right to show there is more content to the right.

## PR Checklist


cienvaras (approved):
Thanks, works as expected.

PR #1397, opened 2023-11-16
SHS 5119 jQuery UI dialog z-index

mndonx:
# READY FOR REVIEW

## Summary
This PR updates the jquery ui variables inside the custom theme. The issue seems to be related to 
https://www.drupal.org/project/gin_lb/issues/3351882

## Urgency
medium

## Steps to Test
1.Visit a Layout Builder Page like https://hs-colorful.suhumsci.loc/admin/structure/types/manage/hs_person/display/default/layout
2. Click on a  Pencil icon next to one of the blocks, and then click "Configure"
3. Ensure that the modal window appears in front of the grayed-out panel (instead of behind it.) 

## PR Checklist


mndonx:
@cienvaras When I came back to my computer after walking away for awhile, I could no longer see a discrepancy between the 10.0.3 release branch and develop, so I think I'll leave it here. However, I find the checkbox/radio input theming to be hard to read here. Is that new? It seems like the generic input theming is making them 100% wide, and too big. Perhaps that's a known issue. 
Screenshot from 2023-11-15 20-13-16

mndonx:
For posterity, I think something like this would fix the issue with the checkboxes: 

```
input[type="checkbox"],
input[type="radio"] {
  line-height: normal;
  min-height: 0;
  width: auto;

  + label {
    display: inline-block;
    width: auto;
    margin-bottom: 0;
    margin-top: hb-calculate-rems(4px);
  }
}
```

cienvaras:
@mndonx Thanks, it works as expected! The checkbox styles are not new, it's been like this at least since I started in the project. We can talk later with Tori about fixing it, but let's leave it untouched for now.

cienvaras (approved):

PR #1404, opened 2023-11-22
Shs 5134 photo album aria

mndonx:
# READY FOR REVIEW

## Summary
This PR makes adjustments to how the colorbox gallery pulls in attributes for identifying link purposes.

## Urgency
medium

## Steps to Test
1. Visit an example page with a Photo Gallery like http://hs-colorful.suhumsci.loc/components/photo-album
2. The thumbnail gallery images should be links. In the markup, the links should have an `aria-title` attribute. This attribute should be (in this order): the alt text of the image if it isn't blank, the caption if it isn't blank, or generic text "Thumbnail Image" if neither of those exist.

I realize this strays from the original AC. But I'm realizing that the problem here is that the aria-title exists because sometimes the images don't have alt text. And that's a problem for describing the to the user what they are clicking on and why. At the same time, there's some sort of bug with the colorbox templates where even if there's no alt text on the image, it pulls in some weird additional markup (`{"alt":""}` or something like that), so there won't ever be an error saying "hey, there's no alt text here." 

I tried to approach this with giving best to worst aria-title text possible to the link. Maybe the generic text could be improved. Please let me know what you think! 

## PR Checklist

### DIFF
--- a/docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
+++ b/docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
@@ -14,13 +14,14 @@
  */
 #}
 
-{# set a generic link label in case there is no alt attribute on the image or a caption #}
+{# set a generic link label in case there is no alt attribute on the image #}
 {% set link_label = "Open image in the gallery" %}
-{# If the image has an alt attribute, use it as the link label. #}
-{% if attributes['data-cbox-img-attrs']|render != '{&quot;alt&quot;:&quot;&quot;}' %}
-  {% set link_label = "Open " ~ attributes['data-cbox-img-attrs'] ~ " image in the gallery" %}
+
+{# If the image has an alt attribute, use it in the link label. #}
+{% set alt_text = image['#alt'] %}
+
+{% if alt_text %}
+  {% set link_label = "Open " ~ alt_text ~ " image in the gallery" %}
 {% endif %}
-{% set link_label = link_label|raw %}
-{% set link_label = link_label|replace({'{&quot;alt&quot;:&quot;':'', '&quot;}':''}) %}
 
 <a href="{{ url }}" aria-controls="colorbox" aria-label="{{ link_label }}" role="button" {{ attributes }}>{{ image }}</a>
--- a/docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
+++ b/docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
@@ -1,0 +1,27 @@
+{#
+/**
+ * @file
+ * Default theme implementation to display a formatted colorbox image field.
+ *
+ * Available variables:
+ * - image: A collection of image data.
+ * - url: An URL the image can be linked to.
+ * - attributes: Link attributes.
+ *
+ * @see template_preprocess_colorbox_formatter()
+ *
+ * @ingroup themeable
+ */
+#}
+
+{# set a generic link label in case there is no alt attribute on the image or a caption #}
+{% set link_label = "Thumbnail Image" %}
+{# If the image has an alt attribute, use it as the link label. #}
+{% if attributes['data-cbox-img-attrs']|render != "{&quot;alt&quot;:&quot;&quot;}" %}
+  {% set link_label = attributes['data-cbox-img-attrs'] %}
+{# If the image has a title attribute, use it as the link label. #}
+{% elseif attributes['title']|render != "" %}
+  {% set link_label = attributes['title'] %}
+{% endif %}
+
+<a href="{{ url }}" aria-controls="colorbox" role="button" aria-title="{{ link_label|raw }}" data-cbox-img-attrs="{{ link_label|raw }}" {{ attributes }}>{{ image }}</a>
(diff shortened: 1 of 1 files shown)


docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
```diff
@@ -0,0 +1,27 @@
...
+{% set link_label = "Thumbnail Image" %}
+{# If the image has an alt attribute, use it as the link label. #}
+{% if attributes['data-cbox-img-attrs']|render != "{&quot;alt&quot;:&quot;&quot;}" %}
+  {% set link_label = attributes['data-cbox-img-attrs'] %}
+{# If the image has a title attribute, use it as the link label. #}
+{% elseif attributes['title']|render != "" %}
+  {% set link_label = attributes['title'] %}
+{% endif %}
+
+<a href="{{ url }}" aria-controls="colorbox" role="button" aria-title="{{ link_label|raw }}" data-cbox-img-attrs="{{ link_label|raw }}" {{ attributes }}>{{ image }}</a>
```
cienvaras: Please change `aria-title` to `aria-label`

cienvaras (changes requested):
I think this approach works, but there are a couple of things that can be improved:

1.  The variable `attributes['data-cbox-img-attrs']` is a JSON string, it contains special chars and extra text that is included in the alt text and read by the screen reader. This can confuse end users:
    Screenshot 2023-11-23 at 14 17 04

    It should be possible to parse this string and extract the alt text itself.

2. I think that the labels can be more descriptive. The purpose of the links is not clear, there's not descriptive text anywhere indicating that this is a gallery. While this is also a content issue, we could create links that provide a better description of the purpose. For example instead of adding just the alt text in the `aria-label`,  we could set the text to be "Open [ALT] image in the gallery" (or something similar). We could just use "Open image in the gallery" for images without an alt text.

Also, it would be good to check with Tori if the goal is to have users to set correct alt text for all images. Editoria11y reports the missing alt texts, but only if the `aria-label` is not present. Maybe it would be better to leave the `aria-label` out and use the error report as an incentive for editors to fix all the missing alt texts.

docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
mndonx: Thank you for the feedback, very helpful!

mndonx:
@cienvaras I've made those changes -- should I ask Tori about her preference now or after we send them the changes to review?

mndonx:
Also, I removed the "caption/credit" possibility. Describing what will happen if clicked seems to trump a description that might end up just being a photo credit.

docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
```diff
@@ -0,0 +1,26 @@
...
+{# set a generic link label in case there is no alt attribute on the image or a caption #}
+{% set link_label = "Open image in the gallery" %}
+{# If the image has an alt attribute, use it as the link label. #}
+{% if attributes['data-cbox-img-attrs']|render != '{&quot;alt&quot;:&quot;&quot;}' %}
+  {% set link_label = "Open " ~ attributes['data-cbox-img-attrs'] ~ " image in the gallery" %}
+{% endif %}
+{% set link_label = link_label|raw %}
+{% set link_label = link_label|replace({'{&quot;alt&quot;:&quot;':'', '&quot;}':''}) %}
+
+<a href="{{ url }}" aria-controls="colorbox" role="button" aria-label="{{ link_label }}" data-cbox-img-attrs="{{ link_label }}" {{ attributes }}>{{ image }}</a>
```
aasarava: @mndonx @cienvaras I don't think we want to replace the `data-cbox-img-attrs` value with a cleaned up label. That probably still needs to be in JSON format for whatever the Colorbox plugin does with it. My guess is that is uses the value in there to set the alt attribute on the expanded image.

aasarava: The rest of the plan (setting an appropriate, cleaned up `aria-label` attribute) makes sense to me. I'll update the AC on the ticket.

mndonx: That makes sense, thanks! I'll update the PR.

aasarava (approved):

docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
```diff
@@ -0,0 +1,26 @@
...
+#}
+
+{# set a generic link label in case there is no alt attribute on the image or a caption #}
+{% set link_label = "Open image in the gallery" %}
+{# If the image has an alt attribute, use it as the link label. #}
+{% if attributes['data-cbox-img-attrs']|render != '{&quot;alt&quot;:&quot;&quot;}' %}
+  {% set link_label = "Open " ~ attributes['data-cbox-img-attrs'] ~ " image in the gallery" %}
+{% endif %}
+{% set link_label = link_label|raw %}
+{% set link_label = link_label|replace({'{&quot;alt&quot;:&quot;':'', '&quot;}':''}) %}
```
cienvaras: This works, but checking the colorbox module code I found that `data-cbox-img-attrs` can include more attributes, not only `alt`. To prevent weird values, you can get the alt text from the `image` variable:

```
{% set alt_text = image['#alt'] %}
```

That would also make the use of the `|raw` filter unnecessary.

cienvaras (changes requested):

docroot/themes/humsci/humsci_basic/templates/contrib/colorbox-formatter.html.twig
mndonx: Oh my, you are right, we don't need that stuff anymore. Thanks!

cienvaras (approved):
Works great, thanks!

PR #1412, opened 2023-11-27
Migrate all ECK entities and types to hs_entity

pookmish:
# READY FOR REVIEW

## Steps to Test
1. checkout this branch
2. `composer install -n` (requires php 8.2. do `composer update` if you're on 8.1)
3. sync a sites `blt drupal:sync --site=[your choice]`
4. verify the upgrade is successful and the site still functions.

## PR Checklist


docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,438 @@ function su_humsci_profile_final_task() {
...
-  user_role_grant_permissions(UserInterface::ANONYMOUS_ROLE, $perms);
-  $perms = [
-    'create field_hs_hero_overlay_color',
-    'edit field_hs_hero_overlay_color',
-    'edit own field_hs_hero_overlay_color',
-    'view field_hs_hero_overlay_color',
-    'view own field_hs_hero_overlay_color',
-  ];
-  user_role_grant_permissions(UserInterface::AUTHENTICATED_ROLE, $perms);
+function su_humsci_profile_update_last_removed() {
```
joegl, line 55: What's this do?

pookmish: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_last_removed/10

When deleting update hooks you have to add/update this to the last update hook that was deleted.

joegl: Oh I didn't even know that. Neat, thanks!

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,444 @@ function su_humsci_profile_final_task() {
...
+        ->accessCheck(FALSE)
+        ->execute();
+      foreach ($ids as $id) {
+        $sandbox['items'][] = "$eck_name:$id";
+      }
+    }
+    $sandbox['items'] = array_filter($sandbox['items']);
+    $sandbox['count'] = count($sandbox['items']);
+    \Drupal::messenger()
+      ->addStatus(t('Updating @count ECK items.', ['@count' => count($sandbox['items'])]));
```
joegl, line 450: This is pedantic, but should the messenger message say "Migrating" instead of "Updating" 😆

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,444 @@ function su_humsci_profile_final_task() {
...
 
-/**
- * Disable hs_revision_cleanup module.
- */
-function su_humsci_profile_update_8204() {
-  \Drupal::service('module_installer')->uninstall(['hs_revision_cleanup']);
-}
+  $batched_items = array_splice($sandbox['items'], 0, 100);
+  foreach ($batched_items as $item) {
+    [$eck_type, $id] = explode(':', $item);
```
joegl, line 455: TIL about array destructuring shorthand. Cool

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,444 @@ function su_humsci_profile_final_task() {
...
-/**
- * Uninstall deprecated webform_jqueryui_buttons module.
- */
-function su_humsci_profile_update_8205() {
-  \Drupal::service('module_installer')->uninstall(['webform_jqueryui_buttons']);
-}
+  // When the conversion is complete, update all the field data.
+  if (!empty($sandbox['items'])) {
+    return;
+  }
```
joegl, line 487: I assume the return here just restarts the batch process on the next batch, so that once all the batching completes it executes the entity reference id migration below.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,444 @@ function su_humsci_profile_final_task() {
...
+      $convert_ids = $sandbox['convert'][$eck_type] ?? [];
+
+      $query = \Drupal::database()
+        ->select("{$entity_type}__$field_name", 'f')
+        ->fields('f', ['entity_id', 'delta', "{$field_name}_target_id"])
+        ->execute();
+
+      while ($entry = $query->fetchAssoc()) {
+        $new_target_id = $convert_ids[$entry["{$field_name}_target_id"]] ?? FALSE;
+        if (!$new_target_id > 0) {
```
joegl, line 520: I'm having a hard time understanding this conditonal. Is it checking if `$new_target_id` is not TRUE _and_ if it's greater than 0. Also wouldn't we want the $new_target_id to be a value greater than 0?

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
```diff
@@ -25,308 +29,444 @@ function su_humsci_profile_final_task() {
...
+
+  // Go through each of the ECK types and bundles and create new HS Entity
+  // bundles with the matching fields, form and display settings from the ECK.
+  foreach (_su_humsci_profile_update_eck_info() as $eck_name => $bundles) {
+    $eck_type = \Drupal::config("eck.eck_entity_type.$eck_name");
+    foreach ($bundles as $bundle_id => $new_bundle_id) {
+      $eck_bundle = \Drupal::config("eck.eck_type.$eck_name.$bundle_id");
+
+      $eck_form = EntityFormDisplay::load("$eck_name.$bundle_id.default");
+      $eck_display = EntityViewDisplay::load("$eck_name.$bundle_id.default");
```
joegl, line 304: This looks like it's only going to clone the default EntityViewDisplay and EntityFormDisplay for each bundle. Do any ECK bundles have additional View or Form displays we'd have to account for?

joegl (commented):
I spent some time thoroughly reviewing the code (and learned a lot). I had a few comments and questions. I have not reviewed the views update code as thoroughly yet. This is pretty awesome.

I've also pulled this down and tested on a number of sites locally (including the giant Economics site) and haven't noticed any issues with the migration process.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
pookmish: hmm. depends on which way you look at it I guess. Or should it be `UpGRADEing` them? haha

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
pookmish: Yep that exactly what's happening here. We can't pass sandbox to the next update hook so we gotta stay in the same one.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
pookmish: I can't remember what the use case here was.  There was a site I tested on that had some orphans or something. It should always be greater than 0, but there were some that acted weird. I don't recall which site either.

docroot/profiles/humsci/su_humsci_profile/su_humsci_profile.install
pookmish: It's unlikely. I don't believe they use view displays to the extent that you and I do.

joegl:
@pookmish I merged this with the 10.0.6 release tag and created a new deployment artifact. Then I deployed 10.0.6 to the dev environment and copied down all prod databases. After the databased finished copying down, I deployed this ECK deployment artifact. There were a number of errors during the deployment I've documented below. I've also copied the entire log to a file I've shared with you here: https://drive.google.com/file/d/1GGqqV0S5WEQPLNvGJ2oK9_bK2jb87T_O/view

```
[error]  Adding non-existent permissions to a role is not allowed. The incorrect permissions are "opt-in or out of google analytics tracking"
```
In: su_humsci_profile_update_9704 
On: 49 sites.
Comments: Some of these sites (but not all) didn't run any of the other updates before 9704 in the logs (not sure why). I'm not sure how this permission is related to this update hook.

```
[error]  InvalidArgumentException: Field field_url is unknown. in Drupal\Core\Entity\ContentEntityBase->getTranslatedField() (line 583 of /mnt/www/html/humscigryphondev/docroot/core/lib/Drupal/Core/Entity/ContentEntityBase.php). 
```
In: su_humsci_profile_update_9701
On: 3 sites (biology, music, west)

```
Call to a member function bundle() on null
```
In: su_humsci_profile_update_9701
On: 7 sites (chemistry, dlcl, iranianstudies, oconnell, politicalscience, sitp, southasia)

```
[error]  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'humscigrydbNUMBER.event_collections__field_session_information' doesn't exist: SELECT 1 AS "expression" FROM "event_collections__field_session_information" "t" WHERE ("field_session_information_value" IS NOT NULL) OR ("field_session_information_summary" IS NOT NULL) OR ("field_session_information_format" IS NOT NULL) LIMIT 1 OFFSET 0; Array ()
```
In: su_humsci_profile_update_9704
On: 2 sites (jewishstudies, philit2023)
Comments: These two didn't run any of the other updates before 9704 in the logs. Not sure why.

```
[error]  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'admin_toolbar_tools.extra_links:entity.hs_entity_type.edit_form.' for key 'id': INSERT INTO "menu_tree" ("id", "menu_name") VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1); Array (     [:db_insert_placeholder_0] => admin_toolbar_tools.extra_links:entity.hs_entity_type.edit_form.course_collections__instructor     [:db_insert_placeholder_1] => admin )
```
In: su_humsci_profile_update_9700
On: psychology

I'd like to try copying the databases down again with this code active instead of doing the 10.0.6 pull down and deploy I did here. But I'm going to wait until I hear from you in case you want to investigate any of these on the dev environment.

jbickar:
Where are we on this?

pookmish:
The PM, the client and the dev team are all aware that we still have adjustments to do. All involved still communicate about it regularly.

joegl:
@pookmish I've been investigating all the views with broken/missing handlers related to ECK relationships. There were two main culprits:

1. The majority of ECK relationships seem to be intact but any field/filter using the "title" field from the relationship was broken. This is because we're migrating the "title" field on ECK to the "label" field on the `hs_entity`. The key problem here is the label field is part of the entity definition for normal Drupal entities (such as `hs_entity`) but is not for any ECK entities, so the table to pull the field data from is actually just `hs_entity` and not `hs_entity_field_data`. I've updated the upgrade code and exported configuration in the linked commit to address it (and tested it locally): https://github.com/SU-HSDO/suhumsci/pull/1412/commits/f2df83ece53ff46be65d671230f375eb76ecf83b . H&S uses the title/label field quite a lot in these views so this actually clears up around 90% or more of the broken views fields from what I can see.

2. All reverse relationships to an ECK entity are broken in the view. I haven't been able to figure out yet if there is anything we can do to address these in the migration or if we should just alert H&S to any views with a reverse relationship to an ECK field so they can address them.

joegl:
> @pookmish I've been investigating all the views with broken/missing handlers related to ECK relationships. There were two main culprits:
> 
> 1. The majority of ECK relationships seem to be intact but any field/filter using the "title" field from the relationship was broken. This is because we're migrating the "title" field on ECK to the "label" field on the `hs_entity`. The key problem here is the label field is part of the entity definition for normal Drupal entities (such as `hs_entity`) but is not for any ECK entities, so the table to pull the field data from is actually just `hs_entity` and not `hs_entity_field_data`. I've updated the upgrade code and exported configuration in the linked commit to address it (and tested it locally): [f2df83e](https://github.com/SU-HSDO/suhumsci/commit/f2df83ece53ff46be65d671230f375eb76ecf83b) . H&S uses the title/label field quite a lot in these views so this actually clears up around 90% or more of the broken views fields from what I can see.
> 2. All reverse relationships to an ECK entity are broken in the view. I haven't been able to figure out yet if there is anything we can do to address these in the migration or if we should just alert H&S to any views with a reverse relationship to an ECK field so they can address them.

@pookmish 

I think I actually fixed it for all views issues now. The HumSci Entity does not have a `field_data` table like most other entities (for whatever reason 🤷). The `field_data` table usually stores things like the bundle, uid, uuid, changed date, created date, etc., These columns are part of the base `hs_entity` and no `hs_entity_field_data` table exists. I've updated the upgrade code accordingly, tested locally, and all views appear to migrate without issues now: https://github.com/SU-HSDO/suhumsci/pull/1412/commits/8d8aa3a1c4a01adbb91ea6aefb39270b2d36512b

joegl:
@pookmish I pushed another commit to fix the reverse relationships: https://github.com/SU-HSDO/suhumsci/pull/1412/commits/f93985ba31caf15a73a537b63b8e828110a5c811

It could maybe use an improvement or two, but seems to clear up the remaining issues with views migrations.

PR #1418, opened 2023-12-12
SHS 5154 layout builder theming

mndonx:
# READY FOR REVIEW

## Summary
This PR: 
- Ensures checkboxes are close to their corresponding field labels in layout builder
- Ensures the Revision Information section is closed by default.
- Ensures that the field label and field is clear and distinguishable from other fields. 

## Need Review By (Date)
_['12/12']_

## Urgency
_['medium']_

## Steps to Test
1. Visit a page that uses layout builder (like the homepage)
2. Edit the homepage and click layout
3. Verify that checkboxes have been themed to be smaller and next to their label
4. Verify that the revision information detail element is closed by default
5. Note that the fields and labels are better distinguished (grouped together, with more margin between each pair.) 

## PR Checklist


cienvaras:
@mndonx Can you please take a look at the Codeclimate errors?

mndonx:
@cienvaras Sorry about that, this is ready for re-review. Thanks!

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -45,6 +45,16 @@ function hs_layouts_preprocess_container(&$variables) {
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function hs_layouts_preprocess_details(&$variables) {
```
cienvaras: Please remove this preprocess, it's not doing anything as you're using JS for collapsing the element.

docroot/themes/humsci/humsci_basic/src/js/shared/vertical-tabs/vertical-tabs.js
```diff
@@ -0,0 +1,10 @@
+function closeDetails() {
+  // Close Revision Information Details element by default.
+  if (document.getElementById('edit-revision-information')) {
```
cienvaras: Is it possible to use a more specific selector? There can be other element with the same ID in a different form and we need to make sure that we're targeting the right one.

cienvaras (changes requested):
Works good, I just found a couple of things that can be improved.

docroot/modules/humsci/hs_layouts/hs_layouts.module
mndonx: Ah, yes, I was wondering. I saw this worked at first, but then realized that JS was overriding it on page load.

mndonx:
@cienvaras Ready for re-review. Thank you!

cienvaras (approved):
Thanks! Works as expected.

PR #1422, opened 2023-12-19
SSH-3160: Update carousel controls placement

mariannuar:
# READY FOR REVIEW

## Summary
This PR fixes the carousel controls placements on the traditional and colorful theme

## Urgency
low

## Steps to Test
1. Log in to hs_traditional and hs_colorful locally.
2. Visit a carousel, such as on one of the example pages below
3. Confirm the following behavior is happening:
   - On Traditional theme, dot navigation should be horizontally aligned with arrow navigation, same as it is on Colorful theme.
   - Arrow navigation should always stay at same distance from bottom right edges of the container and not jump outside.

Traditional:
https://politicalscience.stanford.edu/
Current HP: 

Mobile:

Colorful: 
https://religiousstudies.stanford.edu/
Current HP:

Mobile:

## PR Checklist


cienvaras (changes requested):
@mariannuar Looks good, but I found a couple of cases when the elements don't work as expected. A couple of examples:

In traditional, when the resolution changes from 768px to 767px, the arrows are misaligned:

Screenshot 2023-12-22 at 16 03 14

Going 1px down to 766px puts the arrow back inside the carousel. This only happens if the resolution decreases, going from 766 to 767px doesn't have any effect.

There are other cases when the arrows and dots jump and then go back to the original position. For example, when switching from 1200 to 1199px

These jumps might not affect the end users as they usually don't resize the windows in a way that this is noticeable, but it's included in the ticket, so it's necessary to find a fix or at least find the cause and confirm that there are no negative effects on the UX for the typical use cases

I think there are 2 possible causes:

- The carousel JS is changing sizes/styles and  there's a delay or a minor difference in the breakpoints. If this is the case, it might be a bit difficult to fix but lets at least document the issue and the cause
- The image styles breakpoints might be slightly different than the ones used in the CSS and there might be a jump when a new image loads.

mariannuar:
@cienvaras Yes, you're totally right. There's a delay in carousel JS when changing sizes/styles when the window is resized. A way to fix this (but I'm not sure what are your thoughts about it) is to change also the position of the arrows and dots through JS, I think that's the only way. I've been thinking and trying through CSS but nothing seems to fix because of the delay. However, I don't think there are negative effects on the UX for these use cases, since they won't usually resize windows in this way as you mentioned. Let me know what are you thoughts about this. Thanks!

mariannuar:
@cienvaras This is ready for review again. 😅

mariannuar:
@cienvaras This is ready for review again. Here are some screen recordings of how it's currently working for me. Let me know if there are still issues related to js for you. Hopefully, this is working now :( :

- https://www.loom.com/share/ec0b3a5ed3a049d2a1062ae7fe19bd80?sid=b653fc13-8c62-4bd8-98f0-b478c0a038b2
- https://www.loom.com/share/41b4f1c83cfb4b3db124550735467ddd?sid=e1dc681e-0880-4684-9a81-04c3f445067d

docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.carousel.scss
```diff
@@ -504,17 +504,65 @@
       }
     }
 
+    // There was some css being added by a js library that was also causing the issue of the arrows position.
+    .slick-list {
+      @include hb-colorful {
```
cienvaras: If you're going to add the same code to both themes, you don't need to use the `@include` mixin. Just add the rules in a regular way.

docroot/themes/humsci/humsci_basic/src/scss/components/_pattern.carousel.scss
```diff
@@ -504,17 +504,65 @@
...
+      }
+    }
+
     &__arrow {
       margin: 0;
       padding: 0;
       font-size: 0;
 
+      @include hb-colorful {
+        position: absolute;
```
cienvaras: Same as above. Shared sass rules should be added without using `@include`.

cienvaras (changes requested):
@mariannuar Looks better, we're almost there! There are a couple of improvements in the code, please check the inline comments. Also, a couple of minor issues:

1. In traditional, the arrows are slightly misaligned with the dots. Ideally, the midpoint of the arrows should be aligned with the midpoint of the dots:

    Screenshot 2024-01-22 at 10 54 24

2. In Colorful, there's some overlap between the arrows and the content around 775px:

    Screenshot 2024-01-22 at 11 06 27

mariannuar:
@cienvaras Thank you!!! This is ready for review again.

cienvaras (approved):
@mariannuar Looks good! 🚀 

Putting merge on hold for a bit until Sprint 41 branch gets approved and merged, to avoid conflicts.

cienvaras:
@mariannuar I'm closing this PR to create a new one containing only the commits specific to this ticket. This branch has commits included in Sprint 41 release, but that release haven't been merge and I don't want to add them to the Sprint 42 release.

PR #1428, opened 2024-01-03
SHS-5171: adjusting titles to include date info

mndonx:
# READY FOR REVIEW

## Summary
This PR adds date information to event titles so events with the same name will have differentiating information within the title's link for accessibility. The date is visually hidden. 

## Urgency
medium

## Steps to Test
1. Set up the Statistics site locally
2. Visit the homepage
3. Note that the date component has multiple events with the same name, but now have hidden text showing the date. 

I was hoping for a cleaner solution, but the link is built in the view. Let me know if you think I should do this by rewriting the title to include the date from in the View itself. 

## PR Checklist


cienvaras:
@mndonx I had some problems setting up statistics in my local for testing this and also I think we need to review the approach. We're hardcoding some things in the template (the `<h2>` for example) that might cause problems when the pattern is used in other contexts or when the view have a different configuration. I'll put this PR on hold and revisit it next week.

cienvaras:
Closing this PR, fixing this in the view configuration is a better approach.

cienvaras:
Closing this PR, fixing this in the view configuration is a better approach.

PR #1451, opened 2024-02-09
SHS-5052: Video preview in CKeditor should better reflect rendered size

mariannuar:
# READY FOR REVIEW

## Summary
When embed a video in a CKEditor textarea and choosing the "Align center and break text" alignment, then the video preview is the full width of the CKEditor area to indicate that it will be rendered at full width.

## Need Review By (Date)
02/09

## Urgency
medium

## Steps to Test on Colorful and Traditional
1. Go to any page, edit it or create a new one
2. In a WYSIWYG area, embed a video and select the "Align center and break text" alignment. 
3. The preview of the video should be the full width of the CKEditor area

## PR Checklist


docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
```diff
@@ -322,6 +322,11 @@ $px-only: true;
       }
     }
   }
+
+  // Fix Video preview in CKeditor for "Align center and break text" option
+  .drupal-media-style-align-center {
```
cienvaras, line 327: This always applies to images and I think this is an unintended side effect. Ideally we should add different styles to images and videos. I don't think this is possible with the current markup, but we can add a preprocess to add an extra attribute with the media type. Let's review this next week.

cienvaras (changes requested):

docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
mariannuar: Oh yes, you're right! I forgot about images! Yes. I can work on that next week. Thanks!

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -295,3 +295,11 @@ function humsci_basic_preprocess_paragraph(&$variables) {
     }
   }
 }
+
+/**
```
cienvaras: This preprocess is not necessary, please remove it. I initially thought that the `data-` attribute was going to be applied on the `<figure>`, but it's applied on the `<article>` that contains the media entity. That element already has a class that indicates the type (`hb-media-image`, `hb-media-video`, etc.)

docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
```diff
@@ -322,6 +322,19 @@ $px-only: true;
...
+    max-width: none;
+
+    [data-media-source='video'] {
+      max-width: 100%;
+    }
+
+    [data-media-source='image'] {
+      max-width: 25%;
```
cienvaras: Please use `50%` here, to ensure that the image looks exactly the same as it does in the current version.

Also, keep in mind the following:
- There are more media types besides images and video. All the other media types should look exactly the same as before adding these changes. Maybe you can use `not()` to target everything except video.
- The media element should be centered. Maybe you need to add `margin-left: auto; margin-right: auto;`

cienvaras (changes requested):
Still some things to fix. Please check the comments in the code.

docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
mariannuar: Thank you!!!

docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
```diff
@@ -322,6 +322,21 @@ $px-only: true;
       }
     }
   }
+
+  // Fix Video preview in CKeditor for "Align center and break text" option
+  .drupal-media-style-align-center {
+    max-width: none;
+
+    [class*="hb-media"]:not(.hb-media-video) {
```
cienvaras: I'm pretty sure all media elements use the `<article>` tag. In that case, this rule can be simplified:

```suggestion
    article:not(.hb-media-video) {
```

cienvaras (changes requested):
Looks great! Just a minor code improvement left, please check the code comments.

mariannuar:
@cienvaras Thank you. Code improved!

cienvaras (approved):
🚀

PR #1466, opened 2024-02-29
SHS-5501: Add "Raised Cards" and "Uniform Height" options to Private Collection paragraph

hectorlj:
# READY FOR REVIEW

## Summary
Adds Uniform Height and Raised Cards checkboxes to the Private Collection Component

## Steps to Test
1. Create new Private Page
2. Click on the `Add Component` button
3. Add a Private Collection component
4. Verify that clicking the `Raised Cards` checkbox causes the `Uniform Height` checkbox to appear, already enabled.
5. Set the `Items Per Row` to 2
6. Disable the `Uniform Height` checkbox
8. Using the `Add Component` button in the Collection component, add 2 Postcard components
9. Add a paragraph or two to the first component and add only one line to the second component
10. Save the new page and navigate to it
11. Verify that the postcards are different heights, but still have the styling from the `hb-raised-cards` class
12. Edit the new page
13. Click on the `Edit` button on the Collection component
14. Enable the `Uniform Height` checkbox
15. Save the page and navigate to it
16. Verify that the postcards are now the same height and still have the styling from the `hb-raised-cards` class

## Notes
- May need some help on the front-end. The uniform height class is applied when the option is unchecked. This behavior does not match the `Collection` component.
- No update hook was needed since the configuration files are not affected by config_ignore, and the `Private Collection` component did not already have the `Raised Cards` checkbox

## PR Checklist


cienvaras:
@hectorlj Code looks good and works as expected, but definitely needs some FE work. I'll keep the PR open to add the extra code here. Thanks!

PR #1470, opened 2024-03-04
Four Kitchens Release - Sprint 45

cienvaras:
# READY FOR REVIEW

## Summary
- SHS-4946 - Implement update to Spotlight color values
- SHS-4945 - Implement update to Tertiary color values
- SHS-5498 - Shortcut menu is not displaying in admin toolbar for site editors when they log in
- SHS-5497 - Respect reduced motion for nav caret animation
- SHS-5504 - Implementation: Manage Layout not working properly on Mathematics site
- SHS-5505 - Banner Image (s) - with partial overlay and text Title is not aligned properly when right aligned
- SHS-5531 - Bug: Arrows misaligned on carousel on one site
- SHS-5501 - Add "Raised Cards" and "Uniform Height" options to Private Collection paragraph

## Need Review By (Date)
03/13

## Urgency
medium

### SHS-4946
1.  Check that `spotlight-lighten-8` is no longer in any color mapping file and is not being referenced anywhere.
2. Confirm the [`Color Band`](https://hs-colorful.suhumsci.loc/color-bands) as the `Light` option color is close to the `Spotlight-BG`(spotlight) color

### SHS-4945
1. Check that `tertiary-reversed-darken-10` is no longer in any color mapping file and is not being referenced anywhere.

### SHS-5498
1. Login as a site admin and contributor
2. Shortcut toolbar menu is opened by default

### SHS-5497
1. Go to any page
2. Check the nav caret animation in all the kind of navigation (main nav, sidebar nav)
3. On your computer, go to System Settings > Accessibility > Display and turn on the `Reduce motion` setting
4. Confirm there's no nav caret animation

### SHS-5504

1. Create a flexible page.
2. Add two or more postcards with the "Style" field set to "Vertical Linked Card". Leave the "Link" field empty in one of them.
3. Save the page. Confirm that there are no errors in the javascript console.
4. Confirm that the postcards with links work as expected.

### SHS-5505

1. Visit a page with a "Banner Image (s) - with partial overlay and text" component (for example https://hs-traditional.suhumsci.loc/)
2. Confirm that at least one of the slides is right aligned. If not, edit the paragraph and change the alignment of one of the slides to "right"
3. Confirm that the title is correctly aligned in all resolutions

Screenshot 2024-03-04 at 15 23 41

### SHS-5531

1. Create a flexible page on https://clas.suhumsci.loc/ or https://hs-traditional.suhumsci.loc/
2. Add a Banner image(s) with text box on the Full-width Region. Add some slides
3. Confirm the arrow and dots are well aligned as expected
4. Plus, test this also on Colorful (When testing this, saw that the arrows were misaligned (out of the box) in small viewports)

### SHS-5501

Create new Private Page
Click on the Add Component button
Add a Private Collection component
Verify that clicking the Raised Cards checkbox causes the Uniform Height checkbox to appear, already enabled.
Set the Items Per Row to 2
Disable the Uniform Height checkbox
Using the Add Component button in the Collection component, add 2 Postcard components
Add a paragraph or two to the first component and add only one line to the second component
Save the new page and navigate to it
Verify that the postcards are different heights, but still have the styling from the hb-raised-cards class
Edit the new page
Click on the Edit button on the Collection component
Enable the Uniform Height checkbox
Save the page and navigate to it
Verify that the postcards are now the same height and still have the styling from the hb-raised-cards class

## PR Checklist


docroot/modules/humsci/hs_admin/hs_admin.module
```diff
@@ -19,6 +20,12 @@ function hs_admin_page_attachments(&$variables) {
   if ($logged_in) {
     $variables['#attached']['library'][] = 'hs_admin/hs_admin';
   }
+
+  if (\Drupal::state()->get('set_default_shortcuts: ' . \Drupal::currentUser()->id())) {
+    $variables['#attached']['library'][] = 'hs_admin/default_shortcuts';
+    // Remove state after the first check.
+    \Drupal::state()->delete('set_default_shortcuts: ' . \Drupal::currentUser()->id());
```
pookmish: Why wouldn't you add the library all the time and use the logic in the JS to do what it's supposed to. Using the state isn't really what state is intended for and you're basically adding it, just to remove it.

docroot/modules/humsci/hs_admin/hs_admin.libraries.yml
```diff
@@ -5,3 +5,8 @@ hs_admin:
       assets/css/hs_admin.css: {}
   js:
     assets/js/hs_admin.js: {}
+
+default_shortcuts:
+  version: VERSION
```
joegl:
```suggestion
  version: VERSION
```

I would drop the `version: VERSION` from library definitions unless you intend to use a version and update it with every change. We discovered a [recent change](https://www.drupal.org/node/3301716) in Drupal core 10.1+ doesn't re-aggregate library files unless the library version is changed. Using `VERSION` here as the version reflects the Drupal core version -- if Drupal Core does not get upgraded in a release, it wouldn’t re-aggregate the CSS/JS in the library. If you leave a version out entirely, it will get a unique hash and re-aggregate with every change.

docroot/modules/humsci/hs_admin/hs_admin.module
```diff
@@ -196,3 +203,13 @@ function hs_admin_field_widget_complete_paragraphs_browser_form_alter(&$field_wi
     }
   }
 }
+
+/**
+ * Implements hook_user_login().
+ */
+function hs_admin_user_login(UserInterface $account) {
+  if ($account->isAuthenticated() && $account->hasPermission('access toolbar')) {
```
joegl: Instead of setting a state, why not do the access check when adding the library in the `hook_preprocess_node_edit_form`?

docroot/modules/humsci/hs_admin/hs_admin.libraries.yml
ahughes3: @cienvaras

docroot/modules/humsci/hs_admin/hs_admin.module
ahughes3: @cienvaras

docroot/modules/humsci/hs_admin/hs_admin.libraries.yml
cienvaras: @joegl Thanks, fixed.

docroot/modules/humsci/hs_admin/hs_admin.module
cienvaras: @joegl We need to execute the JS code only once when the user logs in. It's not possible to add and execute the JS  in `hook_user_login()`, so the only way that we found to do this was to set a temporary state for loading the JS in `hook_page_attachments()` once and then remove it to prevent it from loading again. We considered adding all the logic in Javascript as @pookmish suggested in a previous related comment, but we couldn't find a way to react to login events from javascript.

docroot/modules/humsci/hs_admin/hs_admin.module
cienvaras: Please see the [comment below](#discussion_r1520447847)

docroot/modules/humsci/hs_admin/hs_admin.module
pookmish: You're already using the local storage in the JS anyways, why not just leverage it a little bit more.

 ```
if window.localStorage.getItem('do-the-thing') !== "false"
do the thing
window.localStorage.setItem('do-the-thing', false)
```

Or use a cookie that only active for this window

Comment (cienvaras) [reply docroot/modules/humsci/hs_admin/hs_admin.module]: @pookmish @joegl Thanks for the advice, I just pushed the updated code. Moved the logic to the JS code as requested and removed the state.

File: docroot/modules/humsci/hs_admin/hs_admin.module [hunk 4]
```diff
@@ -19,6 +20,9 @@ function hs_admin_page_attachments(&$variables) {
   if ($logged_in) {
     $variables['#attached']['library'][] = 'hs_admin/hs_admin';
   }
+
+  // Show shortcuts toolbar when user is logged in.
+  $variables['#attached']['library'][] = 'hs_admin/default_shortcuts';
```
pookmish, line 25: I still think you should add the access check for `access toolbar`. That'll avoid the JS adding to anon users. Or you could just add it to the `hs_admin` library.

docroot/modules/humsci/hs_admin/assets/js/default_shortcuts.js
```diff
@@ -0,0 +1,20 @@
...
+    if (!settings.user.uid) {
+      // If the user is not logged in, clear the active toolbar tab ID.
+      window.localStorage.removeItem('Drupal.toolbar.activeTabID');
+      return;
+    }
+
+    // Get the toolbar shortcuts toolbar button. Only users with the
+    // 'access shortcuts' permission will have this button.
+    const shortcutsItem = context.querySelector('#toolbar-item-shortcuts');
+    console.log('shortcutsItem', shortcutsItem);
```
pookmish: Do you need the log?

pookmish (changes requested):

docroot/modules/humsci/hs_admin/assets/js/default_shortcuts.js
cienvaras: Just removed it, sorry.

docroot/modules/humsci/hs_admin/hs_admin.module
cienvaras: We need the script to run for anon users to clear the `Drupal.toolbar.activeTabID` localstorage item when the users log out. Otherwise that item will persist and we wouldn't be able to switch the selected toolbar only on login.

pookmish: I see now. I'm not a big fan of that approach since it runs on every page for ever user and every ajax. I feel there's probably a way to accomplish this but I'm unable to think of an alternative at this time. The speed at which those couple lines runs is fast and doesn't impact anywhere else on the site, so that's the main thing.

Any chance you've search d.o to see if someone mentioned anything about the storage sticking around?

cienvaras: > I'm not a big fan of that approach since it runs on every page for ever user and every ajax.

Yes, I agree. That's why we did the initial implementation using state, but I understand why that approach is not ideal either.

Even when the JS runs in every case, we think that it's impact on performance is small, specially for anon users. If you consider that's helpful, we can add extra checks to return early if the code is executed in an ajax call (using `once` or checking that `context` matches whole document).

> Any chance you've search d.o to see if someone mentioned anything about the storage sticking around?

No, we couldn't find anything relevant for this use case. From what I see, persisting the selected toolbar state is the designed behavior, but we need to override it to improve usability for our users.

pookmish: > we can add extra checks to return early if the code is executed in an ajax call

I actually think doing that would decrease the performance of this part, so I think what you currently have is best case for this situation. It all runs synchronously with other behaviors I believe anyways.

Thanks for checking d.o. That's too bad it's kinda designed that way. It would be nice if they provided a setting to change that behavior. :shrug:

pookmish (approved):
I'm ok with this if @joegl is good too.

joegl:
> I'm ok with this if @joegl is good too.

Yup I'm just re-running a flaky test.

joegl (approved):

PR #1475, opened 2024-03-08
HSD8-000: Add patch to allow group blocks in layout builder

joegl:
# READY FOR REVIEW

## Summary
Add patch to allow group blocks in layout builder

## Need Review By (Date)
_['10/30', 'asap', etc.]_

## Urgency

## Steps to Test
1. _[First testing step]_
2. ...

## PR Checklist

### DIFF
--- a/patches/layout_builder-allow-hs-blocks.patch
+++ b/patches/layout_builder-allow-hs-blocks.patch
@@ -1,0 +1,13 @@
+diff --git a/modules/layout_builder/js/layout-builder.js b/modules/layout_builder/js/layout-builder.js
+index 76b751c6..9353579d 100644
+--- a/modules/layout_builder/js/layout-builder.js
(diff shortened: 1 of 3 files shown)


patches/layout_builder-allow-hs-blocks.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/modules/layout_builder/js/layout-builder.js b/modules/layout_builder/js/layout-builder.js
```
pookmish, line 1: I never know when it has to be `a/core/modules...` and when it doesn't. One of those weird things.

pookmish (approved):

patches/layout_builder-allow-hs-blocks.patch
joegl: Yup took me a bit lol

PR #1488, opened 2024-03-28
SHS-5560:  Skip to Main Content

cienvaras:
# READY FOR REVIEW

## Summary
- Add "Main Content Anchor Target" block provided by SWS
- Add custom template and theming for "Main Content Anchor Target"
- Add update hooks to add block to all existing default and overridden layouts
- Remove `main-content` id from all node pages
- Update page main content container to use `<main>` element
- Add fallback js to add the `main-content` id if not present in page

## Urgency
medium

## Steps to Test
1. Visit default layout builder configuration for all content types. Confirm that the "Main Content Anchor Target" is present in all cases
2. Find a node with an overridden layout configuration (*). Click on the "Layout" tab and confirm that the "Main Content Anchor Target" is present
3. Select some random pages and tab to the "Skip to Main Content" link. Click it or press enter to confirm that it works as expected
4. Remove the "Main Content Anchor Target" from a default or overridden layout. Visit the page and confirm that the js fallback works by using the "Skip to Main Content" link
5. Inspect any page and confirm that there's a `<main>` element wrapping the page content (everything except the header and footer)

(*) You can find the ids of the nodes with overridden layout executing the following query in the SQL console:

```
SELECT entity_id FROM node__layout_builder__layout;
```

## PR Checklist


cienvaras:
@joegl This one is ready to review. We used the "Main Content Anchor Target" block you provided and it works really good for our requirements, thanks a lot for the suggestions and the  code.

There's still a problem though, We're still having errors in some sites when running the update hook that adds to block to the nodes with an overridden layout:

```
>  [notice] Update started: hs_layouts_update_10002
>  [error]  Drupal\purge\Plugin\Purge\Invalidation\Exception\InvalidExpressionException: The URL is invalid. in Drupal\purge\Plugin\Purge\Invalidation\UrlInvalidation->validateExpression() (line 84 of /app/docroot/modules/contrib/purge/src/Plugin/Purge/Invalidation/UrlInvalidation.php). 
>  [error]  Drupal\Core\Entity\EntityStorageException: The URL is invalid. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 
>  [error]  Drupal\Core\Entity\EntityStorageException: The URL is invalid. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 
>  [error]  The URL is invalid. 
>  [error]  Update failed: hs_layouts_update_10002 
```

From the debugging we did, the problem is caused because the uri used by Drush when invoked by BTL has an underscore on it:

```
/app/vendor/bin/drush @self updb --uri=hs_traditional --no-interaction --ansi in /app/docroot
```

I'm pretty sure I saw the same issue in Acquia when we tested the old implementation in the dev environment last year, so I think it's a good idea to take a look before merging the branch. Please let us know if you need more info or any action from us.

docroot/modules/humsci/hs_layouts/hs_layouts.install
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  $sandbox['#finished'] = count($sandbox['ids']) ? 1 - count($sandbox['ids']) / $sandbox['total'] : 1;
+}
+
+/**
+ * Add the main content anchor block to the overridden layouts.
+ */
+function hs_layouts_update_10002(&$sandbox) {
+  $nodeStorage = \Drupal::entityTypeManager()->getStorage('node');
```
joegl, line 59:
```suggestion
  $node_storage = \Drupal::entityTypeManager()->getStorage('node');
```

Formatting for consistency. Not a big deal though.

joegl:
> @joegl This one is ready to review. We used the "Main Content Anchor Target" block you provided and it works really good for our requirements, thanks a lot for the suggestions and the code.
> 
> There's still a problem though, We're still having errors in some sites when running the update hook that adds to block to the nodes with an overridden layout:
> 
> ```
> >  [notice] Update started: hs_layouts_update_10002
> >  [error]  Drupal\purge\Plugin\Purge\Invalidation\Exception\InvalidExpressionException: The URL is invalid. in Drupal\purge\Plugin\Purge\Invalidation\UrlInvalidation->validateExpression() (line 84 of /app/docroot/modules/contrib/purge/src/Plugin/Purge/Invalidation/UrlInvalidation.php). 
> >  [error]  Drupal\Core\Entity\EntityStorageException: The URL is invalid. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 
> >  [error]  Drupal\Core\Entity\EntityStorageException: The URL is invalid. in Drupal\Core\Entity\Sql\SqlContentEntityStorage->save() (line 817 of /app/docroot/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php). 
> >  [error]  The URL is invalid. 
> >  [error]  Update failed: hs_layouts_update_10002 
> ```
> 
> From the debugging we did, the problem is caused because the uri used by Drush when invoked by BTL has an underscore on it:
> 
> ```
> /app/vendor/bin/drush @self updb --uri=hs_traditional --no-interaction --ansi in /app/docroot
> ```
> 
> I'm pretty sure I saw the same issue in Acquia when we tested the old implementation in the dev environment last year, so I think it's a good idea to take a look before merging the branch. Please let us know if you need more info or any action from us.

This isn't due to the drush command. The `--uri` in this case refers to the [drush/multisite alias for the site](https://github.com/SU-HSDO/suhumsci/blob/edeca9d55aa89fe99135898d74c3bbfe0f8f9644/blt/blt.yml#L111), so the underscore is correct.

The update hook goes through most the nodes on the site and re-saves them. There must be a problem node this is coming from.

joegl:
@cienvaras I pulled the `hs_traditional` site locally and I ran into an error on node 12231 where the `layout_builder__layout` field did not exist. You may want to adjust the query condition (not NULL may not be enough). Or maybe add a line like the below into the foreach loop:

```
  if (!$node->hasField('layout_builder__layout')) {
    continue;
  }
```

I was not able to replicate the URI error locally.

docroot/modules/humsci/hs_layouts/hs_layouts.install
cienvaras: Fixed.

cienvaras:
@joegl 

> I pulled the hs_traditional site locally and I ran into an error on node 12231 where the layout_builder__layout field did not exist.

I wasn't able to reproduce this issue, but I added the validation you suggested, to be one the safe side:

I'm still having the URL validation issue. I did some debugging and was able to find the exact problem:

- Error only happens when `Purge` module is enabled (it's disabled in local when running  `blt drupal:sync`, but after the update hooks are executed)
- It affects nodes that:
  - Have the option "Generate automatic URL alias" checked
  - Have an alias that doesn't match the `pathauto` pattern
- Error is triggered when a new alias is created, using `pathauto`. There's a `postSave` hook in the Purge module tries to validate the new path and fails because it finds an underscore in the URL
- Some nodes that trigger the error:
  - Traditional:
    - `/vertical-timeline-without-and-collection` (12136)
    - `/collection-real-examples-uniform-and-stretch` (12296)
    - `/accordions-secondary` (16296)
  - Colorful:
    - `/collections-cards-uniform-and-stretch-1` (12126)
    - `/collection-real-examples-uniform-and-stretch` (12296)
    - `/accordions-secondary` (16296)

If you think that this is not going to be a problem in Acquia, I think we can go ahead and merge the PR (assuming there's no more feedback). Otherwise, we can find a workaround to ensure that the update hook doesn't fail.

joegl:
> If you think that this is not going to be a problem in Acquia, I think we can go ahead and merge the PR (assuming there's no more feedback). Otherwise, we can find a workaround to ensure that the update hook doesn't fail.

Thanks for the updated investigation and notes! 👍 

I tested the [purge validation function](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21UrlHelper.php/function/UrlHelper%3A%3AisValid/10) that is failing with a number of test URL's:
```
    if (!UrlHelper::isValid($url, TRUE)) {
      throw new InvalidExpressionException('The URL is invalid.');
    }
``` 

As you mentioned, any URL with an underscore in the domain failed validation: `$url = "https://test_site.stanford.edu/my-url";`

I'm guessing the environment being tested in is using an underscore in the domain. This goes back to your original point/question about the drush `uri`. If that's the case, then this should not be an issue on Acquia.

If you can confirm the test URL uses underscores then we can merge this in.

Also, I apologize for all the edits on my comment 😄

joegl (approved):
Plan to merge this once tests pass. Thanks @cienvaras

cienvaras:
Thanks @joegl! And yes, I've only had issues on sites with an underscore in their machine name (`hs_traditional`, `hs_colorful`, etc.)

joegl:
@cienvaras A few conflicts to resolve but then we can merge

cienvaras:
@joegl Conflicts fixed

PR #1489, opened 2024-03-28
SHS-4277: Compile assets on server

hectorlj:
# READY FOR REVIEW

## Summary
Adds build commands to compile front end assets on deploy

## Steps to Test
1. Check that all the github actions builds are successful, since they use the blt configuration to build the containers so it should be similar
2. Check that the theme works on the branch's tugboat preview ( Both sites use `U: tugboat P: tugboat`):
   - [ ] [Traditional theme tugboat preview](https://hs-traditional-trvzsw50a6qpe6iqdqqmtdnwil8pnzot.tugboatqa.com/)
   - [ ] [Colorful theme tugboat preview](https://hs-colorful-trvzsw50a6qpe6iqdqqmtdnwil8pnzot.tugboatqa.com/) 

## Notes
- We're installing the latest node package in the lando appserver service and in the tugboat app container since that is what is installed in the [ci build image](https://github.com/pookmish/drupal8ci/blob/3b69c8c521c57c1d1c18f35cc9dd4daa89781af3/Dockerfile#L1) 
- Might be good to test this in an acqua test environment
- If this PR is merged into a release branch, then the release branch's tugboat preview will need to be built with the `Build with no Base Preview` option.
- Developers working with lando will need to update their `.lando.yml` files and rebuild their containers using `lando rebuild -y`

## PR Checklist


.gitignore
```diff
@@ -51,9 +51,15 @@ docroot/modules/stanford
 docroot/themes/stanford
 docroot/profiles/stanford
 
-# Ignore custom theme build artifacts
+# Ignore custom theme build artifacts amd compiled files
 node_modules
 styleguide
+docroot/themes/humsci/humsci_airy/css/humsci_airy.css
```
cienvaras, line 57: Shouldn't we also ignore compiled js here too?

cienvaras (commented):
@hectorlj Left a question about ignored files. Also, tugboat is failing with an error that seems related to this PR changes. Can you please take a look?

.gitignore
hectorlj: Fixed. Added the compiled js directory to `.gitignore` and removed the existing compiled code from the repo

hectorlj:
Tugboat preview just needed to be built with the `Build with no Base Preview` option. Shouldn't be an issue once code is merged into `develop`.

.circleci/config.yml
```diff
@@ -55,6 +55,126 @@ disable_telemetry: &disable_telemetry
   run:
     name: Disable Telemetry
     command: vendor/bin/blt blt:telemetry:disable --no-interaction
+#Jobs
```
cienvaras: @hectorlj SWS just removed this in a [recent PR](https://github.com/SU-HSDO/suhumsci/commit/09f8725349e2e7c596132bf738b9966d19974efe) to use Github Actions instead of CircleCI. Are you bringing it back for a specific reason or is this just a mistake?

cienvaras (changes requested):
Please check comments in the code. We need to make sure that this PR is compatible with the change to Github Actions that SWS made.

.circleci/config.yml
hectorlj: Good catch! Looks like this was a mistake that happened when merging develop into this branch. Not surprising considering how old the branch is. I'll remove it

cienvaras:
@hectorlj I'm getting some errors when I run 

```
 [ExecStack] Running ./hooks/common/theme/build-theme.sh in /app
./hooks/common/theme/build-theme.sh: line 6: lando: command not found
./hooks/common/theme/build-theme.sh: line 7: lando: command not found
 [ExecStack]  Exit code 127  Time 0.006s
 [error]  Executing target-hook frontend-reqs failed.
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
The command failed. This often indicates a problem with your configuration. Review the command output above for more detailed errors, and consider re-running with verbose output for more information.
 [error]  Command `source:build:frontend-reqs --site=hs_traditional` exited with code 1.
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
The command failed. This often indicates a problem with your configuration. Review the command output above for more detailed errors, and consider re-running with verbose output for more information.
 [Exec] Running git checkout /app/config/envs/local/config_ignore.settings.yml
Updated 1 path from the index
 [Exec] Done in 0.02s
 [error]  Command `source:build:frontend --site=hs_traditional` exited with code 1.
For troubleshooting guidance and support, see https://docs.acquia.com/blt/support/ 
```

I think that we should use a different script for `blt` or use an environment variable, because it finds the `.lando.yml` even when it's running inside of docker.

I also added a couple of minor fixes, but other than that looks great.

hectorlj:
@cienvaras Just updated the script to also check that `lando` is a valid command. That should work for the `blt` hook commands as well 🤞🏼

cienvaras (approved):
Thanks Hector! Everything looks good to me, I'll ask SWS to review it.

cienvaras:
@joegl Can you please take a look at this PR? We didn't included it in the sprint release because it included changes in the CI process and we thought it's better to review it on its own. Ideally we would have liked to test it in the Acquia's dev environment, but I know it's being used for something else right now.

We tested everything locally and on tugboat and everything works as expected. If everything looks good to you, it can be merged and tested on stage but if not please let us know to define the next steps. Thanks in advance!

composer.json
```diff
@@ -208,8 +208,19 @@
...
+        ],
+        "test-theme": [
+            "npm ci --prefix docroot/themes/humsci/humsci_basic",
+            "npm run test --prefix docroot/themes/humsci/humsci_basic"
+        ],
         "post-drupal-scaffold-cmd": [
             "patch -p1 < patches/scaffold.patch"
+        ],
+        "post-install-cmd": [
+            "@build-theme"
```
joegl: What's the purpose of compiling theme assets on `composer install`? It's a bit of a hindrance on backend dev workflows.

joegl (commented):
I'm curious what the benefits/goals of this approach are? From what I can tell it's:

1. Removing compiled CSS/JS from being commited to there repo.
2. Compiling the CSS/JS assets on every `composer install` run

This adds additional cost/time to deployment/CI/test/etc workflows (which can cost money based on time spent) in a trade-off for what benefit?

composer.json
cienvaras: Our goal here was to have a fully operational site with all assets build after the install, but you're right, it can impact BE work. I'll remove it from there.

cienvaras:
@joegl Yes, the goal of this PR is to remove the compiled theme assets from the repo and build them in the CI pipeline. The main benefit is to prevent conflicts and errors by removing the manual compilation step and ensuring that the CI will always generate the assets from the most recent source code. It adds some extra time to the CI, but in my experience in other projects it doesn't have a significative impact on the overall times.

joegl:
> @joegl Yes, the goal of this PR is to remove the compiled theme assets from the repo and build them in the CI pipeline. The main benefit is to prevent conflicts and errors by removing the manual compilation step and ensuring that the CI will always generate the assets from the most recent source code. It adds some extra time to the CI, but in my experience in other projects it doesn't have a significative impact on the overall times.

I like the sound of that especially for front-end oriented projects/repo's and I'm wondering how it can best fit into this "mono-repo" we have for the entire stack. I'm going to think about it and let me know if you have any ideas.

composer.json
joegl: Because the compiled CSS/JS files for the theme were removed from the repo and added to the .gitignore, won't these files not exist unless they're manually re-built? Or is there a separate CI step for that?

cienvaras: There's a separate step defined in `blt.yml` that builds the theme.

joegl:
@pookmish Do you have any thoughts/concerns with this new workflow?

pookmish:
> The main benefit is to prevent conflicts

As there are 8 files conflicting in this PR. It's a good example :smile

This is an interesting approach. It's something that we've planned or discussed for various projects in the past, but those never worked because they weren't a mono-repo like this is.

My main concern would be the possibility of unnecessary code, like any `node_modules`, getting included in the artifact. I don't see any changes to the [deploy gitignore](https://github.com/SU-HSDO/suhumsci/blob/develop/blt/deploy/.gitignore).

I also don't see any compiling done for our tests on (now) github actions. Unless that is accomplished with the `frontend-reqs` command established in the blt.yml file.

joegl:
> > The main benefit is to prevent conflicts
> 
> As there are 8 files conflicting in this PR. It's a good example :smile
> 
> This is an interesting approach. It's something that we've planned or discussed for various projects in the past, but those never worked because they weren't a mono-repo like this is.
> 
> My main concern would be the possibility of unnecessary code, like any `node_modules`, getting included in the artifact. I don't see any changes to the [deploy gitignore](https://github.com/SU-HSDO/suhumsci/blob/develop/blt/deploy/.gitignore).
> 
> I also don't see any compiling done for our tests on (now) github actions. Unless that is accomplished with the `frontend-reqs` command established in the blt.yml file.

These were being compiled on `composer install` but I asked 4k to remove that because it seemed burdensome for back-end work to re-compile on every install. Maybe we put that back?

pookmish:
> These were being compiled on `composer install` but I asked 4k to remove that because it seemed burdensome for back-end work to re-compile on every install. Maybe we put that back?

Yeah, I'm not a fan of during the composer install. I agree, it feels burdensome.

cienvaras:
@joegl @pookmish 

> My main concern would be the possibility of unnecessary code, like any node_modules, getting included in the artifact. I don't see any changes to the [deploy gitignore](https://github.com/SU-HSDO/suhumsci/blob/develop/blt/deploy/.gitignore).

That's right, we missed updating that gitignore. It's fixed now.

> I also don't see any compiling done for our tests on (now) github actions. Unless that is accomplished with the frontend-reqs command established in the blt.yml file.

Compiling is being handled by the `frontend-reqs` in `test_existing_site`. `test_new_site` doesn't run that command, but we can add an extra step to compile the assets.

Besides those issues, I think that the main concern is the extra CI time. In Github, we can reduce it caching the compiled assets and recompiling only if there are changes in the source files. This can be achieved using the [Cache action](https://github.com/actions/cache) and the [Paths filter](https://github.com/dorny/paths-filter).

We don't have a solution for Acquia deployments, but as they happen less frequently, we consider that the impact of the extra time is not significant.

Please let us know what you think of the proposed solution for Github Actions. We're are ready to proceed with the implementation and testing, but before we'd like to know if it aligns with your requirements. Thanks!

joegl:
@cienvaras 

> Besides those issues, I think that the main concern is the extra CI time. In Github, we can reduce it caching the compiled assets and recompiling only if there are changes in the source files. This can be achieved using the [Cache action](https://github.com/actions/cache) and the [Paths filter](https://github.com/dorny/paths-filter).

I'm not too concerned about a few additional minutes in the Github Actions CI time, although I agree caching those assets and only re-building when source changes would be ideal.

> Compiling is being handled by the frontend-reqs in test_existing_site. test_new_site doesn't run that command, but we can add an extra step to compile the assets.

Would you be able to document the additional steps now for setting up the repo and fetching updates? I want to make sure I know how and when to compile these correctly, and also update any of our documentation accordingly.

> We don't have a solution for Acquia deployments, but as they happen less frequently, we consider that the impact of the extra time is not significant.

If I understand correctly, the compiling would happen on our local machine when we build the deployment artifact and push it up?

Ultimately, my main concern is what the course of action would be if there was a compiling error, especially on a deploy date. A compiling issue involving these assets isn't currently in my wheelhouse to troubleshoot and I would prefer not to hold up a deploy to resolve compiling issues. Of course, it's definitely possible I am overestimating the possibility of an error occurring.

cienvaras:
@joegl Updated the PR with some final improvements and fixes:

- **Linting**: The initial work had linting implemented, but it got lost when the CI was moved to Github Actions. I added an extra required step in the sites-test workflow to lint the SCSS files on all branches, ensuring that there are no possible errors that can cause a build error. I also updated `stylelint` version and added an updated config file. That caused some conflicts with Code Climate, because the [stylelint versions available in Code Climate](https://docs.codeclimate.com/docs/stylelint) are quite outdated. Because of that, I disabled stylelint in Code Climate. I think that's ok because the theme has its own linter and all the legacy themes are not maintained anymore (and will probably be removed soon), but if you consider that's better to keep it on, let me know to explore options.
- **Cache theme `node_modules`**: theme npm packages are cached in the lint step and then reused in all other steps
- **Ensured that the theme build is executed in all necessary steps**:
    - `test_existing_sites`: the theme is build when `blt drupal:sync` is executed (via the `frontend-reqs` option in `blt.yml`)
    - `text_new_site`: added an extra step to build the theme, because `blt` is not used here
    - `deploy_branch`: the theme is build when `blt deploy` is executed (via the `frontend-reqs` option in `blt.yml`)
    - Tugboat: added an extra step to build the theme

I checked out the Acquia repo and confirmed that the generated `css` and `js` files are present. You mentioned in a previous comment that in some cases you deploy from your local env. In that case, the theme will be automatically build when executing `blt deploy`, so no extra steps are needed.

The theme is build everytime that `blt drupal:sync` is executed, so there shouldn't be any problem when developing locally.

I also updated the theme related scripts and the project README with the necessary information. Please let me know if there's anything missing.

joegl (commented):
Thanks @cienvaras ! I think I'm starting to understand this better.

From what I've gathered from your helpful comment and reviewing, it looks like the assets for all relevant themes get compiled/built using the `npm run build` command in the `humsci_basic` theme directory and there are a number of ways to trigger it. The shell script seems intended to install the npm dependencies and also run it within lando if applicable.

It also looks like if I wanted to rebuild assets after merging in some theme changes without running a `blt drupal:sync`, I could easily run `composer build-theme`. I/We may want to add this to the documentation (I'll also make a note of it internally).

I tested `blt drupal:sync`, `blt deploy`, `npm run theme-build`, `composer build-theme` and `npm run build --prefix docroot/themes/humsci/humsci_basic` locally -- they all appeared to work great on my machine.

Just a few things before merging this into the next release:
- The `11.0.1-release` branch should be merged in first.
- The tugboat build seems to be failing -- is this expected and is there a change required outside of the repo? The PR notes mention something about this.
- I'm not sure if you use `nvm` but we could use it to control the npm version in the repo root. Obviously this wouldn't work for any CI at the moment because the image uses a specific version and doesn't have `nvm` in it (afaik).

cienvaras:
@joegl Thanks for your comments, it seems we're almost there!

The `composer` scripts were added initially to support the automatic compilation when running `composer install`. I was thinking that they can be removed and use `npm` for this, but if you consider them useful we can keep it.

`npm run theme-build` does exactly the same as `composer build-theme`.

> The 11.0.1-release branch should be merged in first.

Alright, that sounds good. I updated the branch and fixed the merge conflicts, I'll try to keep everything updated and working to make things easier when we're ready to merge.

> The tugboat build seems to be failing -- is this expected and is there a change required outside of the repo? The PR notes mention something about this.

That's because Tugboat builds use a base preview by default and node is not installed there. I just triggered a build from scratch, it should work as expected.

> I'm not sure if you use nvm but we could use it to control the npm version in the repo root. Obviously this wouldn't work for any CI at the moment because the image uses a specific version and doesn't have nvm in it (afaik).

Yes, we use it. we had a `.nvmrc` in the theme root, I just updated it to match the node version being used by the CI and added another one in the project's root.

joegl (approved):
Ready to merge this. Thanks @cienvaras

PR #1501, opened 2024-04-10
SHS-5568: Allow editors to select style on Accordion component

hectorlj:
# READY FOR REVIEW

## Summary
Adds `Style` select to allow editors to apply the `hb-accordion-secondary` class to Accordion components

## Steps to Test
1. Create a new flexible page
2. Add two Accordion components
3. On one of the two components set the `Style` select to `Secondary`
4. Save the new page
5. The Accordion component with the `Style` select set to `Secondary` should now be using the `Secondary` style
6. The Accordion component with the `Style` select set to `Default` should use the default style

## PR Checklist


cienvaras:
@hectorlj Can you please update the code so the `hb-accordion-secondary` class is added to the `attributes` variable of the `<details>` element and remove the wrapper `<div>`? I'll need to update the styles anyways because currently the  `hb-accordion-secondary` class needs to be completely outside of the paragraph element. Thanks!

cienvaras:
@hectorlj Thanks! I updated the CSS and it's working great. Just a couple of extra things I missed in my initial review:
- Can you please add an update hook to set all existing accordions style to "Default"
- Can you please confirm that the tests fails are not related with these changes? I don't think so, but better be on the safe side.

hectorlj:
@cienvaras Looks like the menu tests are the culprits, but we're not touching any menu stuff here. Sometimes the tests will pass if rerun so I'm thinking it's a time out issue or something along those lines.

cienvaras (approved):
Looks good, thanks! 🚀

PR #1507, opened 2024-04-17
Four Kitchens Release - Sprint 48

cienvaras:
# READY FOR REVIEW

## Summary

- SHS-5571 -  A11y Level A: Empty headings on Vertical Timeline components
- SHS-5572 -  The arrows on a Spotlight (with no image) are displaying incorrectly on mobile
- SHS-5568 - Allow editors to select style on Accordion component
- SHS-5091 - A11y Level A: Empty heading, on view on Bing Nursery
- SHS-5583 - Empty Headings in Banner/Spotlight Components

## Need Review By (Date)
04/24

## Urgency
medium

### SHS-5571

1. Go to the following pages
    -  https://biology.suhumsci.loc/academics/student-awards
    - https://chemistry.suhumsci.loc/research/centers-affiliates/center-molecular-analysis-and-design/apply
    - https://anthropology.suhumsci.loc/recommended-timeline
    - https://physics.suhumsci.loc/undergraduate/senior-thesis-and-honors

2. Confirm that if there isn't a title on a Vertical Timeline component, there isn't an empty heading
3. Add a title and confirm there's a heading only when there's a title on the Vertical Timeline component

### SHS-5572

1. Create a Flexible Page on Colorful and Traditional and add a component of Spotlight without adding images and add another with images
2. Confirm the position of the arrow in all viewports (specially on mobile) in both components look better 

### SHS-5568

1. Create a new flexible page
2. Add two Accordion components
3. On one of the two components set the `Style` select to `Secondary`
4. Save the new page
5. The Accordion component with the `Style` select set to `Secondary` should now be using the `Secondary` style
6. The Accordion component with the `Style` select set to `Default` should use the default style

### SHS-5091

1. Go to this page https://bingschool.suhumsci.loc/news-events/bing-times
2. Confirm there isn't more than one heading level on view on Bing Nursery 

### SHS-5583

1. Go to these pages

- https://music.suhumsci.loc/events/faculty-ensembles/st-lawrence-string-quartet
- https://music.suhumsci.loc/events/symposium-performing-intermedia
- https://music.suhumsci.loc/pianoseminar2023

2. Confirm that if there's no title, there's no an empty heading level being displayed in the markup
3. Edit each component and a title to confirm it's being displayed as expected

## PR Checklist


docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.install
```diff
@@ -217,3 +217,49 @@ function hs_paragraph_types_update_10004(&$sandbox) {
   }
   $sandbox['#finished'] = count($sandbox['ids']) ? 1 - count($sandbox['ids']) / $sandbox['total'] : 1;
 }
+
+/**
+ * Set default field_accordion_style value for existing content.
+ */
+function hs_paragraph_types_update_10005(&$sandbox) {
```
pookmish: Why is the update necessary? In the preprocess you are checking for `secondary`. Seems to me that `default` and `null` will behave the same.

pookmish: If you are trying to prevent an existing paragraph being edited and having an "error", you should make the field not required and remove the `default` option.

cienvaras: @pookmish We just want to make things easier for users. If an user edits a page with accordions, they'll have to manually set the field in all existing accordions to "Default" before saving, because it's a required field.

pookmish: ok, yeah, see my comment just before yours.

cienvaras: @pookmish Thanks! Yeah, it makes sense to remove the option. I updated the config and removed the update hook. 

The only detail is that we need to keep the "Default" label for usability purposes. I added a `hook_field_widget_complete_WIDGET_TYPE_form_alter()` to change the default "- None -" label to "Default". Please take a look and let me know if that works.

pookmish: Yeah that looks good. less impact on the code deployment as well. I wish there was a UI option to change the `_none` label.
The only possible thing that you could do is use `$context` [in this part](https://github.com/SU-HSDO/suhumsci/pull/1507/files#diff-c0831e03dcd9e444fe998d19c7d63430991498d1eba660f25f1ad6a6116196b5R147) in the hook to determine if it's the expected field. But it works the same way.

pookmish (commented):
I approve pending a review by @joegl

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
```diff
@@ -138,3 +138,13 @@ function hs_paragraph_types_field_widget_complete_paragraphs_browser_form_alter(
     $field_widget_form['widget'][$key] = $element;
   }
 }
+
+/**
+ * Implements hook_field_widget_complete_WIDGET_TYPE_form_alter().
+ */
+function hs_paragraph_types_field_widget_complete_options_select_form_alter(array &$field_widget_form, FormStateInterface &$form_state, array $context) {
+  // Change the default option label for field_accordion_style.
+  if (!empty($field_widget_form['widget']['#field_name']) && $field_widget_form['widget']['#field_name'] == 'field_accordion_style') {
```
joegl: Formatting.

```suggestion
  if (
    !empty($field_widget_form['widget']['#field_name']) && 
    $field_widget_form['widget']['#field_name'] == 'field_accordion_style'
  ) { 
```

docroot/modules/humsci/hs_layouts/hs_layouts.module
```diff
@@ -111,12 +111,23 @@ function hs_layouts_preprocess_pattern_spotlight(&$variables) {
...
     !$paragraph->field_hs_accord_expcol->isEmpty() &&
     $paragraph->get('field_hs_accord_expcol')->getString()
   ) {
     $variables['attributes']->addClass('hb-accordion_toggle-all');
   }
+
+  if (
+    $paragraph->hasField('field_accordion_style') &&
+    !$paragraph->field_accordion_style->isEmpty() &&
+    $paragraph->get('field_accordion_style')->getString() == 'secondary'
```
joegl: It feels like there should be a way to achieve this with less than 3 conditions, but I can't think of one. Works for me.

joegl (changes requested):
This LGTM just a small formatting change.

docroot/modules/humsci/hs_layouts/hs_layouts.module
pookmish: I believe the 2nd condition can be excluded because if it's empty `$paragraph->get('field_accordion_style')->getString()` should return `NULL` I believe.

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
pookmish: If making this change, I would suggest to use `$context` instead of the `$field_widget_form`. I believe the field storage definition should be part of that array, so you can easily check the field name method.

cienvaras: @pookmish @joegl Changed the condition to use `$context` instead. Didn't split it into multiple lines because the new condition is shorter.

cienvaras: btw, the field definition is not in the `$context` array, just the field items object.

docroot/modules/humsci/hs_layouts/hs_layouts.module
cienvaras: @pookmish @joegl Yes, `getString()` doesn't return `NULL` but an empty string, which works the same in this case. I removed the `isEmpty()` condition.

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
pookmish: ahh, yeah from the items, you can get the field definition. but the way you did works too. :+1:

joegl (approved):
LGTM.

PR #1542, opened 2024-05-30
SHS-5641: Site editors shouldn't remove imported photos

codechefmarc:
# READY FOR REVIEW

## Summary
Removes the ability for non-administrators from deleting media that is attached to nodes that have been imported via a migration
* Removes the "Delete Media item" contextual menu item in the node edit form
* Removes the "Remove" button on the media edit form
* Only applies to non-adminstrators
* Only applies to media that was part of a migration
* Note: Work was completed in a forked version of the `stanford_migrate` module. Forked version is located here: https://github.com/fourkitchens/stanford_migrate
* Note: There is a [PR over on the forked repo](https://github.com/fourkitchens/stanford_migrate/pull/2). Once that is merged in to the `8.x` branch over there, the `composer.json` file here will need to be modified to point to the main branch over there.

## Need Review By (Date)
2024-05-31

## Urgency
Medium

## Steps to Test
1. Login to a site as a non-admin
2. Make sure the site contains migrated nodes with media entities attached. Development and testing was done with the Physics site.
3. Edit a node that has been imported via a migration. Content types used for development and testing are Person and Event. You can tell if the node was part of a migration because there are some fields that are disabled and have an orange background:

Screenshot 2024-05-30 at 3 55 31 PM

4. Click on the contextual menu for the Image or Profile Image field
5. Verify that the "Delete Media item" is no longer available

Screenshot 2024-05-30 at 3 56 37 PM

6. Click "Edit media item"
7. Verify that on the media edit form, there is new messaging at the top informing that the media can not be deleted because it was part of a migration:

Screenshot 2024-05-30 at 3 57 46 PM

8. Verify that the "Remove" button is not available next to the media item name:

Screenshot 2024-05-30 at 3 58 26 PM

9. Verify that other fields for the image can be edited, such as Focal Point, alt text, and caption.
10. Optionally, log in to the site as an admin and verify that admins can delete media from the contextual menu and from the media edit form.

## PR Checklist

---
  - https://app.asana.com/0/0/1207858685626376


codechefmarc:
@joegl Just checking in on this one. Thanks!

joegl:
@codechefmarc 

I merged the PR in the fork https://github.com/fourkitchens/stanford_migrate/pull/2.
I also updated the PR here to the latest code.

I am not exactly sure how we want to pull this in now before merging. Options are:
1. Change the composer constraint for `fourkitchens/stanford_migrate` to `8.x-dev` and pull from the 8.x branch
2. Cut a release for `fourkitchens/stanford_migrate`, add the repo to packagist, and then change the composer constraint to the release version E.g., `^8.0`

We can do the first option for now but I think in the long run we'll want to do the second (so may as well do it now?).

cienvaras:
@joegl I agree that doing a release is probably the best way to go, but I was wondering: wouldn't it be better to merge the changes into the original https://github.com/SU-SWS/stanford_migrate repo? I understand if you want to keep things separate, but I prefer to ask

joegl:
> @joegl I agree that doing a release is probably the best way to go, but I was wondering: wouldn't it be better to merge the changes into the original https://github.com/SU-SWS/stanford_migrate repo? I understand if you want to keep things separate, but I prefer to ask

The changes right now are too specific to the H&S stack. We'd need to revise them before merging.

cienvaras:
@joegl 

> The changes right now are too specific to the H&S stack. We'd need to revise them before merging.

Ok, understood. It makes a lot of sense then to have a separate repo/package for this. We only have two questions/concerns:

1. We forked the repo into the fourkitchens org temporarily for dev purposes, but we think that it would be better to move it to the `SU-HSDO` now that we'll keep it and use in production. What do you think?
2. Do you have an idea on how to handle updates? Maintaining a separate repo will definitely add some overhead, but we'd like to figure out the best way to handle it and reduce impact on both teams, while keeping the code as up to date as possible.

(Adding @ahughes3 for visibility)

joegl:
@cienvaras I don't think we need this code to live in the `stanford_migrate` module and was thinking we could build it into [the existing `hs_migrate` module ](https://github.com/SU-HSDO/suhumsci/tree/develop/docroot/modules/humsci/hs_migrate) in the `suhumsci` repo, add `stanford_migrate` as a dependency, and move this code in there. Then any custom code for H&S migration features could also be added there in the future. We'd also then close out this fork.

Overall conceptually forking sounded like a good idea and I'm glad we tried it out, but as we worked through it it's brought on a few more complications including the additional maintenance overhead. Let me know what you think 👍

cienvaras:
@joegl Thanks for the quick response! Yes, I like the idea of adding the new functionality in `hs_migrate`, that'll definitively save us a lot of work versus maintaining a fork. @codechefmarc will proceed with the implementation.

codechefmarc:
This PR and brach are now outdated and will be replaced with a new PR/branch. The plan is to use the code that was added to `stanford_migrate` in `hs_migrate` instead - therefore not relying on changes in the `su-sws/stanford_migrate` repo and keeping all changes in `suhumsci`. I ran into issues refactoring on this branch, so I will create a new one soon and delete this one.

codechefmarc:
This PR has been replaced with a newer one - https://github.com/SU-HSDO/suhumsci/pull/1609

PR #1554, opened 2024-06-12
SHS-5639 issues with toolbar not behaving correctly after logging out then logging back in

dalin-:
# READY FOR REVIEW

## Summary
On logout, custom code was previously cleaning up `localStorage`.  However Toolbar also stores data in `sessionStorage`.  To avoid conflicts between the two we need to clear out both.  

This PR also fixes some pre-existing JS errors, both when logged in and logged out.  I first thought that these might be the cause of the problem.  They weren't, but I fixed them anyway.  @cienvaras you wrote the original code.  Can you check this to see if I did that right?  It wasn't totally clear to me what the code that I changed is trying to do. 

## Need Review By (Date)
2024-06-14

## Urgency
medium

## Steps to Test
1. In a fresh session, log in to a site.
2. Note that you see the "Shortcuts" toolbar active.
3. Log out.
4. Immediately log back in
5. Note that the "Shortcuts" toolbar is again active (before this PR you'd be in a strange state where multiple toolbars were active)

## PR Checklist


cienvaras (approved):
@dalin- Thanks, looks good!

I think that the JS errors you fixed were initially caused by an update in the editoria11y module that changed its internal structure in a way that doesn't match our custom script. Your fixes prevent the JS errors, which is good for now, but it would be a good idea to create a ticket to review the changes and determine if our custom script is still needed or need customization.

PR #1569, opened 2024-06-28
SHS-5513: Colorband link text character limit

mariannuar:
# READY FOR REVIEW

## Summary
In the `Colorband` component, update help text for the Link field:
`Both URL and Link text are required when adding a link. When displayed, the Link text is trimmed to 80 characters.`
And as the user typing, they see indication of how many characters they can use (with a "soft" limit)

## Need Review By (Date)
07/01

## Urgency
medium

## Steps to Test
1. On Colorful and Traditional, import configuration
2. Add a `Colorband` component
3. Confirm the help text for the Link field is `Both URL and Link text are required when adding a link. When displayed, the Link text is trimmed to 80 characters.` and that as you're typing, you can see indication of how many characters you can use

## PR Checklist

---
  - https://app.asana.com/0/0/1207719828600729

### DIFF
--- a/config/default/core.entity_form_display.paragraph.hs_clr_bnd.default.yml
+++ b/config/default/core.entity_form_display.paragraph.hs_clr_bnd.default.yml
@@ -30,7 +30,11 @@ content:
     settings:
       placeholder_url: ''
       placeholder_title: ''
-    third_party_settings: {  }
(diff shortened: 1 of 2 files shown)


cienvaras (commented):
Works as expected, thanks!

PR #1575, opened 2024-07-09
SHS-5683: Blockquote implementation

mariannuar:
# READY FOR REVIEW

## Summary
Update blockquotes in wysiwygs

## Need Review By (Date)
07/12

## Urgency
medium

## Steps to Test
1. On Colorful and Traditional
2. Create a Flexible Page
3. Add a `Text Area` component and add a `Blockquote`
4. Check it does look according to what's in Option A in [Figma](https://www.figma.com/design/hOtD5MStjFIavUrWapHKHN/UX%2FDesign-Tickets-Work?node-id=1384-56&t=YEcvFLF7ZsYur3nv-4)
5. Add a `Testimonial` component too to confirm it still look the same
6. Please test in all the color variants in both themes
7. Also check the blockquote block in footer (normal and dark) looks good
8. Add blockquotes in the following pages to confirm they look good:

- https://hs-traditional.suhumsci.loc/news/jackelyn-hwang-wins-jane-addams-award-best-article-community-and-urban-sociology-section
- https://hs-traditional.suhumsci.loc/people/abhinanda-sarkar
- https://hs-traditional.suhumsci.loc/publications/privilege-and-punishment-how-race-and-class-matter-criminal-court
- https://hs-traditional.suhumsci.loc/components/text-area-typography

9. Check VRT https://percy.io/da667d23/suhumsci/builds/35211009 for more testing

## PR Checklist

---
  - https://app.asana.com/0/0/1207762326049451
  - https://app.asana.com/0/0/1207792433158565


docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
```diff
@@ -264,15 +264,24 @@ $px-only: true;
   blockquote:nth-child(n) {
     @include hb-blockquote;
     overflow: visible;
-
-    font-size: 24px;
+    font-size: 24px !important;
+    line-height:35.56px !important;
```
cienvaras: Always try to use relative dimensions (`LINE_HEIGHT_PX/FONT_SIZE`) for `line-height`, to allow it to automatically adjust to font size. In this case `35.56/28 = 1.27`

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
```diff
@@ -249,29 +249,58 @@
...
   }
 
-  // Don't apply these rules to blockquotes in testimonials, the structure is not the same and need
-  // a different implementation.
   &:not(.hb-testimonial__quote) {
-    @include hb-quote-wrapper;
+    position: relative;
+    color: var(--palette--primary);
+    font-style: normal;
+    line-height: hb-calculate-rems(35.56px);
```
cienvaras: Use relative value for `line-height` (see comment above).

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
```diff
@@ -249,29 +249,58 @@
...
+    &::after {
+      content: ' ';
+      position: absolute;
+      background-repeat: no-repeat;
+    }
+
+    &::before {
+      @include hb-icon-background('quote-mark', 'tertiary');
+      
+      width: hb-calculate-rems(68px);
```
cienvaras: In the figma designs, the quote icons are 36x31px, but the implemented version is a bit larger. It seems that the svgs have some whitespace around the quotes and that makes harder to match the exact size, but try to get a bit more closer to the expected size.

docroot/themes/humsci/humsci_basic/src/scss/tools/_mixins.text.scss
```diff
@@ -249,29 +249,58 @@
...
+      
+      .hb-dark-pattern.hb-local-footer & {
+        @include hb-icon-background('quote-mark', 'tertiary');
+      }
+    }
+
+    &::after {
+      @include hb-icon-background('close-quote-mark', 'tertiary');
+
+      width: hb-calculate-rems(40px);
```
cienvaras: Same as above, please check dimensions and try to match the size in the designs.

docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss
```diff
@@ -264,15 +264,24 @@ $px-only: true;
   blockquote:nth-child(n) {
     @include hb-blockquote;
     overflow: visible;
-
-    font-size: 24px;
+    font-size: 24px !important;
```
cienvaras: I think some of these `!important` can be removed. For example, the only reason the the `!important` for setting the font size to `28px` in larger screens is because this line also has `!important`. We should try to avoid using `!important` as much as possible.

cienvaras (changes requested):
@mariannuar Looks great overall. Just some minor fixes needed, let me know if you have any questions. Please notice I updated the branch with the latest code in `fk-stndfd-sprint-54`.

mariannuar:
@cienvaras Thank you! This is ready for review again 🙂

cienvaras (approved):
@mariannuar Thanks for the changes, everything look better now, specially the quote icons.

I added some extra changes:
- Moved some of the overrides to the testimonial component styles, to prevent doing extra overrides in the mixin. This reduces the code size a bit and prevents overriding some basic styles for all the blockquotes.
- Removed the `&:not(.hb-testimonial__quote)` styles from `import.scss` and added back the `!important`. While your solution works, it adds an extra class to the CKEditor styles that's not relevant here. If the class changes in the future or if we need to override a new class, then we would need to do extra changes here. We need to ensure that all blockquotes in CKEditor have  the same styles, and `!important` is the best approach for that in this case. At least you were able to remove the font related styles because they were not needed anymore.

Let me know if you have any question about my changes!

PR #1579, opened 2024-07-16
SHS-5675: Helping users avoid 'Table cell missing context (e.g. headings) A11y warning

codechefmarc:
# READY FOR REVIEW

## Summary
- When adding a table to a WYSIWYG field with CKeditor, table headings are now added by default.
- This can be turned off using the table function built into CKEditor, if needed.
- Prevents accessibility tools from flagging poor accessibility due to no table header.

## Need Review By (Date)
7/22/2024

## Urgency
low

## Steps to Test
1. Login to the site and add a flexible page
2. In the Text Area field, add a table using the table tools

Screenshot 2024-07-16 at 5 02 19 PM

3. When the table is created, verify that there is a header row, marked with a dark red background.

Screenshot 2024-07-16 at 5 02 32 PM

4. Enter some data into the table, and save the flexible page.
5. On the front-end, verify that the table header is styled appropriately.

Screenshot 2024-07-16 at 5 04 39 PM

@cienvaras - The table is styled correctly when saved on the front-end. However, on the back-end when editing the table, the header text color is too dark against the dark red background so it is hard to see:

Screenshot 2024-07-16 at 4 21 15 PM

I know you mentioned to me where the CSS is handled but I've forgotten. I'm happy to fix this if you'd like or if you or Mari could take it on and let me know where the change goes? I would like to know for future tickets. Thank you!

## PR Checklist

---
  - https://app.asana.com/0/0/1207835764616004


cienvaras:
@codechefmarc For updating the wysiwyg styles, check `docroot/themes/humsci/humsci_basic/src/scss/ckeditor/imports.scss`. This is the entry point for the WYSIWYG styles. Probably the white color is being overridden with a gin definition with a higher specificity, so just need to override it again manually in this file.

After making the change, you can build the assets running `npm run theme-build` or start watch mode with `npm run theme-watch` in the project's root.

Leave me a comment here if you have any issue or if you run out of time, so I can do the fix after the holiday. Also, please notice that I updated the branch and changed the base, so you'll probably need to re-import config or re-install.

codechefmarc:
@cienvaras - Thank you! I added the CSS for that for CKEditor and that works. This is ready for review.

cienvaras (approved):
@codechefmarc Thanks for adding the CSS, works great! I just did a quick change to use the FE theme variable (`--palette--white`) instead of the gin variable, to prevent any future issues in case of a change in the value.

PR #1686, opened 2024-11-19
HSD8-000: Add phpcs (Drupal PHP CodeSniffer) to CI

joegl:
# READY FOR REVIEW

## Summary
_[briefly summarize the changes here]_

## Need Review By (Date)
_['10/30', 'asap', etc.]_

## Urgency

## Steps to Test
1. _[First testing step]_
2. ...

## PR Checklist


phpcs.dist.xml
dalin-, line 1: While PHP CS will pick this up, I think it would be better to name as `phpcs.xml`.

phpcs.dist.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<ruleset name="Project Standard" namespace="Advomatic\project\CS\Standard">
```
dalin-: Ooops.  The namespace might be better as `Stanford\project\CS\Standard`

dalin- (commented):
Some suggestions.

phpcs.dist.xml
joegl: Thanks for the catch!

joegl:
> Some suggestions.

Thanks @dalin- 👍

joegl:
I'm going to merge this even though it's failing. We can clean-up the code as needed and time permitting and this won't be a required check for merging, but it will be nice to have.

PR #1879, opened 2025-06-23
SHS-6289: Bug | Lock ui_patterns to 1.10 to fix views and utility classes

cienvaras:
# Summary
- Lock version of UI Patterns module to `1.10` (it was updated to `1.13` in latest update) to prevent issues in views with patterns with custom classes and rewrites. The update also prevents `patches/contrib/ui_patterns_views-preview.patch` from being applied (which may be the cause of the issue instead of the update itself).

**This PR is more a proof of concept than a definitive fix. Locking the version of UI Patterns might not be the best approach for fixing this issue.

## Backstory
- [ClickUp Ticket](https://app.clickup.com/t/36718269/SHS-6289)

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. Visit https://hs-colorful-bv6ubbpuokahx7qfthr0zihbqcukhcjw.tugboatqa.com/default-views/events and confirm that elements in the cards styled by utility classes (categories, dates, etc.) look as expected (You can compare it with the [test site from the 11.11.2 release](https://hs-colorful-a9vadk3m4hvphxicmu8suqaazlnvoilf.tugboatqa.com/default-views/events), which has the bug).

# Review Tasks

### Code
- [ ] Are tests provided?

### Code security

### General

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -165,7 +165,7 @@
         "drupal/term_condition": "^2.0",
         "drupal/token_or": "^2.2",
         "drupal/transliterate_filenames": "^2.0",
-        "drupal/ui_patterns": "^1.2",
+        "drupal/ui_patterns": "1.10",
         "drupal/ui_patterns_field_variants": "1.x-dev",
         "drupal/video_embed_field": "^2.4",
         "drupal/view_unpublished": "^1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "dd91e7ace76191223df386b63618b0b1",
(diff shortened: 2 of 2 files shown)


dalin-:
Also (this is probably a Q for SWS), while HSDP is still on Composer Patches 1.x, add this
```
  "extra": {
    "composer-exit-on-patch-failure": true
  }
```
to help prevent these kinds of problems. 

In 2.x this is the default (and only) behaviour.

ahughes3 (approved):
updates look good!

joegl:
@cienvaras Thanks for locking the module in and resolving this for the release tomorrow.

@dalin- Thanks for the suggestion, I'll take a look at this in the future.

@ahughes3 Can you create a ticket for me to investigate this a bit more in-depth next week? Thanks!

I will proceed with merging for the release tomorrow as this solves the issue. Then I'll look at how we can upgrade `ui_patterns` and resolve the patch conflict.

PR #1901, opened 2025-07-17
SHS-6321: Auto-connect SUNet IDs to Imported Person nodes

nJim:
# Summary
- Creates an event subscriber (`UserLoginEventSubscriber`) that fires when a user logs in
- Implements a service (`PersonAuthorship`) that handles content authorship assignment logic
- Supports both SSO and native Drupal user authentication methods
- Automatically assigns 'author' role to users who don't have appropriate content editing permissions
- Provides user-friendly notifications with direct links to assigned content

## Backstory
- When a user logs into a site, they should be able to edit their own user profile. This is currently a chore for site managers who are coordinating access between a growing list of person nodes. This is part of an initiative with Jasper Ridge Biological Preserve.

## Need Review By (Date)
Not set.

## Urgency
Not set.

## Steps to Test
1. Setup the test
  a. [Login as an admin user](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/user)
  b. Navigate to the [site options config interface](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/admin/config/site-options) and make sure the 'Auto-connect SUNet IDs to Imported Person nodes' setting is enabled.
  c. Verify an [account exists for testing an SSO managed user login](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/user/vnsteere). 
  d. [Modify the Person content type form to place the 'SUNet ID' field on the form](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/admin/structure/types/manage/hs_person/form-display). Set the field to use a text widget. This is not a task needed for most sites, we but we need it for testing. Some sites may choose to include this field. Others may leave it as a hidden field that is populated only by the importer. 
  e. [Edit a person to be tied to the SSO login managed account](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/node/12426/edit). Edit this profile and set the `SUNet ID` to 'vnsteere' to match the user account. Save the node. ** OR use your own account to test the actual SSO system. ** **Set the account to only have the authenticated user role.**

2. Test Drupal managed account
  a. Use the Tugboat terminal to get a login link for the test user: `drush @hs_colorful.local uli --name=vnsteere` or visit the account page to [masquerade as the user](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/user/vnsteere).
  b. Visit this link to trigger a user login.
  c. You should see a message similar to: `A matching person profile was found for Testing a profile for Drupal Login account. You now have permission to edit it.`
  d. Click on the link within this message to see the matching person.
  e. Verify that [the user can edit this person node](https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com/node/12426/edit). Their account is set as the author.

3. Additional variations to test
  a. **Test with user having existing roles**: Create a user with 'contributor' role and verify they don't get 'author' role added
  c. **Test nodes with existing authors**: Create person nodes authored by a different user. Verify that the node is reassigned when the linked account signs in.
  d. **Test with invalid email formats**: Verify the system handles malformed email addresses gracefully
  e. **Test with multiple matching nodes**: Create multiple anonymous-authored person nodes with the same email - verify all are assigned
  f. **Test role assignment logic**: Verify users without 'author', 'contributor', or 'site_manager' roles get 'author' role added
  g. **Test with a developer or site manager**: This auto assignment does not take place for users with the role of developer or site manager. This is to avoid running this code for power users that access the site regularlly and already have advanced content access permissions. 
  h. **Test disabling this feature**: Turn off this feature in the site options config form. Verify that accounts and person content no longer automatically connects when this feature is disabled.

Screenshot 2025-07-17 at 2 39 45 PM

# Review Tasks

### Code
- [ ] Are tests provided?

### Code security

### General

---
  - https://app.asana.com/0/0/1211009486338170
  - https://app.asana.com/0/0/1209646312520660


ahughes3:

Links:
- https://pr1901-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://hs-colorful-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://hs-traditional-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://english-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://history-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://west-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6879203f131084d09c53d9a4

Visual Diffs:
- https://dashboard.tugboatqa.com/6879203f131084d09c53d9a4/visualdiffs

ahughes3:

Links:
- https://pr1901-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://hs-colorful-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://hs-traditional-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://english-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://history-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com
- https://west-qg0afibimnwxiylrx867bhusl0ii34eg.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6879203f131084d09c53d9a4

Visual Diffs:
- https://dashboard.tugboatqa.com/6879203f131084d09c53d9a4/visualdiffs

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

cienvaras (approved):
@nJim LGTM, thanks!

@ahughes3 Ready for you. Not sure how's review is going to work on your side for this PR, so please re-assign as needed.

cienvaras:
@ahughes3  @nJim Sorry, I missed a detail before. When I was testing, I noticed that the SUNet ID field added by SWS in a different PR was not present in the person content type form. I had to add it for manually setting a value while testing. In production, this field will be populated when importing people, so I'm not sure if this is intentional or an omission.

If the field should be visible in the form, we need to add it in an update hook. This is something that can be done in another PR, not necessarily a blocker for merging this PR.

nJim:
@cienvaras I believe this work is currently acceptable:
- Albert has tested this with a real migration and confirmed that the SUNet IDs are coming through
- The team plans to only show this field on a per-site basis so they will add it to JRBP manually
[Conversation for context](https://fourkitchens.slack.com/archives/C03AST832QJ/p1753199281933579?thread_ts=1752851672.410459&cid=C03AST832QJ)

cienvaras:
@nJim Thanks for the context! In that case, I think we're good to go @ahughes3

ahughes3 (approved):
looks good!

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+   */
+  protected function getUserEmail(AccountInterface $account): string {
+    // First, check if this is an SSO user by looking up authmap.
+    $authname = $this->database->select('authmap', 'a')
+      ->fields('a', ['authname'])
+      ->condition('uid', $account->id())
+      ->execute()
+      ->fetchField();
+
+    if ($authname && $this->isValidAuthname($authname)) {
```
joegl: This repeats logic from getUserAuthname(). If getUserEmail() requres getUserAuthname() maybe chain them somehow, or get the authname first and pass it into getUserEmail().

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+  /**
+   * Get the authname (SUNet ID) for a user account.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account.
+   *
+   * @return string|null
+   *   The authname or null if none found.
+   */
+  protected function getUserAuthname(AccountInterface $account) {
```
joegl: I understand this should exist as a service in `stanford_samlauth` but doesn't. Can you add a @todo in the comment to update the code to use a service from `stanford_samlauth` instead if it gets added? Thanks

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    // If no SSO authname, try to get the user's email address directly.
+    $email = $account->getEmail();
+    if (!empty($email)) {
+      return $email;
+    }
+
+    // No email found.
+    return '';
```
joegl: I noticed this returns an empty string while getUserAuthname returns NULL. These should probably be identical returns.

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+   *   The email address used for matching.
+   * @param string|null $authname
+   *   The user's authname (SUNet ID) used for matching.
+   */
+  protected function notifyUser(array $nodes, string $user_email, ?string $authname = NULL): void {
+    foreach ($nodes as $node) {
+      if ($this->isNodeMatch($node, $user_email, $authname)) {
+        $this->addNodeNotification($node);
+      }
+    }
```
joegl: If multiple nodes are assigned to a user, could it create one notification indicating all the assignments instead of one for each node?

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+  protected function assignNodesToUser(array $nodes, AccountInterface $account, string $user_email, ?string $authname = NULL): int {
+    $updated_count = 0;
+
+    foreach ($nodes as $node) {
+      if ($this->isNodeMatch($node, $user_email, $authname)) {
+        $this->ensureUserHasRequiredRole($account);
+        $this->assignNodeToUser($node, $account);
+        $updated_count++;
+      }
+    }
```
joegl: Could we impose a limit here and then log a message? Maybe something like 10 nodes? This could help mitigate issues if a user were to somehow errantly be assigned hundreds of nodes. Obviously set 10 to whatever a reasonable number of nodes would be.

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * Delegates to the PersonAuthorship service to handle the business logic
+   * of assigning person node ownership to users.
+   *
+   * @param \Drupal\user_event_dispatcher\Event\User\UserLoginEvent $event
+   *   The user login event.
+   */
+  public function onUserLogin(UserLoginEvent $event) {
+    $account = $event->getAccount();
+    $this->personAuthorship->processPersonAuthorship($account);
```
joegl: This should have a try/catch to catch any exceptions. Don't want to break the login process if this is falling over.

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,443 @@
...
+   * Check if processing should be skipped for this account.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account.
+   *
+   * @return bool
+   *   TRUE if processing should be skipped, FALSE otherwise.
+   */
+  protected function shouldSkipProcessing(AccountInterface $account): bool {
+    return $account->isAnonymous();
```
joegl: Could we exclude this lookup on roles that already have permission to edit person nodes? At the very least exclude administrator/developer and site manager roles. We don't need to process this every time an admin/manager logs in to the site.

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * React to user login events.
+   *
+   * Delegates to the PersonAuthorship service to handle the business logic
+   * of assigning person node ownership to users.
+   *
+   * @param \Drupal\user_event_dispatcher\Event\User\UserLoginEvent $event
+   *   The user login event.
+   */
+  public function onUserLogin(UserLoginEvent $event) {
```
joegl, line 72: Does this have to be an onUserLogin process? Some sites have thousands of nodes and this could be an expensive query to run on every login. Perhaps this could create a queue that runs on every cron.

joegl (changes requested):
Left comments, feedback and suggestions. I also have three general concerns:

1. There should probably be a clean-up process of some kind that runs during cron and removes ownership if things change. 
2. Could this be a batch/cron process of some kind and not trigger on every login?
3. Since this is put directly in the hs_person module all sites will bu default get this. Can we either put this behind a feature flag or into its own module? There should be a way to turn it off if necessary.

dalin-:
@joegl 
Jim is out this week, but I can answer some of the more product-related questions:

> There should probably be a clean-up process of some kind that runs during cron and removes ownership if things change.

By "if things change", do you mean if a user account is deleted we need to figure out what to do with the node?  I think we should rely on the platform-wide approach to this rather than do a one-off here.  We need the same solution for that regardless of whether authorship was set automatically, or by hand (which is how Jasper Ridge are doing it now). 

> Could this be a batch/cron process of some kind and not trigger on every login?

I don't think so.  A typical scenario is Jasper Ridge telling someone: "Go to the website, log in, and update your profile".  There's nothing for cron to link to until they log in and SSO creates an account for them on the fly. 

But I think your idea of limiting to 10 nodes is a reasonable backstop just in case something insane happens. 

> Since this is put directly in the hs_person module all sites will bu default get this. Can we either put this behind a feature flag or into its own module? There should be a way to turn it off if necessary.

@ahughes3 What are your thoughts here?  Do you have any concerns about this running on other sites?  If so, what do you think about H&S funding that rather than Jasper Ridge?

ahughes3:
> @ahughes3 What are your thoughts here? Do you have any concerns about this running on other sites? If so, what do you think about H&S funding that rather than Jasper Ridge?

We discussed offline - Yes, there are concerns about having the feature enabled on other sites by default. We decided that we want to move forward with a site setting to enable or disable this feature as an HSDP request.

dalin-:
@joegl 

> We decided that we want to move forward with a site setting to enable or disable this feature as an HSDP request.

We're doing this as a separate ticket in the current sprint.  We'll keep this PR open until that's in place.

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
nJim: Added. I don't have any specific throwables begging to be added to the services, but we will make sure things break gracefully and log the event.

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
nJim: Sounds good. Since notifyUser() is just a thin wrapper that filters nodes and then calls addConsolidatedNotification(), I can merge them into a single method.

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
nJim: I don't know enough about the business requirements to weigh-in here, but I will escalate this to the folks who do. 

One thing we are doing to mitigate some risk is to put this feature behind a site setting. So only sites opting into this functionality will have work done with this event.

My gut says that they will need somethings this comprehensive to catch any content that may have changed (and updated email address) or content that may have manual interventions (person created outside of the importers). It is likely that they will want to check all of the people nodes which would negate the purpose of the queue worker.

I will propose two small changes if they make sense:
- We only query recently created/change content on login to minimize the processing time
- We add an admin button to re-run this process for all users in a batch job (to get the previously created content).

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
nJim: Yes. This is why I tend make methods for most things. I've change the logic to:
```
    $skip_roles = ['anonymous', 'site_manager', 'administrator'];
    return !empty(array_intersect($skip_roles, $account->getRoles()));
```

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
nJim: Sounds good. I've added this logic to the findMatchingPersonNodes method so we won't ever load more than 10 entities. The limit is a parameter that defaults to 10 in case we want to do larger batches outside of this login even. The logger records a message if more than 10 were found.

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
ahughes3: @nJim @joegl providing some additional business login/requirements context and feedback

- Since this is specific to Person nodes is there a way to only query Person nodes? I don't think there is a site or would be a site that has 1000s of Person nodes.
- Once a user is connected or has author role does this still need to run? 
- I don't think there is a situation where a user would have more than 1 Person node assigned to them.
- Users can only login with SSO, there's no Drupal login option
- "to catch any content that may have changed (and updated email address) or content that may have manual interventions (person created outside of the importers)." - If the connection is based on SUNet why would the email field impact this? Imported content will have a SUNet email address that can't be changed. 
- If a person node is created manually and we want the functionality would the Site Editor manually add the SUNet ID to the Person node field?

joegl: @nJim @ahughes3 

> We only query recently created/change content on login to minimize the processing time

This could help a little, but I don't think it's necessarily worth implementing.

> We add an admin button to re-run this process for all users in a batch job (to get the previously created content).

I do like this idea, but it could go into a follow-up implementation.

> Since this is specific to Person nodes is there a way to only query Person nodes? I don't think there is a site or would be a site that has 1000s of Person nodes.

It already does this 👍 

> Once a user is connected or has author role does this still need to run?

If the person nodes change, then it would still need to run. For example, new persons gets imported, new persons get manually created, etc., I am thinking of this functionality as it would work platform wide with the CAP importer on every site, but I believe this is mainly getting built for a single site which will do one person import? Because that would change things.

> I don't think there is a situation where a user would have more than 1 Person node assigned to them.

I tend to agree, although there are a couple scenarios, whether it's person nodes getting imported from two sources or the manual creation of a person node that results in duplicates with the same SUNET. Would we want to assign them all? Or would that be considered a content governance issue and we should only be assigning a single node? 

> If a person node is created manually and we want the functionality would the Site Editor manually add the SUNet ID to the Person node field?

They would have to yes, but I believe this field is hidden because it was only intended to be used by importers.

> "to catch any content that may have changed (and updated email address) or content that may have manual interventions (person created outside of the importers)." - If the connection is based on SUNet why would the email field impact this? Imported content will have a SUNet email address that can't be changed.

This would affect some of the answers to the above and is also something I had a question about. The logic here checks for a match on both `field_hs_person_sunet_id` and `field_hs_person_email`. From what I know the intention was to only match SUNets and not emails. The email field is also user editable.

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+    // Check if auto-connect persons is enabled (/admin/config/site-options).
+    $config_pages = \Drupal::service('config_pages.loader');
```
joegl: Instead of a service call this should be injected into the event subscriber.

docroot/modules/humsci/hs_person/src/Service/PersonAuthorship.php
```diff
@@ -0,0 +1,466 @@
...
+   * @return string|null
+   *   The authname or null if none found.
+   */
+  protected function getUserAuthname(AccountInterface $account) {
+    // First, check if this is an SSO user by looking up authmap.
+    $authname = $this->database->select('authmap', 'a')
+      ->fields('a', ['authname'])
+      ->condition('uid', $account->id())
+      ->execute()
+      ->fetchField();
```
joegl: There is actually a service we should use for this:

- In the externalauth module there is an authmap service: `externalauth.authmap`
- This service has a get method: `public function get(int $uid, string $provider) {`
- This will get the authname based on the `uid` and `provider`.
- The provider in this case is "samlauth"
- When injected should be an easy call:
`$authname = $this->authmap->get($account->id(), 'samlauth');`

joegl: The `hs_person.info.yml` should also probably have a dependency on `stanford_samlauth` or `samlauth` itself.

joegl (changes requested):
Great improvements here! Outside of the discussion we are having on business requirements in the other comment, I only have a few more code-based suggestions.

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

docroot/modules/humsci/hs_person/src/EventSubscriber/UserLoginEventSubscriber.php
nJim: I've removed any logic related to the email address lookup. This work now assumes that the account is logging in with an authname matching the node's field_hs_person_sunet_id.

ahughes3:

Links:
- https://pr1901-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-colorful-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://hs-traditional-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://english-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://history-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com
- https://west-wbkctvhct5gainaeng1bewzcp9qoreyd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d

Visual Diffs:
- https://dashboard.tugboatqa.com/688288579a41360a53b6f22d/visualdiffs

nJim:
@joegl I've removed the email matching part of this service. I'm also using the external externalauth.authmap service and added it as a module dependency. Test plan is updated. This is ready for review.

joegl:
> @joegl I've removed the email matching part of this service. I'm also using the external externalauth.authmap service and added it as a module dependency. Test plan is updated. This is ready for review.

@nJim Awesome! I'll take a look at this shortly.

@ahughes3 I wanted to check in if you want to review this again or if you want to wait to get this merged since it's the Friday before deployment.

ahughes3:
@joegl I'm confirming that I'm good with this getting merged if you have approved this so that Jasper Ridge can test and we have more time to test on Stage before the next prod release. Thank you!

joegl (approved):

joegl:
@ahughes3 This has been merged 👍

PR #1934, opened 2025-08-11
HSD8-1774: Adding back to top to default flexible page layout

ahughes3:
# Summary

Added the back to top block to the default configuration for the flexible (basic page) layout

## Need Review By (Date)

8/13

## Urgency

Low

## Steps to Test

- Review the Configuration addition and confirm it looks correct

- I believe we can't functionally test unless we have a newly provisioned site but the goal would be that on a newly provisioned website the Back To Top Block is on the Flexible Page Layout Display 

# Review Tasks

### Code
- [ ] Are tests provided?

### Code security

### General

### DIFF
--- a/config/default/core.entity_view_display.node.hs_basic_page.default.yml
+++ b/config/default/core.entity_view_display.node.hs_basic_page.default.yml
@@ -10,6 +10,7 @@ dependencies:
     - node.type.hs_basic_page
     - system.menu.main
   module:
+    - hs_blocks
     - hs_layouts
     - layout_builder
     - menu_block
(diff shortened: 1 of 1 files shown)


joegl:
@ahughes3 `'core.entity_view_display.node.hs_*'` configurations are ignored, so these changes won't be imported on existing sites. There will also need to be a database update to make this change programmatically. 4k or I can definitely help with that 👍 😄

ahughes3:
@joegl can the update hook exclude sites that have customizations to the flexible page display or somehow not overwrite other customizations to the display in the process? It's a block that's been manually added to a lot of sites which is why it came up to make it a default but if we can get it added to sites that haven't added it that would be ideal.

joegl:
> @joegl can the update hook exclude sites that have customizations to the flexible page display or somehow not overwrite other customizations to the display in the process? It's a block that's been manually added to a lot of sites which is why it came up to make it a default but if we can get it added to sites that haven't added it that would be ideal.

There can definitely be a way to just insert this block at either the bottom or top of the a specific region. However, I didn't catch that the intention for this was just to affect the default layout (and that you expected to place these manually in existing sites). I'm happy to look at the database update to insert it into existing layouts or we can just go with the manual option.

ahughes3:
@joegl I discussed this internally with the H&S team and we are good with having the block placed by default for new sites going forward and placing the block manually for existing sites where it hasn't been added.

joegl:
@ahughes3 Sounds good Albert 👍

joegl (approved):

PR #1999, opened 2025-10-03
Fix DDEV config file extension

cienvaras:
# Summary
[DDEV config file](https://docs.ddev.com/en/stable/users/configuration/config/) should use `.yaml` extension, it fails if extension is `.yml`

## Backstory
Minor fix to the work done in [this PR](https://github.com/SU-HSDO/suhumsci/pull/1983), that changed how ddev config is handled

## Need Review By (Date)
10/08

## Urgency
medium

## Steps to Test
1. In a fresh dev environment, follow the instructions in the DDEV readme to set up the environment
2. Confirm that everything works as expected and that you're able to access a site after the setup


joegl:
Thanks for catching this @cienvaras

joegl (approved):

PR #2004, opened 2025-10-07
SHS-6346: Lazy load videos

mariannuar:
# Summary
Add logic to lazy load videos

## Backstory
[ClickUp Ticket](https://app.clickup.com/t/36718269/SHS-6346)

## Need Review By (Date)
10/08

## Urgency
medium

## Steps to Test
1. Here's a list of pages with videos, please take a look at them in your local environment (`ddev.site` -  import config before testing and remember to update the URL of the following) or test environment:

- https://creees.stanford.edu/events/video-recordings 
- https://biology.stanford.edu/ (This doesn't apply, it's an auto-playing local video, so it should look the same)
- https://anthropology.stanford.edu/
- https://classics.stanford.edu/
- https://humanbiology.stanford.edu/
- https://internationalrelations.stanford.edu/
- https://music.stanford.edu/play/ensembles/chamber-music-stanford
- https://statistics.stanford.edu/
- https://urbanstudies.stanford.edu/

2. If you want to test this in a Tugboat preview, please go to [Colorful](https://hs-colorful-7mm016d7tpoyqnmfbcclnwzgwrj4jp5s.tugboatqa.com/components/text-area-typography) or [Traditional](https://hs-traditional-7mm016d7tpoyqnmfbcclnwzgwrj4jp5s.tugboatqa.com/components/text-area-typography)
3. Every place where there was a video, there's should be thumbnail for that video and a `Play button`
4. If you click on the button, it should reproduce/load the video
5. There are some views that are configured to use Media Blazy (Example of use: https://creees.stanford.edu/events/video-recordings). The idea is to not use this anymore, so please edit the view in the following way:
6. Go to fields, and exclude from display the `Media: Blazy` and include in display the `Media: Video URL`:

Screenshot 2025-10-07 at 9 47 38 AM

7. The formatter of `Media: Video URL` have to be `Plain text`
8. Then add a new field for `Media: Thumbnail` and configure it like this:

Screenshot 2025-10-13 at 4 27 00 PM

In the `Rewrite results`, enable "Override the output of this field with custom text" and add this markup:

```
<article class="hb-media-video">
  <div class="hb-video-lazy" data-video="{{ field_media_oembed_video }}">
    <div class="hb-video-lazy__thumb">
      {{ thumbnail__target_id }}
      <span class="hb-video-lazy__play" role="button" aria-label="Play video">
        <span hidden>Play video</span>
      </span>
    </div>
  </div>
</article>
```

9. Now go to the `Pattern: Settings`
10. Add `Image` as the destination for  `Media: Thumbnail` and only for this field
11. Save it, and now the cards should look like this and should work as mentioned in step 6

Screenshot 2025-10-07 at 9 51 23 AM

12. Just a heads up that in International Relations there's a link to media that when click on the play button it redirects to the URL of the `anchor` so please remove it in this scenarios

Screenshot 2025-10-13 at 5 17 26 PM

13. Just a heads up that if the video is unavailable, the thumbnail will not build correctly in the server

---
  - https://app.asana.com/0/0/1211722332419909


config/default/core.entity_view_display.media.video.full.yml
```diff
@@ -19,13 +18,10 @@ bundle: video
 mode: full
 content:
   field_media_oembed_video:
-    type: oembed
+    type: string
```
cienvaras, line 23: oembed is able to generate the thumbnail using an image style (look up for the "Thumbnail" field in the "Manage display" settings):

Screenshot 2025-10-07 at 6 24 04 PM

That would probably work better than generating it dynamically in the template. If possible, use a responsive image style to avoid showing huge images in mobile. We don't know the rendered size of the video beforehand, we need the largest images for each resolution, so "Full Responsive" can be a good option (the original quality of the thumbnail could be an issue here, but maybe it's possible to tweak it in the oembed settings).

docroot/themes/humsci/humsci_basic/src/js/shared/media/lazy-load-video.js
```diff
@@ -0,0 +1,83 @@
...
+      const videos = once('video-lazy', '.video-lazy', context);
+
+      videos.forEach((video) => {
+        const provider = video.dataset.videoProvider;
+        const { videoId, isPlaylist, playlistId } = video.dataset;
+
+        // YouTube playlist: get first video thumbnail dynamically
+        if (provider === 'youtube' && isPlaylist) {
+          // Use a free CORS proxy to fetch the playlist HTML
+          const proxyUrl = `https://corsproxy.io/?https://www.youtube.com/playlist?list=${playlistId}`;
```
cienvaras: This is not a good idea, we're relying on a free external service for showing data and that can bring security and availability issues. That's why it's better to generate the thumbnails in the server (ideally using oembed).

docroot/themes/humsci/humsci_basic/templates/field/field--field-media-oembed-video.html.twig
```diff
@@ -1,3 +1,58 @@
...
+#}
+{% for item in items %}
+  {% set video_url = item.content['#context'].value ?? item.content['#url'] ?? '' %}
+  {% set provider = video_url matches '/youtu/' ? 'youtube' : (video_url matches '/vimeo/' ? 'vimeo' : 'other') %}
+  {% set video_id = '' %}
+  {% set thumb = '' %}
+  {% set is_playlist = false %}
+
+  {# YouTube #}
+  {% if provider == 'youtube' %}
```
cienvaras: You're getting the thumbnail here, but also have similar code in the js file. There might be something I'm missing, but why in both places? It should only happen in the server, ideally using oembed as mentioned above, to simplify all the logic.

docroot/themes/humsci/humsci_basic/templates/views/views-view-field--custm-creees-media-videos--field-media-oembed-video.html.twig
```diff
@@ -0,0 +1,84 @@
+{#
```
cienvaras: Isn't it possible to use the new logic changing the view configuration, instead of adding a new template? We would need to do this for any custom view that uses videos and that would complicate things for H&S. Maybe after switching to oembed it will be more simple.

cienvaras (changes requested):
@mariannuar This is a good start, but still needs work. In a nutshell, we need to change how thumbnails are being handled. Please check the inline comments for the details. Also, I checked the [Text Area Typography](https://hs-colorful-7mm016d7tpoyqnmfbcclnwzgwrj4jp5s.tugboatqa.com/components/text-area-typography) in Tugboat and the vimeo thumbnails are not working, due to CORS issues:

Screenshot 2025-10-07 at 6 35 56 PM

That's why we need to generate the thumbnails in the server.

Finally, make sure the the thumbnail and the rendered video are the same size. The last video in the [Text Area Typography](https://hs-colorful-7mm016d7tpoyqnmfbcclnwzgwrj4jp5s.tugboatqa.com/components/text-area-typography) has a larger thumbnail (it might be because the video is not available, but still worth checking) and the content around it moves when the video is loaded.

mariannuar:
Hi @callinmullaney This is ready for review again

callinmullaney (commented):
@mariannuar Nice work on this PR. I can confirm that this is working great and your documentation on how to update a view is accurate. I'm seeing these videos only being loaded to my browser upon clicking the play button. 
Screenshot 2025-10-14 at 11 41 32 AM


I have two suggestions and two comments to consider but may be out of scope for this ticket:
1. I think we need a unique image style and responsive image just for "Video Thumbnail". It can likely keep similar crop/scale effects but separating this into it's own image style/responsive image should make long term maintenance easier because we'll know exactly what it affects.
2. In relation to creating its own image style/responsive image can you review the size/resolution of what is getting served? I'm seeing low resolution thumbnails being rendered for all instances which makes the larger videos look blurry. https://hs-colorful.suhumsci.loc/components/text-area-typography for example. In this screenshot you'll see I'm loading a 295px/166px image into a 882px/496px space: 
Screenshot 2025-10-14 at 12 51 13 PM

The following are comments to consider but may be out of scope:
1. Using the video thumbnail from youtube makes this PR consistent with the current implementation but also results in a lot of black bars. Consider adding "thumbnail" media reference field so authors can customize their video thumbnails. If no image is provided then it defaults back to the current functionality you've implemented so far. 
Screenshot 2025-10-14 at 1 19 41 PM
2. There is an assumption with the styles in this PR that every video will be `aspect-ratio: 16 / 9;` Consider using JS to calculate the actual aspect-ratio and removing any height/width attributes. Something like 

```
const player = video.querySelector('video, iframe');
const playerWidth = player.getAttribute('width') ? player.getAttribute('width') : player.offsetWidth;
const playerHeight = player.getAttribute('height') ? player.getAttribute('height') : player.offsetHeight;
const aspectRatio = '${playerWidth} / ${playerHeight}';
const poster = video.querySelector('.video__placeholder');
const playBtn = video.querySelector('.video__play');

// Set aspect ratio of player.
player.style.aspectRatio = aspectRatio;

// Clear static height/width attributes.
player.removeAttribute('height');
player.removeAttribute('width');
```

mariannuar:
Hi @callinmullaney I have added the new patch so the thumbnails served have higher resolution. However, I found out that with existing videos, the thumbnail will look blurry. If we add the same video with everything as a new Media, it will serve a thumbnail with higher resolution. Please check out the difference between the first thumbnail (Align — Break text) and the second one (Align-centre) in this [page](https://hs-colorful-7mm016d7tpoyqnmfbcclnwzgwrj4jp5s.tugboatqa.com/components/text-area-typography) and let me know your thoughts.

About this:

> I think we need a unique image style and responsive image just for "Video Thumbnail". It can likely keep similar crop/scale effects but separating this into it's own image style/responsive image should make long term maintenance easier because we'll know exactly what it affects.

That's a good idea. However, I kind of remember that we want to avoid having more image styles in this site. Part of the work [here](https://app.clickup.com/t/36718269/SHS-6395) and all the tickets before that one, has been to rework the image style and optimize that.


> Using the video thumbnail from youtube makes this PR consistent with the current implementation but also results in a lot of black bars. Consider adding "thumbnail" media reference field so authors can customize their video thumbnails. If no image is provided then it defaults back to the current functionality you've implemented so far.

Great idea. I think it is necessary to do that. We will address this in a follow up ticket.

callinmullaney:
@mariannuar The results of the patch are looking promising. The vimeo videos are still not ideal resolution for the larger videos but this appears to be the limit of what vimeo offers. They still look better than the lower resolution that we were previously getting. I tested this with a youtube video and the thumbnail they provide look really good. Also, the existing responsive image you're using for the thumbnails should work perfectly so thanks for pushing back on my previous feedback and providing context.
Screenshot 2025-10-17 at 10 38 17 AM

Couple things:
1. Instead of adding the patch to the codebase I think we should use the URL instead: https://git.drupalcode.org/project/drupal/-/merge_requests/9518.diff 
2. I found the following open issue we might want to implement depending on client feedback. No action item here but something to keep in mind if we do need to figure out a solution for previously generated thumbnails. I applied this patch locally and it seemed to work great. Basically, it adds a bulk operations action to the media library that allows site admin to generate new thumbnails: https://www.drupal.org/project/drupal/issues/2983456

mariannuar:
@callinmullaney Ready for another review! Thanks!

callinmullaney (approved):
@mariannuar Lots of great work went into this PR. Everything looks good to me and it's ready to ship over to the client.

mariannuar:
Thank you @callinmullaney!

@ahughes3 @Jnwibowo7321 This should be ready for review!

ahughes3:

Links:
- https://pr2004-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com
- https://hs-colorful-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com
- https://hs-traditional-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com
- https://english-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com
- https://history-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com
- https://west-wwzjexzooi0psmlzefxnvghedshzwu2b.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/68f814570f93f808081fcf07

ahughes3:

Links:
- https://pr2004-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-colorful-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-traditional-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://english-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://history-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://west-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af

Visual Diffs:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af/visualdiffs

ahughes3:

Links:
- https://pr2004-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-colorful-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-traditional-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://english-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://history-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://west-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af

Visual Diffs:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af/visualdiffs

Jnwibowo7321 (approved):
LGTM on tugboat

Jnwibowo7321:
Checking with you @cienvaras if the changes you've requested are also in or would this need another round of review?

cienvaras (approved):

cienvaras:
@Jnwibowo7321 Yes, the changes I requested are included, we're good to go.

ahughes3:

Links:
- https://pr2004-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-colorful-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-traditional-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://english-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://history-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://west-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af

Visual Diffs:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af/visualdiffs

docroot/themes/humsci/humsci_basic/templates/field/field--field-media-oembed-video.html.twig
```diff
@@ -1,3 +0,0 @@
-{{ attach_library('humsci_basic/video-with-caption') }}
```
joegl, line 1: Keeping this template around might be handy in case we want to use this formatter or need to revert.

composer.json
```diff
@@ -249,7 +249,9 @@
...
                 "Cannot read properties of undefined (reading 'settings') with dialog.position.js https://www.drupal.org/project/drupal/issues/3356667": "patches/core/core-mr-8892.patch",
                 "https://www.drupal.org/project/drupal/issues/3207813: ModuleHandler skips all hook implementations when invoked before the module files have been loaded": "patches/core/core-mr-6976.patch",
-                "https://www.drupal.org/project/drupal/issues/3202896: Do not display oEmbed resource error to anonymous users": "patches/core/core-3202896-mr-9357-20250410.patch"
+                "https://www.drupal.org/project/drupal/issues/3202896: Do not display oEmbed resource error to anonymous users": "patches/core/core-3202896-mr-9357-20250410.patch",
+                "Fix oEmbed video thumbnail resolution https://www.drupal.org/project/drupal/issues/3447326": "https://git.drupalcode.org/project/drupal/-/merge_requests/9518.diff",
+                "Adds bulk operations action to media library that allows site admin to generate new thumbnails https://www.drupal.org/project/drupal/issues/2983456": "https://www.drupal.org/files/issues/2025-09-16/2983456-168-10.5.x.patch"
```
joegl: Please download patches into the codebase. See doc here: https://github.com/SU-HSDO/suhumsci/blob/develop/patches/README.md

joegl (changes requested):
A couple of suggestions. There should probably be a no-javascript fallback of some kind (even just displaying the link to the video if JS is disabled).

This seems to be a completely 100% custom implementation. What is the plan for maintaining/future proofing? Is there no library or other module that could be used? How is this an improvement over the core "lazy" loading implementation?

Thanks!

ahughes3:

Links:
- https://pr2004-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-colorful-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-traditional-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://english-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://history-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://west-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af

Visual Diffs:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af/visualdiffs

ahughes3:

Links:
- https://pr2004-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-colorful-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://hs-traditional-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://english-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://history-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com
- https://west-otf3aux0fcwxt7vmu3d2audsfncuv2nn.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af

Visual Diffs:
- https://dashboard.tugboatqa.com/69028437659fe652c28595af/visualdiffs

Jnwibowo7321:
Hi @mariannuar just confirming with you if this is ready for re-review again after Joe's feedback?

mariannuar:
Hi @joegl @Jnwibowo7321 ! Sorry for the late reply. Here are the answers to your questions @joegl ( cc: @cienvaras)

> There should probably be a no-javascript fallback of some kind (even just displaying the link to the video if JS is disabled).

Thank you so much for this. It totally makes sense and I missed it. I added a `<noscript>` link inside the video wrapper so users without JavaScript can still access the video directly.

> This seems to be a completely 100% custom implementation. What is the plan for maintaining/future proofing?

For future proofing & maintenance:

- The implementation uses plain JavaScript and Drupal’s behavior system, so it remains compatible across core updates.
- It relies only on stable, long-standing embed URL patterns from YouTube/Vimeo.
- If a change occurs upstream, it can be patched easily without dependency on a third-party library.
- We have used similar approaches to this one in several projects for similar cases.

> Is there no library or other module that could be used?

Yes, there's a module ([Blazy](https://www.drupal.org/project/blazy)) that offers some functionalities similar to what this script does. However, we are trying to stop even to remove it because it offers a lot of functionalities that are already in core and we are not using.

> How is this an improvement over the core "lazy" loading implementation?

Lazy loading (via loading="lazy") only defers rendering of iframes or images that are immediately in the DOM. It doesn’t prevent the iframe’s request from being made at page load. This script delays the creation of the iframe itself until the user explicitly clicks “Play,” resulting in significantly faster initial page loads, especially for multiple embedded videos. Lazy loading loads the video, but if the user does not click on it, there is no point in having an iframe in the DOM.
 
 Let us know if you have more questions. Thank you!

Jnwibowo7321:
Great thank you @mariannuar , passing it back to you @joegl

joegl:
> Hi @joegl @Jnwibowo7321 ! Sorry for the late reply. Here are the answers to your questions @joegl ( cc: @cienvaras)
> 
> > There should probably be a no-javascript fallback of some kind (even just displaying the link to the video if JS is disabled).
> 
> Thank you so much for this. It totally makes sense and I missed it. I added a `<noscript>` link inside the video wrapper so users without JavaScript can still access the video directly.
> 
> > This seems to be a completely 100% custom implementation. What is the plan for maintaining/future proofing?
> 
> For future proofing & maintenance:
> 
> * The implementation uses plain JavaScript and Drupal’s behavior system, so it remains compatible across core updates.
> * It relies only on stable, long-standing embed URL patterns from YouTube/Vimeo.
> * If a change occurs upstream, it can be patched easily without dependency on a third-party library.
> * We have used similar approaches to this one in several projects for similar cases.
> 
> > Is there no library or other module that could be used?
> 
> Yes, there's a module ([Blazy](https://www.drupal.org/project/blazy)) that offers some functionalities similar to what this script does. However, we are trying to stop even to remove it because it offers a lot of functionalities that are already in core and we are not using.
> 
> > How is this an improvement over the core "lazy" loading implementation?
> 
> Lazy loading (via loading="lazy") only defers rendering of iframes or images that are immediately in the DOM. It doesn’t prevent the iframe’s request from being made at page load. This script delays the creation of the iframe itself until the user explicitly clicks “Play,” resulting in significantly faster initial page loads, especially for multiple embedded videos. Lazy loading loads the video, but if the user does not click on it, there is no point in having an iframe in the DOM.
> 
> Let us know if you have more questions. Thank you!

Thanks Maria! This is a fabulous explanation and makes sense to me,

joegl (approved):
@ahughes3 @Jnwibowo7321 This should be reviewed in-depth on stage before release to make sure there aren't any tweaks needed.

ahughes3:
@joegl Would it be possible to roll this one back from this week's prod release? We won't be able to test this one properly on Stage and I have some questions I would like to sync up with 4K about around the play button icon.

joegl:
> @joegl Would it be possible to roll this one back from this week's prod release? We won't be able to test this one properly on Stage and I have some questions I would like to sync up with 4K about around the play button icon.

Yea no problem 👍 I'll do that later today

PR #2008, opened 2025-10-08
DDEV improvements

dalin-:
# Summary
* Fixed problems I ran into trying to switch to DDEV
* Fixed future problems if we try to roll out changes to everyone (e.g. upgrade PHP, or add new sites)

## Need Review By (Date)
no rush

## Urgency
low

## Steps to Test
If you've never tried DDEV, then try these new install instructions.  
If you've already got DDEV, then maybe try a `ddev delete` and start again.


callinmullaney (approved):
@dalin- documentation updates look good. I ran into the same `ddev auth ssh` issue when initially setting up the project. I also ran into `Unknown MySQL server host...` but ended up restarting my computer to resolve. It was likely the same conflict with lando like you mentioned here. :shipit:

callinmullaney:
@joegl Small PR ready for you to take a look. Just updating some ddev documentation.

joegl:
@dalin- We recently made a change to move `.ddev/config.yaml` to `.ddev/default.config.yaml` and `.gitignore` the `config.yaml` because it's being used to detect a `ddev` install in the build scripts, and then run commands through the `ddev` wrapper.  

See: 
* #1983
* https://github.com/SU-HSDO/suhumsci/blob/develop/hooks/common/theme/theme-build.sh
* https://hsdoit.slack.com/archives/C03DD43ACPK/p1758660429215659

For example, I have `ddev` installed locally but don't have this repo set up with DDEV. Because of the conditionals in the theme-build, having the `.ddev/config.yaml` runs the build commands through `ddev` on my machine. 

I definitely understand wanting to go this route, but then we'd need to update the build scripts to account for the scenario outlined above.

joegl:
@dalin- @nJim Thinking about this. Couldn't we just use the `$DDEV_PROJECT` variable in all the build scripts to detect DDEV? Since the scripts are triggered during other commands, those commands theoretically would all be ran with `ddev` and `$DDEV_PROJECT` would be set, but running without `ddev` would not set that variable. What was the original reason for using the `config.yaml` and shell detection as the conditional and not `$DDEV_PROJECT`?

dalin-:
@joegl 
Well this sent me down a bit of a rabbit hole where I discovered about 5 different ways to compile the theme.  I made a few changes to try and unify the approaches.  

> What was the original reason for using the config.yaml and shell detection as the conditional and not $DDEV_PROJECT?

It was replicating [what was done for Lando](https://github.com/SU-HSDO/suhumsci/blob/2f858850c5b4e9c9d5f0a8d87415fa8376842665/hooks/common/theme/theme-build.sh#L10).  Why was Lando doing it that way?  I think so that you can install the entire site on a fresh laptop with only 3 dependencies: Git, NVM, and DDev/Lando.  

Take a look at this and lemme know what you think.

hooks/common/theme/theme-test.sh
dalin-, line 1: Yes `npm run test` is broken.  There's a ticket in the next sprint to look into that.

cienvaras (approved):

joegl:
@dalin- The build-theme script is failing on my local now. It hangs on the "nvm" line. 

```
composer build-theme
> [ -d "$HOME/.nvm" ] && [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" && command -v nvm  &> /dev/null && nvm install
sh: 1: nvm: not found
Script [ -d "$HOME/.nvm" ] && [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" && command -v nvm  &> /dev/null && nvm install handling the build-theme event returned with error code 127

nvm
```

dalin-:
@joegl 
I've been trying to figure out which section of this oneliner is failing for you, but I'm struggling.  Can you try each section between `&&` in your terminal, one at a time?

joegl:
> @joegl I've been trying to figure out which section of this oneliner is failing for you, but I'm struggling. Can you try each section between `&&` in your terminal, one at a time?

Yea, sorry for the short comment earlier. I'll pull down again later and do a little more debugging on my end before reporting back.

dalin-:
@joegl 
Any chance that you could review this?  I think we need to make more changes to Ddev config to stay compatible with the current stable version.

dalin-:
@joegl 
We've got a few different incompatibilities with this old ddev config and both Acquia Cloud Next, and the current stable version of DDev.   We've got some awkward workarounds for those things.  But it'd be much easier if we could roll them out to everyone in this standardized config file.  

Let me know if it's easier to break out the `composer.json` change into a separate PR.

composer.json
```diff
@@ -221,12 +221,12 @@
     },
     "scripts": {
         "build-theme": [
-            "npm ci --prefix docroot/themes/humsci/humsci_basic",
-            "npm run build --prefix docroot/themes/humsci/humsci_basic"
+            "[ -d \"$HOME/.nvm\" ] && [ -s \"$NVM_DIR/nvm.sh\" ] && \\. \"$NVM_DIR/nvm.sh\" && command -v nvm  &> /dev/null && nvm install",
```
joegl: @dalin- Let's get this in ASAP for you. I think the one change I'm stumbling on and causing issues are these lines added to the composer.json. When I remove them everything runs fine for me.

Also, couldn't this part be built into the `theme-get-command` script?

joegl (changes requested):

composer.json
dalin-: Ahh, good idea.  I was trying to think about how to make the existing command work.  But yes, it's even easier to just call it directly.  

Ready for another review.

dalin- (commented):
Ready for another review.

joegl (approved):
In minimal testing this works and looks good to me. We can re-visit in a follow up if I run into any issue, but I want to unblock your locals.

joegl:
@dalin- I may have to revert something here. Running basic commands, like a blt drupal:sync, is causing DDEV to re-write all my local.settings.php when I'm not using DDEV, and blow up my local environment.

joegl:
> @dalin- I may have to revert something here. Running basic commands, like a blt drupal:sync, is causing DDEV to re-write all my local.settings.php when I'm not using DDEV, and blow up my local environment.

Scratch that. I had installed DDEV for the project to test it and so `.ddev/php` existed. I wiped the directory and checked it back out on my local to get rid of `.ddev/php`. This should hopefully prevent commands from being wrapped with `ddev`. I may at some point create and implement an override variable in the `theme-get-command.sh` to force a particular wrapper, and leave it empty by default. I don't think I'm at that point yet though.

PR #1967, opened 2025-09-08
HSD8-1785: Update default Country value for Person Address field

ahughes3:
# Summary

Updated the Mailing Address field on the Person content type to have the Country set to United States by default so that the Address, City, State, Zip fields are displayed by default.

## Backstory

When manually creating a Person node a user has to select United States even though it's the only option before seeing any of the Address Fields.

## Need Review By (Date)
9/24

## Urgency
medium

## Steps to Test

1. Log into a site - https://pr1967-xrnoz0fgq1rzftbsuoernwblyk7xk270.tugboatqa.com/
2. Create a Person node
3. Under the Contact information tab look for the Mailing Address section
4. Expand the Mailing Address section and confirm that the Street Address, City, State, Zip fields are displayed by default

# Review Tasks

### Code
- [ ] Are the naming conventions following our standards?
- [ ] Are PHP functions and variables in `snake_case` and not `camelCase`?
- [ ] Does Drupal code follow [Drupal Coding Standards](https://www.drupal.org/docs/develop/standards/php/php-coding-standards)?
- [ ] Does the code have sufficient inline comments?
- [ ] Is there anything in this code that would be hidden or hard to discover through the UI?
- [ ] Are there any [code smells](https://blog.codinghorror.com/code-smells/)?
- [ ] Are tests provided?

### Code security
- [ ] Is all [user input properly sanitized when rendered](https://www.drupal.org/docs/8/security/drupal-8-sanitizing-output)?
- [ ] Any obvious [security flaws or new areas for attack](https://www.drupal.org/docs/8/security)?

### General
- [ ] Is there anything included in this PR that is not related to the problem it is trying to solve?
- [ ] Is the approach to the problem appropriate?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211210952045894

PR #2046, opened 2025-11-07
SHS-6411: For mailing address, allow all countries, and make all subfields optional

dalin-:
# Summary
* Allow addresses from any country, but default to the U.S.
* Make all address sub-fields optional.

This is only a change to how the address is input.  Display remains the same.  I couldn't find any code (other than config) that uses `field_hs_person_mail`. 

## Backstory
Jasper Ridge has many non-US based people.  There's also several profiles to be imported from D7 that have only partial addresses (perhaps just the city, state, country).  
[SHS-SHS-6411](https://app.clickup.com/t/36718269/SHS-SHS-6411)

## Need Review By (Date)
asap

## Urgency
high

## Steps to Test
1. create some user profiles.  
2. Try non-US addresses, or partial addresses.

---
  - https://app.asana.com/0/0/1211210952045894

### DIFF
--- a/config/default/field.field.node.hs_person.field_hs_person_mail.yml
+++ b/config/default/field.field.node.hs_person.field_hs_person_mail.yml
@@ -17,23 +17,50 @@ label: 'Mailing Address'
 description: ''
 required: false
 translatable: false
-default_value: {  }
+default_value:
+  -
+    langcode: null
+    country_code: US
+    administrative_area: ''
+    locality: ''
+    dependent_locality: null
+    postal_code: ''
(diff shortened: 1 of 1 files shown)


joegl:
@ahughes3 I'm going to wait for you to take a look because I know we had ran into issues with this previously.

joegl (approved):
Albert and I reviewed in our meeting today. LGTM

joegl:
For reference, previous attempts and reversion:
- #1992
- #1967

Making the field overrides optional prevented the required issue we ran into earlier.

PR #2061, opened 2025-11-19
Update drush alias hosts to point to ACN

pookmish:
# Summary
Update drush alias hosts to point to ACN


joegl (approved):

joegl:
I confirmed the prod CNAME still points to the cloud classic because it has not been upgraded to ACN yet, and once it is upgraded the CNAME should change to ACN, thus no changes will be required to the drush hosts for the prod upgrade.

PR #1997, opened 2025-10-02
SHS-6341: Remove modernizr javascript dependency

cienvaras:
# Summary
- [Modernizr is deprecated](https://www.drupal.org/node/3333253), remove it as a theme dependency

## Backstory
- [ClickUp ticket](https://app.clickup.com/t/36718269/SHS-6341)

## Need Review By (Date)
10/08

## Urgency
medium

## Steps to Test
1. No special testing needed, as Modernizr was only used in obsolete browsers. Just confirm that everything works as expected as part of the regular QA.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211566031442925

PR #2093, opened 2026-01-13
SHS-6490: Add JS class to <html> earlier to prevent FOUC issue

cienvaras:
# Summary
- Add a script in the page header that adds a `js` class to the `html` element. Drupal also does this out-of-the-box, but the script is located at the end of the file, which causes a delay that displays an incorrectly styled version of the main navigation. 

## Backstory
- [ClickUp Ticket](https://app.clickup.com/t/36718269/SHS-6490)

## Need Review By (Date)
01/21

## Urgency
medium

## Steps to Test
1. Visit different pages on the test sites, both as anonymous and logged user
2. Confirm that in all cases, the main menu looks as expected and that the issue described in [the ticket](https://app.clickup.com/t/36718269/SHS-6490) is not happening anymore

**Note:** The original issue is hard to reproduce, because it was produced by a delay in the execution of a script and that depends on a lot of different conditions: network speed, processing power, size of the page, etc. It didn't happen only for logged in users when navigating to a second level link as described in the ticket, there were other instances where the issue could happen. However, with this fix, the issue shouldn't happen anymore under any conditions. 

---
  - https://app.asana.com/0/0/1212365599241238
  - https://app.asana.com/0/0/1212843343328685

### DIFF
--- a/docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
+++ b/docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
@@ -1,0 +1,1 @@
+document.documentElement.classList.add('js');
--- a/docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
+++ b/docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
@@ -1,1 +1,2 @@
+// Add a `js` class to the html element before page renders, to prevent FOUC.
 document.documentElement.classList.add('js');
--- a/docroot/themes/humsci/humsci_basic/humsci_basic.info.yml
+++ b/docroot/themes/humsci/humsci_basic/humsci_basic.info.yml
@@ -8,6 +8,7 @@ dependencies:
   - hs_layouts:hs_layouts
   - fontawesome:fontawesome
 libraries:
+  - humsci_basic/init
   - humsci_basic/main-content-fallback
   - humsci_basic/editoria11y
 
--- a/docroot/themes/humsci/humsci_basic/humsci_basic.libraries.yml
+++ b/docroot/themes/humsci/humsci_basic/humsci_basic.libraries.yml
@@ -1,3 +1,8 @@
+init:
+  header: true
+  js:
+    dist/js/init.js: { preprocess: false }
+
 accordion:
   js:
     dist/js/accordion.js: {}
--- a/docroot/themes/humsci/humsci_basic/webpack.common.js
+++ b/docroot/themes/humsci/humsci_basic/webpack.common.js
@@ -12,6 +12,7 @@ const jsSrcDir = path.resolve(__dirname, 'src/js/');
 
 // Shared JS files. Used in both traditional and colorful themes.
 const shared = {
+  init: 'shared/init/init.js',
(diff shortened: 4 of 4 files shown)


ahughes3:

Links:
- https://pr2093-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com
- https://hs-colorful-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com
- https://hs-traditional-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com
- https://english-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com
- https://history-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com
- https://west-a3xxpgpo0eccs0hddkkmeompqahtksxj.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/696598cd0256e14c274cdfea

Visual Diffs:
- https://dashboard.tugboatqa.com/696598cd0256e14c274cdfea/visualdiffs

mariannuar (approved):
@cienvaras LGTM! 🪄

cienvaras:
@ahughes3 @Jnwibowo7321 Ready for review

Jnwibowo7321 (approved):
LGTM on Tugboat!

docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
```diff
@@ -0,0 +1 @@
+document.documentElement.classList.add('js');
```
joegl, line 2: @cienvaras as I understand it the purpose of the Drupal script is:

1. By default the `no-js` class is added to `<html>`
2. If JavaScript is enabled in the browser, the `no-js` class is removed and `js` is added

The script should also probably remove the `no-js` class then if it exists. Also, could you add a one-line comment here to indicate this is being done manually to fix an FOUC issue caused by the Drupal load order?

Thanks!

joegl (changes requested):

docroot/themes/humsci/humsci_basic/src/js/shared/init/init.js
cienvaras: @joegl Drupal doesn't use a `no-js` class, it only adds a `js` class when javascript is initialized (see https://git.drupalcode.org/project/drupal/-/blob/main/core/misc/drupal.init.js?ref_type=heads#L7).

Modernizr is the one that replaces `no-js` with `js`, but [we removed it because it's deprecated](https://github.com/SU-HSDO/suhumsci/pull/1997). In that PR we also removed all references to `no-js` in the styles and markup, switching to use the `js` Drupal adds. The problem is that `drupal.init.js` is loaded in the page footer and doesn't block rendering, causing the FOUC. That's why I'm adding this script, to add the `js` class before rendering happens.

> Also, could you add a one-line comment here to indicate this is being done manually to fix an FOUC issue caused by the Drupal load order?

Done!

Thanks!

joegl: Awesome, sounds good to me. Thanks!

joegl (approved):

PR #2097, opened 2026-01-14
SHS-6515: Patch Swiper module to prevent and squelch PHP warning

dalin-:
# Summary
Fixing a patch to this contrib module by a well-meaning, but inexperienced engineer where the problems weren't caught by the maintainer. 

Caused this PHP warning
```
Uncaught PHP Exception TypeError: "count(): Argument #1 ($value) must be of type Countable|array, null given" at /var/www/html/docroot/modules/contrib/swiper_formatter/swiper_formatter.module line 233
```

and these visible warnings

image

Additionally there are some JS console log entries:
```
Swiper Loop Warning: The number of slides is not enough for loop mode, it will be disabled or not function properly. You need to add more slides (or make duplicates) or lower the values of slidesPerView and slidesPerGroup parameters
```

However, this appears to be an issue with the underlying library:
https://github.com/nolimits4web/swiper/issues/7586

Since there's no impact on the behaviour of the carousel, and no visual impact, I recommend not digging deeper. 

## Backstory
  [SHS-6515](https://app.clickup.com/t/36718269/SHS-6515)

## Need Review By (Date)
ASAP

## Urgency
high

## Steps to Test

1. Visit 
https://hs-colorful-xhqsir77jwgxxhd0suydp4xw212hdjse.tugboatqa.com/components/hero-images-full-width-region-and-main-region/tall-spotlights-classic 
both logged-in and logged-out.  No warnings visible on screen.  No PHP errors in the logs. 

---
  - https://app.asana.com/0/0/1212663001749735
  - https://app.asana.com/0/0/1212826435462365


cienvaras (changes requested):
@dalin- Works as expected. SWS requires to store patches locally, so please save it in `patches/contrib` and update `composer.json` to use the local version. See https://github.com/SU-HSDO/suhumsci/blob/develop/docs/Patching.md

cienvaras (approved):

cienvaras:
Updated PR to use local patch instead of pulling it from the merge request.

@ahughes3 @Jnwibowo7321 Ready for review

Jnwibowo7321 (approved):
LGTM on Tugboat!

joegl (approved):
LGTM. Always nice to see contributions back to modules 👍

PR #2151, opened 2026-03-12
SHS-6572: Add taxonomy term page view

dalin-:
# Summary
Replace the blank full-page taxonomy term with something useful.

[SHS-6572](https://app.clickup.com/t/36718269/SHS-6572)

## Backstory
Currently when anyone views a taxonomy term, they see nothing:
Screenshot 2026-03-12 at 10 25 24 AM

## Proposed Solution
We set up a default view (`/admin/structure/views/view/hs_default_taxonomy_term/edit/page_1`) that is enabled for all vocabularies.  If a site doesn't like this layout, they can clone the default view and adjust to their liking (just like our other default views). 

## Need Review By (Date)
standard timing

## Urgency
medium

## Steps to Test
1. `/admin/structure/taxonomy`
3. click through to see the list of terms in any vocabulary.  before this PR you would've seen the blank screen
e.g. https://hs-traditional-0vcqrgibufbedzvhmtvawiuuiuwdjjdq.tugboatqa.com/news-categories/students
Screenshot 2026-06-23 at 12 01 07 PM

---
  - https://app.asana.com/0/0/1215895701031946
  - https://app.asana.com/0/0/1213337721864000


ahughes3:

Links:
- https://pr2151-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com
- https://hs-colorful-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com
- https://hs-traditional-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com
- https://english-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com
- https://history-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com
- https://west-5itrbeer7qbjum8lmykhr9ajyofnrjno.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69d963004703f7f22cda2d5d

Visual Diffs:
- https://dashboard.tugboatqa.com/69d963004703f7f22cda2d5d/visualdiffs

ahughes3 (changes requested):
@dalin- Can you update the Formatter for the Body field to be `Summary or trimmed`, the Default body field appears long for some nodes.

image

Everything else looks great, Thank you!

ahughes3 (approved):
LGTM!

config/default/views.view.hs_default_taxonomy_term.yml
```diff
@@ -1,17 +1,19 @@
-uuid: 8e26711b-5aa1-4c8d-b1af-ff36bcac453b
+uuid: aea6d9c0-4183-4291-b9a5-dfe2d8bdcccf
```
joegl, line 1: Getting rid of the `views.view.taxonomy_term` view in the config won't wipe it from existing sites due to the partial config imports (no config gets deleted on imports). That will probably create a conflict then with the two views on sites where they're both enabled. 

We could:
1. Add an update hook to delete the existing view leaving only the new view.
2. Update the status of this view to "FALSE" and have sites opt-in to using this new view.
3. Completely replace the existing view and not do the `hs_default` rename and new UUID.

@dalin- @ahughes3

config/default/views.view.hs_default_taxonomy_term.yml
```diff
@@ -292,18 +511,20 @@ display:
...
+      tags:
+        - 'config:field.storage.node.body'
   feed_block:
     id: feed_block
     display_title: 'Feed Block'
     display_plugin: block
     position: 1
     display_options:
-      enabled: true
+      enabled: false
```
joegl: I'm not sure if this is used anywhere, but I wanted to ask if there is a specific reason for disabling/getting rid of this display?

config/default/core.extension.yml
```diff
@@ -254,6 +254,7 @@ module:
   video_embed_media: 0
   view_unpublished: 0
   viewfield: 0
+  views_autocomplete_filters: 0
```
joegl: I know I'm a stickler when it comes to adding new contrib modules, and I don't have any particular concerns or objections to adding this one after reviewing it, but even a quick overview/justification would be great 👍 

I would also enable the module in an update hook.

composer.lock
```diff
@@ -27035,5 +27096,5 @@
         "php": ">=8.3"
     },
     "platform-dev": {},
-    "plugin-api-version": "2.9.0"
+    "plugin-api-version": "2.6.0"
```
joegl: I'd suggest updating your composer version to 2.9+.

docroot/modules/humsci/hs_views_helper/hs_views_helper.module
```diff
@@ -114,6 +115,26 @@ function hs_views_helper_form_views_exposed_form_alter(&$form, FormStateInterfac
...
+/**
+ * Implements hook_views_pre_render().
+ *
+ * Alters the page title for taxonomy terms rendered by the
+ * hs_default_taxonomy_term view to [vocabulary]: [term].
+ *
+ * This replicates the standard taxonomy term canonical page title handling by
+ * \Drupal\stanford_fields\Hook\StanfordFieldsHooks::preprocessPage().
+ */
+function hs_views_helper_views_pre_render(ViewExecutable $view) {
```
joegl:
```suggestion
function hs_views_helper_views_pre_render(ViewExecutable $view) : void {
```

joegl (changes requested):
This would be a good addition. A couple of questions/comments.

config/default/core.extension.yml
dalin-: @ahughes3 
We didn't end up going in the direction of the original prototype.  This module is a remnant of that.  It provides this exposed filter:
Screenshot 2026-07-14 at 3 30 47 PM

Do you think this is useful?

config/default/views.view.hs_default_taxonomy_term.yml
dalin-: Having just one RSS feed enabled here, without a unified site-wide/platform-wide RSS approach, seems like an accident.  If we do want RSS, then let's think about it more broadly.  But in 2026, I don't think there's a realistic need/use for RSS.  

@ahughes3 What do you think?  Is there a need to add an RSS ticket to the backlog?

config/default/views.view.hs_default_taxonomy_term.yml
dalin-: I went with #1 since we want this view to be enabled on all sites, with the option to override it (like other default views) when a site wants customized taxonomy listings.

composer.lock
dalin-: @joegl 
Fixed.  

To avoid this problem in the future, we could consider enforcing Composer ^2.9

config/default/views.view.hs_default_taxonomy_term.yml
joegl: Unless it's being consumed it's probably a remnant of the default Drupal installation since this is based on a pre-packaged view. I agree it's best to disable it if it's not being used.

dalin-: We bumped this to a separate ticket.  Reverted for now.

config/default/core.extension.yml
dalin-: Albert thinks we don't need this right now.  Reverting.

joegl (approved):

PR #2166, opened 2026-03-20
HSD8-0000: Globalize and update services.yml configuration

joegl:
# Summary
This PR removes all site-specific `services.yml` files and introduces a single global `services.yml` (located at `docroot/sites/services.yml`) for all Drupal multisites. It also updates each site's `settings.php` to stop including its own `services.yml`, and instead includes the global file via `global.settings.php`. The global `services.yml` has also been updated to include new and recommended Drupal configuration settings, ensuring all sites share up-to-date, secure, and consistent configuration.

## Backstory
After updating the autologout module, a warning appeared: `Undefined array key "cookie_samesite" in autologout_page_attachments_alter()`. Investigation revealed that many sites' `services.yml` files were missing the `cookie_samesite` setting, which was added to Drupal 10.1. Rather than update each site individually, this PR:
- Deletes all site-specific `services.yml` files.
- Adds a single global `services.yml` in `docroot/sites/` and includes it for all sites via `global.settings.php`.
- Comments out the line in each site's `settings.php` that previously included its own `services.yml`.
- Updates the global `services.yml` to include new and recommended Drupal configuration settings:
  - `cookie_samesite: Lax` (fixes autologout warning and improves CSRF protection)
  - `security.enable_super_user` (added in Core 10.3, set to `true` by default)
  - `sid_length` and `sid_bits_per_character` (explicitly set to Drupal defaults)
  - `name_suffix` (left empty, matches default)
  - `allowed_file_extensions` (Twig autoloader, matches default)
  - `renderer_config.debug` (left as false, default)
  - `cors.config` (added, left as default)
  - `queue.config` (added, left as default)
- Ensures all sites are using the same, up-to-date configuration, reducing maintenance and risk of drift.

## Urgency
Medium

## Additional Notes
- `security.enable_super_user` is set to `true` (the Drupal default). This can be revisited if a stricter policy is desired.
- All other added/updated settings match Drupal's current defaults and should not impact normal site operation.
- This change greatly simplifies multisite configuration and future maintenance.


joegl:
@pookmish I'm curious what you think mainly about the `cookie_samesite` setting change.

Ignore this and refer to our slack discussion.

PR #2172, opened 2026-03-30
SHS-6537: Media Item edit Button not Clickable

mariannuar:
# Summary
Allow editors to interact with the media item edit button when the image has a link event on a card

## Backstory
[SHS-6537](https://app.clickup.com/t/36718269/SHS-6537)

## Need Review By (Date)
31/03

## Urgency
medium

## Steps to Test
1. Log in to the Tugboat test site ([Colorful](https://hs-colorful-p6cr9whf6hlqzkwwzhc99dyns8dmp7yv.tugboatqa.com/user/login), [Traditional](https://hs-traditional-p6cr9whf6hlqzkwwzhc99dyns8dmp7yv.tugboatqa.com//user/login))
2. Go to the following view that already has the same configuration as this view on [this site](https://ccsre.stanford.edu/people/ccsre-team/affiliated-faculty), where the issue is still happening (If you inspect the code, the anchor should be wrapping the media item edit button, that's why the issue is happening): 

     - https://hs-colorful-p6cr9whf6hlqzkwwzhc99dyns8dmp7yv.tugboatqa.com/default-views/people

3. Select the first person (which is the only one that has an external profile URL) and confirm that you can click on the Media item edit button and interact with any action there. It shouldn't redirect you to another page that is not related to the media item.
4. Please also click on the caption button and confirm you are not redirected to another page and please also confirm this work as an anonymous user
5. Confirm that if you click on any other area of the image, it should redirect you where it's supposed to
6. Please also confirm the previous work on the following view that has the same configuration of the previous view but using Vertical Linked Cards instead:
    - https://hs-traditional-p6cr9whf6hlqzkwwzhc99dyns8dmp7yv.tugboatqa.com/default-views/people
7. Go check other views and confirm they work the same

---
  - https://app.asana.com/0/0/1213158958810963
  - https://app.asana.com/0/0/1213882949416424


dalin- (commented):
@mariannuar 
The code looks like it could be a good solution, but when I test in the browser, the cards are not linked.  Possibly because I had to manually re-run GitHubActions a few times to get flaky tests to complete (maybe that blew away your Views customizations).  I took a look at the views and it's not obvious to me what needs to change to match the ccsre production site.  

Can you re setup the colourful site?

mariannuar:
@dalin- Ready! And sorry, they were not linked because I forgot to rearrange the fields, they must be in this order for the cards to be linked:

Screenshot 2026-03-31 at 1 10 39 PM

dalin- (changes requested):
@mariannuar 
Thanks for setting up Tugboat. This is working well in my testing.  

There's a closely related bug that I think we should also address here:  If I click the caption hide/show buttons, I'm redirected to the external site.
Image
cc @ahughes3 

I'd also love to get @cienvaras ' perspective on whether this code is scoped correctly to cover all the scenarios where this behaviour can occur.  It's probably correct as-is, but I don't have enough understanding of how/where teasers can be overridden.

mariannuar:
@dalin- I’ll wait for @cienvaras’ input and review. Based on his recommendation, I’ll go ahead and make the change for the case you also found. Thanks!

docroot/themes/humsci/humsci_basic/src/js/shared/cards/linked-cards.js
```diff
@@ -12,6 +13,20 @@ import addCardEvents from './event-handlers';
 
       // Loop through each card
       cards.forEach((card) => {
+        /**
+         * Handles click events on a card image link that contains a Drupal contextual
+         * region. Prevents the image anchor from navigating when interacting with
+         * contextual controls, while still allowing contextual links to work normally.
+         *
+         */
+        const cardImageLink = card.querySelector('.hb-card__img a');
```
cienvaras: This doesn't work with vertical linked cards, because the image container class is different. I know the bug only covers regular vertical cards, but let's cover all types, to prevent future issues.

docroot/themes/humsci/humsci_basic/src/js/shared/cards/linked-cards.js
```diff
@@ -12,6 +13,20 @@ import addCardEvents from './event-handlers';
 
       // Loop through each card
       cards.forEach((card) => {
+        /**
```
cienvaras, line 16: This only applies for authenticated users, so let's check first before running this code. You can check the `user` property in `drupalSettings`.

docroot/themes/humsci/humsci_basic/src/js/shared/cards/image-link-handler.js
```diff
@@ -0,0 +1,23 @@
...
+      return;
+    }
+
+    // Block navigation for anything else inside the contextual region
+    // (e.g. the trigger button)
+    if (contextual) {
+      e.preventDefault();
+      e.stopPropagation();
+      // eslint-disable-next-line no-useless-return
+      return;
```
cienvaras: This `return` is not necessary because we're at the end of the function. That's why eslint flagged it. Let's remove it.

cienvaras (changes requested):
@mariannuar There are some small fixes needed. Also, please fix the issue that Dave found about the caption buttons not working. Thanks!

mariannuar:
@cienvaras Ready for another review!

cienvaras (approved):
@mariannuar Looks good, thanks for the fixes!

@ahughes3 @Jnwibowo7321 Ready for you

Jnwibowo7321 (approved):
LGTM on tugboat

joegl (approved):

PR #1979, opened 2025-09-23
SHS-6380: Spotlight Redesign

cienvaras:
# Summary
- Update spotlight component and related image styles, to apply new design

## Backstory
- ClickUp Tickets:
    - [Implementation: Spotlight Redesign](https://app.clickup.com/t/36718269/SHS-6380)
    - [Implement: new approach for long captions](https://app.clickup.com/t/36718269/SHS-6419)
    - [Implementation: Spotlight width](https://app.clickup.com/t/36718269/SHS-6390)

## Need Review By (Date)
11/12

## Urgency
medium

### Spotlight Redesign

1. In both [Traditional](https://hs-traditional-t28udft8krytqgumtb9umgov0q0g9qay.tugboatqa.com/) and [Colorful](https://hs-colorful-t28udft8krytqgumtb9umgov0q0g9qay.tugboatqa.com/) test sites, visit the following pages:
    1. Tall Spotlight(s) - Classic (`/components/hero-images-full-width-region-and-main-region/tall-spotlights-classic`)
    2. Short Spotlight(s) (`/components/hero-images-full-width-region-and-main-region/short-spotlights`
    3. Tall Spotlight(s) - Expanded (`/components/hero-images-full-width-region-and-main-region/tall-spotlights-expanded`)
    4. Short Spotlight(s) - Expanded (`/components/hero-images-full-width-region-and-main-region/tall-spotlights-expanded`)
2. For all spotlights, confirm that they match the new designs ([Traditional](https://www.figma.com/design/7xoXRIQgSMGlUszWYFbI09/SHS-Library---Traditional?node-id=1861-1039&t=NqgOzEOuXJpfPVt5-4), [Colorful](https://www.figma.com/design/UXl20gIVr8Vg4OmaTQBnYr/SHS-Library---COLORFUL?node-id=3893-3490&t=zoBaupMsH7WGf0hM-4))
3. Create a flexible page, add spotlights with the following characteristics (both in the full width and the main content area):
    1. Video with both right and left alignments
    2. No media
    3. Any other field combinations not tested before
4. Confirm that the spotlights match the designs in all cases
5. Edit or add a spotlight. Confirm that the "Image Style" field has the following help text: 
    > Framed images will show the same aspect ratio on all screens, but expanded images will more completely fill the screen. Be sure to set the image focal point and check multiple screen sizes to confirm that important parts of the image are not cut off. 

### Spotlight width

[...]

PR #2207, opened 2026-04-27
SHS-6638: Fix temp unavailable error after deleting media in spotlights

dalin-:
# Summary
Fatal PHP error in custom code.

## Backstory
  [SHS-6638](https://app.clickup.com/t/36718269/SHS-6638)

## Need Review By (Date)
ASAP

## Urgency
high

## Steps to Test
1. Create a spotlight with an image media item
2. Delete the media item (without first changing the spotlight)
    e.g. https://hs-traditional-ksvi6tehwsxttm1mor9iuqvbj0adznt6.tugboatqa.com/components/hero-images-full-width-region-and-main-region/tall-spotlights-classic
3. Try and view the spotlight.  Before this PR you'd see a fatal PHP error.  Now it should work without issue. 

---
  - https://app.asana.com/0/0/1214273606724062
  - https://app.asana.com/0/0/1214354688651999

### DIFF
--- a/docroot/modules/humsci/hs_layouts/hs_layouts.module
+++ b/docroot/modules/humsci/hs_layouts/hs_layouts.module
@@ -190,10 +190,10 @@ function hs_layouts_preprocess_pattern_spotlight(&$variables) {
     }
     else {
       $media = $paragraph->get('field_hs_spotlight_image')->entity;
-      if ($media->bundle() == 'video') {
+      if ($media && $media->bundle() === 'video') {
         $spotlight_type = 'video';
       }
-      elseif ($media->bundle() == 'image' && $spotlight_type == 'classic') {
+      elseif ($media && $media->bundle() === 'image' && $spotlight_type === 'classic') {
         $view_mode = $context->getProperty('view_mode');
         $image_style = $view_mode == 'default' ? 'spotlight_framed_tall' : 'spotlight_framed_short';
         $variables['image']['field_hs_spotlight_image'][0]['#stanford_media_image_style'] = $image_style;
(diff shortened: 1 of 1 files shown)


joegl (commented):
Definitely want to get this fixed. I'm curious though, this is part of the else statement of:

```
if ($paragraph->get('field_hs_spotlight_image')->isEmpty()) {
```

It's interesting the `field_hs_spotlight_image` field can pass the "isEmpty" check (has a value) but then there is no media entity attached. 

When the field is empty, it's setting `hb-spotlight--no-media` class. If we implement the fix here, there's clearly still a scenario unfolding where `$media` has no value, but the `hb-spotlight--no-media` class won't get set either. It might be worth re-visiting that condition to make sure classes are getting set appropriately here.

The logic here was introduced as part of https://github.com/SU-HSDO/suhumsci/pull/1979

dalin-:
@joegl 

> When the field is empty, it's setting hb-spotlight--no-media class. If we implement the fix here, there's clearly still a scenario unfolding where $media has no value, but the hb-spotlight--no-media class won't get set either.

Good catch.  Fixed. 

> It's interesting the field_hs_spotlight_image field can pass the "isEmpty" check (has a value) but then there is no media entity attached.

This is a quirk with Drupal's entity reference fields.  The field isn't empty, it's pointing to media item 1234, but Drupal can't load media item 1234 because it no longer exists.  Drupal doesn't clean up references when an entity is deleted.  It's a difficult problem to solve because there's conflicting "correct" approaches.  

There's an alternative approach with `$entity->field->referencedEntities()`, which we could switch to, but it would be a bigger change.

joegl (approved):

PR #2208, opened 2026-04-27
SHS-6637: Add a 'reset' button to SiteImprove details page

dalin-:
# Summary
Some production sites had a faulty SiteImprove site ID.  There's not a direct link between a site and its ID.  So our custom code determines it on the fly and stores it in the State system (by reviewing the full list of domains in SiteImprove).  If that ID ever changes (it has for some reason.  I'm not sure how or why), there's not currently a way to regenerate it.  This PR adds a button for that. 

## Backstory
  [SHS-6637](https://app.clickup.com/t/36718269/SHS-6637)

## Need Review By (Date)
May 5

## Urgency
medium

## Steps to Test
1. Visit `/admin/config/services/siteimprove/details`
2. Click the reset button.

Note that non-production sites will not have a base_url, API key, username.  If you click the button it'll show an error message telling you to input that.  I have no way to find those values, so obtaining those is an exercise for the reader.  

Note that the existing code was moved from a controller to a form.  To see a diff that tries harder to show moved files, use `git diff --find-renames=20%`

---
  - https://app.asana.com/0/0/1214167860790373
  - https://app.asana.com/0/0/1214618124995366


ahughes3:

Links:
- https://pr2208-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com
- https://hs-colorful-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com
- https://hs-traditional-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com
- https://english-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com
- https://history-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com
- https://west-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69efc5c83b1266d092361daa

Visual Diffs:
- https://dashboard.tugboatqa.com/69efc5c83b1266d092361daa/visualdiffs

cienvaras (approved):
@dalin- LGTM, thanks!

@ahughes3 @Jnwibowo7321 Ready for you. I already added the api username and keys to the [English tugboat test site](https://english-ul5omarylffecc1zraokwbsmsdkwm9tz.tugboatqa.com/). Let me know if you want me to configure other of the test sites. Once merged, it should work in stage, because keys are already configured in acquia.

Jnwibowo7321 (approved):
LGTM on Tugboat!

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -202,7 +209,7 @@ protected function call(string $method, string $endpoint, array $query = []): ob
 
     try {
       $response = $this->http_client->request($method, $this->baseUrl . $endpoint, $options);
-      $response_body = json_decode($response->getBody());
+      $response_body = json_decode($response->getBody(), FALSE, 16, JSON_THROW_ON_ERROR);
```
joegl, line 214: Curious why capping the depth at 16?

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -65,23 +65,28 @@ public function __construct(
...
   ) {
     $this->config = $config_factory->get('hs_siteimprove.settings');
-    $this->baseUrl = $this->config->get('base_url') ?: 'https://api.siteimprove.com/v2';
+    $this->baseUrl = $this->config->get('base_url') ?: 'https://api.us.siteimprove.com/v2';
     $this->cache = $cache;
   }
 
   /**
-   * {@inheritdoc}
+   * Is there enough config info to make API calls?
```
joegl:
```suggestion
   *  Checks whether the API credentials are configured.
```

Phrase as a non-question with what it actually is doing. Something like the above or "Determines if there is enough config info to make API calls", etc.,

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -65,23 +65,28 @@ public function __construct(
...
-   * {@inheritdoc}
+   * Get a list of all sites.
+   *
+   * @param bool $refresh
+   *   Force the list to be refetched from the API.
+   *
+   * @return array
    */
-  public function getSites(): array {
+  public function getSites(bool $refresh = FALSE): array {
```
joegl: I'd either make this protected or move it into the interface.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -165,9 +171,10 @@ public function getCurrentSite(): ?object {
         }
       }
```
joegl, line 174:
```suggestion
        $this->logger->error('Could not find a SiteImprove site for the site: @production_url', ['@production_url' => $production_url]);
      }
```

Moving the new logger here I think makes more sense. Outside of the try/catch block it would produce a double logger when there's an exception, but the goal seems to be to log in the event the foreach loop fails to return a site that matches the production URL.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -165,9 +171,10 @@ public function getCurrentSite(): ?object {
...
     }
-    catch (SiteImproveException $e) {
+    catch (\Exception $e) {
       $this->logger->error('Failed to get current site: @message', ['@message' => $e->getMessage()]);
     }
+    $this->logger->error('Could not find a SiteImprove site for the site: @production_url', ['@production_url' => $production_url]);
```
joegl:
```suggestion
```
See above comment for moving this.

docroot/modules/humsci/hs_siteimprove/hs_siteimprove.install
```diff
@@ -0,0 +1,33 @@
...
+  if (!$site_id) {
+    // On local, keys are probably not set up.  Allow DB updates to continue.
+    // Everywhere else, we want to know if there's a problem.
+    $envs = EnvironmentDetector::getEnvironments();
+    $env = key($envs) ?? 'local';
+    if ($env !== 'local') {
+      throw new UpdateException('Could not retrieve SiteImprove site ID. Check the API settings at /admin/config/services/siteimprove.');
+    }
+    return t('SiteImprove site ID not refreshed. Intentionally allowing execution to continue because this is a local site.')->render();
+  }
```
joegl, line 19:
```suggestion
function hs_siteimprove_update_10001(): string {
  /** @var \Drupal\hs_siteimprove\SiteImprove $site_improve */
  $site_improve = \Drupal::service('hs_siteimprove.connector');
  $site_id = $site_improve->getCurrentSiteId(TRUE);

  if (!$site_id) {
    \Drupal::logger('hs_siteimprove')->warning('Could not refresh SiteImprove site ID. Check the API settings at /admin/config/services/siteimprove.');
    return t('SiteImprove site ID not refreshed. Check API settings if the Broken Links block is not working.')->render();
  }

  return t('SiteImprove site ID refreshed: @id.', ['@id' => $site_id])->render();
}
```

A few things:
1. Environment detector isn't necessarily reliable enough to use in production code like this outside of setting up configuration, nor should it be used to alter database updates. If a database update requires knowledge of local environments it should probably be reconsidered/refactored.
2. I don't think a failure to refresh the site id warrants an exception or hard fail of a database update and potentially break a site deployment. Logging a warning should be sufficient.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -211,7 +218,7 @@ protected function call(string $method, string $endpoint, array $query = []): ob
 
       throw new SiteImproveException('API request failed with status code: ' . $response->getStatusCode());
     }
-    catch (GuzzleException $e) {
+    catch (\Exception $e) {
       throw new SiteImproveException('API request failed: ' . $e->getMessage(), 0, $e);
     }
```
joegl, line 221:
```suggestion
catch (\JsonException $e) {
    throw new SiteImproveException('API response was not valid JSON: ' . $e->getMessage(), 0, $e);
}
catch (\Exception $e) {
    throw new SiteImproveException('API request failed: ' . $e->getMessage(), 0, $e);
}
```

With the `JSON_THROW_ON_ERROR` in `json_encode` it's possible to throw a JsonException. Catch that and report separately from the API request.

joegl (changes requested):

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: Mostly I wanted to get the `JSON_THROW_ON_ERROR`.  The default max-depth for json_decode is 256, but 16 is waaaay deeper than anything that we're currently getting from SiteImprove.  I figured that if they start sending something 16 levels deep, we probably want to know and change how we approach this.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/hs_siteimprove.install
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
joegl: Makes sense to me 👍

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -211,7 +218,10 @@ protected function call(string $method, string $endpoint, array $query = []): ob
 
       throw new SiteImproveException('API request failed with status code: ' . $response->getStatusCode());
```
joegl: This code chunk actually fits better outside of the try block and after the two other catches. Then the order would be:
1. Try API call
2. Try and parse the response as JSON
3. Catch a JSON parse issue
4. Catch an API request issue
5. Check the status code of a successful, parsed call is 200 and return
6. If not 200, throw an exception for the status code

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
```diff
@@ -136,7 +141,7 @@ public function getPagesWithBrokenLinks(): ?array {
     }
 
     try {
-      $pages = $this->call('GET', "/sites/{$site_id}/quality_assurance/links/pages_with_broken_links", ['page_size' => 5]);
+      $pages = $this->call('GET', "/sites/$site_id/quality_assurance/links/pages_with_broken_links", ['page_size' => 5]);
       // Cache for 5 minutes.
       $this->cache->set('hs_siteimprove_broken_links', $pages->items, time() + 900);
```
joegl: I know this is not part of this PR, but this is 15 minutes, not 5.

joegl (changes requested):
@dalin- I caught one more thing I didn't see last time. Otherwise this is good to go.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: Makes sense.  Fixed.

docroot/modules/humsci/hs_siteimprove/src/SiteImprove.php
dalin-: Updated.

joegl (approved):

PR #2212, opened 2026-04-30
SHS-6635: Views with no content in a Collection should not cause an empty color band

mariannuar:
# Summary
- Remove extra div and background in collection when view has no results, no "no results" message and the title is hidden. Also when there are not items or content in a collection.
- Fixed bug that made possible to save a collection without entering a title, even when the title was required by the configured title options

## Backstory
[SHS-6635](https://app.clickup.com/t/36718269/SHS-6635)
[SHS-6672](https://fourkitchens.clickup.com/t/36718269/SHS-6672)

## Need Review By (Date)
05/12

## Urgency
medium

### SHS-6635

1. Go to https://hs-traditional-hhhkddnw8yvffd4dwkjws50fhnlfyyxh.tugboatqa.com/test-empty-view-collection
2. Compare it with https://hs-traditional.stanford.edu/test-empty-view-collection
3. Confirm the Collections that have Views with no results, views with no "no results" message, views title, is hidden or that have nothing inside the collection, are not displaying the background color or any extra div wrapper
4. Confirm there's a VRT page for `/test-empty-view-collection` as well
5. Visit other pages that have a Collection with a view with results and confirm they work and look the same: https://hs-traditional-hhhkddnw8yvffd4dwkjws50fhnlfyyxh.tugboatqa.com/mixed-pages/mixed-page-test-collection-developers-catalog

### SHS-6672

6. Create a Flexible page or Priv page
7. Add a Collection
8. Don't add anything in the form
9. If you save it, it should display an error message related to the required fields: Items and Title
10. Please add any item, but don't add a Title
11. Collapsed the "Collection" component
12. Try to save it and now you should have a warning related to the required Title

## Notes
- The "Collection" `#parents` warning issue: Fixed "Undefined array key #parents" warnings in FormState::getError() when "Items" is required but empty. This is a warning that doesn't affect the site's functionality and it's only visible when the devel module is enabled (local and tugboat).
- The "Conditional title field required not being announced as required on save" issue: The title field in Collections paragraphs was created conditionally required using `#states`, which is client-side only and does not enforce validation on form submission when the component is collapsed. Added a server-side `#element_validate` callback in the form alter that checks the value of `field_title_settings` and throws a form error if the title is required but empty.

---
  - https://app.asana.com/0/0/1211009488090612
  - https://app.asana.com/0/0/1214951500561595


ahughes3:

Links:
- https://pr2212-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com
- https://hs-colorful-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com
- https://hs-traditional-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com
- https://english-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com
- https://history-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com
- https://west-uxxgegvm5w3ebpiuzorfqageznijwft6.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69f3d3d36e883d5ae37dc859

Visual Diffs:
- https://dashboard.tugboatqa.com/69f3d3d36e883d5ae37dc859/visualdiffs

cienvaras (commented):
@mariannuar Looks good in general, but you're hiding the content for all empty views, even when the collection title is present. Can you confirm if that's the expected behaviour? Based on the steps to reproduce in the ticket, it should be only for collections without title, but I'm not 100% sure.

cienvaras (approved):
@mariannuar LGTM! Just made a very small change.

@ahughes3 @Jnwibowo7321 Ready for your review. The issue that @Jnwibowo7321 reported ("Undefined array key #parents") is a warning that doesn't impact the site's functionality and it's only visible when the development module is enabled (only on local and tugboat), so it won't be visible to anyone on the production site. You can confirm this after this is deployed to stage.

The other [bug found that made possible to save collections without a title after collapsing a collection paragraph](https://fourkitchens.clickup.com/t/36718269/SHS-6672) is fixed.

Jnwibowo7321 (approved):
LGTM on Tugboat

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,42 @@ function humsci_basic_preprocess_field(&$variables) {
...
+
+/**
+ * Implements hook_preprocess_ENTITY__BUNDLE().
+ */
+function humsci_basic_preprocess_paragraph__hs_collection(&$variables) {
+  $paragraph = $variables['paragraph'];
+  $variables['has_content'] = FALSE;
+
+  foreach ($paragraph->field_hs_collection_items as $item) {
+    $entity = $item->entity;
```
joegl:
```suggestion
    $entity = $item->entity;
    if (!$entity) {
      continue;
    }
```

Verify entity exists.

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,42 @@ function humsci_basic_preprocess_field(&$variables) {
...
+
+    if ($entity->bundle() !== 'hs_view') {
+      $variables['has_content'] = TRUE;
+      continue;
```
joegl:
```suggestion
      break;
```
Once `has_content` is TRUE, the loop can be broken and becomes unnecessary for further looping.

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,42 @@ function humsci_basic_preprocess_field(&$variables) {
...
+    }
+
+    $view->setDisplay($view_field->display_id);
+    $view->preExecute();
+    $view->execute();
+
+    $has_empty_handlers = !empty($view->display_handler->getHandlers('empty'));
+
+    if (!empty($view->result) || $view_field->show_title || $has_empty_handlers) {
+      $variables['has_content'] = TRUE;
```
joegl, line 789:
```suggestion
      $variables['has_content'] = TRUE;
      break;
```
Content has been found and reported; can break the loop now as further runs are unnecessary.

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
```diff
@@ -177,12 +179,51 @@ function hs_paragraph_types_field_widget_complete_paragraphs_form_alter(array &$
...
+  $parents = $element['#parents'];
+
+  if (!$parents) {
+    return;
+  }
+
+  // Find 'field_title' in the parents array to locate the subform boundary.
+  $field_title_pos = array_search('field_title', $parents);
+
+  if (!$field_title_pos) {
```
joegl:
```suggestion
  if ($field_title_pos === FALSE) {
```
The `array_search` function returns the index of the array or `false` if not found. If the first element in the array contains the field title, `array_search` returns `0`. However, it would not pass the current condition because in this case `0` equals `FALSE`. Change to a strict `boolean` match for this scenario.

config/default/field.field.paragraph.hs_collection.field_hs_collection_items.yml
```diff
@@ -25,7 +25,7 @@ entity_type: paragraph
 bundle: hs_collection
 label: Items
 description: ''
-required: false
+required: true
```
joegl, line 28: @Jnwibowo7321 This change (which is a good improvement) will cause editors to run into a validation error when trying to save pages that have collections without items. I just want to call this out as behavior that should probably be communicated to editors.

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,42 @@ function humsci_basic_preprocess_field(&$variables) {
...
+      continue;
+    }
+
+    $view->setDisplay($view_field->display_id);
+    $view->preExecute();
+    $view->execute();
+
+    $has_empty_handlers = !empty($view->display_handler->getHandlers('empty'));
+
+    if (!empty($view->result) || $view_field->show_title || $has_empty_handlers) {
```
joegl:
```suggestion
    $view->setDisplay($view_field->display_id);
    $has_empty_handlers = !empty($view->display_handler->getHandlers('empty'));

    if ($view_field->show_title || $has_empty_handlers) {
      $variables['has_content'] = TRUE;
      break;
    }

    $view->preExecute();
    $view->execute();  

    if (!empty($view->result)) {
      $variables['has_content'] = TRUE;
      break;
    }
```

Evaluate empty handlers and title separately and before executing the view and checking results. Executing the view can be expensive for performance, and should be avoided if not necessary.

joegl (changes requested):

cienvaras (approved):

ahughes3 (approved):
LGTM!

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
```diff
@@ -177,12 +179,51 @@ function hs_paragraph_types_field_widget_complete_paragraphs_form_alter(array &$
           'required' => [$header_conditions],
           'visible' => [$header_conditions],
         ];
+
+        // Keep server-side validation by adding an element_validate callback
+        // instead of relying solely on #states for the required check.
+        $element['subform']['field_title']['widget'][0]['value']['#element_validate'][] = 'hs_paragraph_types_validate_collection_title';
```
joegl: Because this relies on another field in the form for validation, you can add the validation to the form itself. Then the validator code has access to all field values and you don't have to traverse up the parent elements to get the other field value. See matching code in the validator.


```suggestion
        $element['subform']['#element_validate'][] = 'hs_paragraph_types_validate_collection_title';
```

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
```diff
@@ -177,12 +179,51 @@ function hs_paragraph_types_field_widget_complete_paragraphs_form_alter(array &$
...
+
+  $requires_title = in_array($title_settings_value ?? '', [
+    'collection_title_visible',
+    'collection_title_sr_only',
+  ]);
+
+  if ($requires_title && empty($element['#value'])) {
+    $form_state->setError($element, t('Title field is required.'));
+  }
+}
```
joegl, line 219:
```suggestion
function hs_paragraph_types_validate_collection_title(&$element, FormStateInterface $form_state, &$form) {
  $title_settings = $element['field_title_settings']['widget'][0]['value']['#value'] ?? '';
  $title_element = $element['field_title']['widget'][0]['value'] ?? NULL;

  $requires_title = in_array($title_settings, [
    'collection_title_visible',
    'collection_title_sr_only',
  ]);

  if ($requires_title && $title_element !== NULL && empty($title_element['#value'])) {
    $form_state->setError($title_element, t('Title field is required.'));
  }
}
```

Now with the validator attached to the entire subform, this code can be simplified.

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,54 @@ function humsci_basic_preprocess_field(&$variables) {
...
+
+    $view->preExecute();
+
+    $view->execute();
+
+    if (!empty($view->result)) {
+      $variables['has_content'] = TRUE;
+      break;
+    }
+  }
```
joegl:
```suggestion
  // Store referenced entities in a variable and only load once since both loops
  // use it.
  $items = $paragraph->field_hs_collection_items->referencedEntities();

  // Check if any of the referenced entities are not hs_view paragraphs and skip
  // more expensive Views checks if so.
  foreach ($items as $entity) {
    if ($entity->bundle() !== 'hs_view') {
      $variables['has_content'] = TRUE;
      return;
    }
  }

  // All remaining item are hs_view paragraphs; check for content.
  foreach ($items as $entity) {
    $view_field = $entity->field_hs_view->first();
    if (!$view_field) {
      continue;
    }

    // Load the view and check access.
    $view = Views::getView($view_field->target_id);
    if (!$view || !$view->access($view_field->display_id)) {
      continue;
    }

    // Check for a view title and an empty results text or handler.
    $view->setDisplay($view_field->display_id);
    $has_empty_handlers = !empty($view->display_handler->getHandlers('empty'));
    if ($view_field->show_title || $has_empty_handlers) {
      $variables['has_content'] = TRUE;
      return;
    }

    // Execute the view to check for results (expensive, less performant check
    // saved for last).
    $view->preExecute();
    $view->execute();

    if (!empty($view->result)) {
      $variables['has_content'] = TRUE;
      return;
    }
  }
```

Another suggestion for this. Since it's looping through all paragraphs and immediately stopping and marking `has_content` as `true` if they are not an `hs_view` paragraph, we should run that check first before even looking at an `hs_view` paragraph, because that check is more expensive.

joegl (changes requested):
A few code suggestions. Also a I had a couple questions.

1. For the empty color bands and paragraphs -- is there really not a CSS solution and does it require the potentially expensive PHP code and creation of a `has_content` variable? I know Drupal can make this difficult and understand the necessity if it exists. (@mariannuar @cienvaras )
2. I noticed that some of these functions and changes are made to both the private and regular collection paragraph, and some only affect the regular collection. Is that intentional? Or should the private paragraph also have the `has_content` application in both the TWIG and `.theme` code?

docroot/modules/humsci/hs_paragraph_types/hs_paragraph_types.module
mariannuar: Good call on attaching to the subform. I kept one thing different from the "read straight off $element" suggestion, and wanted to flag why.
Reading field values off the subform render array ($element['field_title_settings']['widget'][0]['value']['#value']) works when the paragraph is open, but fails when it's collapsed: Paragraphs doesn't render the field widgets as active inputs in collapsed state, so there's no #value on the render tree and the required check silently passes. The suggested code didn't hit this because it read from the value tree via $form_state->getValue(), which Paragraphs preserves regardless of open/collapsed. So I'm sourcing the values from the value tree using the subform's #parents — keeps the no-parent-climbing win, but survives a collapsed paragraph.

joegl: Ahh that's interesting. I wondered why it was reading from the $form_state. Thanks 👍

docroot/themes/humsci/humsci_basic/humsci_basic.theme
```diff
@@ -740,3 +741,49 @@ function humsci_basic_preprocess_field(&$variables) {
...
+  $items = $paragraph->field_hs_collection_items->referencedEntities();
+
+  // Check if any of the referenced entities are not hs_view paragraphs and skip
+  // more expensive Views checks if so.
+  foreach ($items as $entity) {
+    if ($entity->bundle() !== 'hs_view') {
+      $variables['has_content'] = TRUE;
+      return;
+    }
+
```
joegl, line 766:
```suggestion
  $views = [];
  foreach ($items as $entity) {
    if ($entity->bundle() !== 'hs_view') {
      $variables['has_content'] = TRUE;
      return;
    }
    $views[] = $entity;
  }
  
  foreach ($views as $entity) {  
```
Store the views and loop through them separately after.

If there are multiple paragraphs in a collection, and the first one is a view with no empty message and no results, it's going to go through the full execution of a view only for has_content to end up true because one of the other paragraphs has content in it. If it loops through everything except a view first and returns with has_content, we avoid executing a view unnecessarily.

joegl (approved):
This looks good, thanks @mariannuar for all the iteration 👍 

@cienvaras Did you still want to give this a review?
@ahughes3 It's Monday of release week -- do you want to get this merged in yet or wait until next release?

cienvaras (approved):

PR #2246, opened 2026-05-27
Display the correct non-Acquia environment in Environment Indicator

dalin-:
# Summary
All non-Acquia environments were showing as "local"

## Need Review By (Date)
no rush

## Urgency
low

## Steps to Test
1. visit any page on tugboat
2. look at the favicon.  Does it have an "L" overlaid (for "local"), or "C" (for CI)?

### DIFF
--- a/docroot/sites/settings/environment_indicator.php
+++ b/docroot/sites/settings/environment_indicator.php
@@ -30,7 +30,7 @@ switch (EnvironmentDetector::getAhEnv()) {
 
   default:
     $envs = array_filter(EnvironmentDetector::getEnvironments());
-    $env = key($envs);
+    $env = key($envs) ?? 'Local';
     $config['environment_indicator.indicator']['bg_color'] = '#086601';
     $config['environment_indicator.indicator']['fg_color'] = '#fff';
     $config['environment_indicator.indicator']['name'] = $env;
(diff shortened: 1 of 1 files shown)


docroot/sites/settings/environment_indicator.php
```diff
@@ -8,19 +15,24 @@
...
+
   case 'prod':
     $config['environment_indicator.indicator']['bg_color'] = '#000';
     $config['environment_indicator.indicator']['fg_color'] = '#fff';
     $config['environment_indicator.indicator']['name'] = 'Production';
     break;
+
   default:
+    $envs = array_filter(EnvironmentDetector::getEnvironments());
+    $env = key($envs);
```
joegl:
```suggestion
    $env = key($envs) ?? 'Local';
```

There is a rare scenario where these is no environment identified. We could fallback to "Local" or even use "Unknown".

joegl (commented):
I like the intent here. I made a quick suggestion.

There is also the fact that the keys are not necessarily user-oriented in formatting (e.g., "local", "ah_other", "ci" instead of "Local", "Acquia Other", "CI", etc.,) but since this is really a developer only indicator I'm not concerned about that.

PR #2250, opened 2026-05-28
SHS-6610 discovery better way to redirect profiles to external urls

dalin-:
# Summary
Currently, if we want to redirect a node elsewhere (e.g. a person's bio to their canonical Stanford bio), then we're doing it on the listing page within a view.  This is problematic for a few reasons:
* It makes the view more complicated
* It only works on the listing page.  There's other cases to consider:
    * Someone may visit the page directly 
    * The full-page version will still appear in site-wide search results
    * Google will, no doubt, still find the full-page version (same for AI bots).  

The 1.x version of Rabbit Hole module can already do this, but it's rather wonky for admins bypassing the redirect.  We had a fair bit of custom code that tried to get around this, but not in a scalable way. Upgrading to 2.x fixes that problem, but will require cleaning up a fair bit of other code too. 

If we move forward with this: 
* there is still a bit of old custom code that was designed for Rabbit Hole 1.x that will need to be re-evaluated.  See to-dos in code
* config is wonky.  I had all sorts of trouble where I had to remove a bunch of UUIDs to get it to import on Tugboat (or a fresh dump of the live site).  I'm not sure if that's because we're mid-D11 upgrade or what.  Re-evaluate after D11 is launched.
* config_ignore the rabbit_hole stuff.  
* expand the "Bypass Rabbit Hole" permissions to match those that can edit content (i.e. add Author/Preparer)
* code review.

## Backstory
  [SHS-6610](https://app.clickup.com/t/36718269/SHS-6610)

## Need Review By (Date)
This is a discovery ticket, so still a draft.

## Urgency
N/A

## Steps to Test

See this demo video: https://us06web.zoom.us/clips/share/zPPMKZ5DSA6eihXt-EF5Cw

If there's a Tugboat build for this PR, then:
1. Visit /admin/config/content/rabbit-hole
2. Configure the Profile content type:
    * Page redirect
    * redirect to `[node:field_hs_person_external_profile:uri]`
    * Fallabck: Display the page
    * uncheck "Allow these settings to be overridden for individual entities"
    * check "Display a message when viewing the page"
3. Visit a profile that doesn't have a value for External Profile URL (e.g. /people/rachel-ragnhild-carlson ).  There is no redirect for anon users, there is no message on screen.
4. Edit a profile and set the External Profile URL (e.g. /node/14326/edit ).  Save.  There is a redirect for anon users, and a message on screen.


ahughes3:

Links:
- https://pr2250-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com
- https://hs-colorful-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com
- https://hs-traditional-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com
- https://english-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com
- https://history-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com
- https://west-7davlvsgbeyiijk1zqedw3sizwmkmrrb.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a1895398cf38f7cea9c3ce1

Visual Diffs:
- https://dashboard.tugboatqa.com/6a1895398cf38f7cea9c3ce1/visualdiffs

PR #2261, opened 2026-06-03
SHS-6585: drush command to query all sites for reporting

dalin-:
# Summary
For some things we can't do reporting in SiteImprove. For these, we need a tool to run queries against all databases.

## Backstory
  [SHS-6585](https://app.clickup.com/t/36718269/SHS-6585)

### Approach:
A new custom drush command (in `/drush`). It takes two arguments:
* an SQL query
* An environment to run on.

It first confirms that the SQL query is a select statement and not a write statement. Then iterates over each site, and send a query to the given environment.  It reports back a list of each site and environment and the results of the SQL query.  There a few different output formats.

## Need Review By (Date)
no rush

## Urgency
low

## Steps to Test
We used it to get a list of all sites that use Asset Injector for either custom CSS or JS
```
 drush humsci:sql:query "SELECT name FROM config WHERE (name LIKE 'asset_injector.css.%' OR name LIKE 'asset_injector.js.%') AND data LIKE '%s:6:\"status\";b:1;%'" prod --format=raw
```


joegl (commented):
I'll have to think about this one. I see the value and understand the intent, but it feels tightly-scoped and over-engineered  and includes quite a few guardrails that `sql:query` itself doesn't have (though I'm a huge proponent of those guardrails when running anything across all sites on production).

My initial thoughts:
- Are direct SQL queries even necessary here? Running SQL directly should generally be a last resort when the Drupal API can do the job. For the example in the description, you could probably get the same results with `drush php-eval` and `\Drupal::configFactory()->listAll('asset_injector')`.
- Does this need to run on production? Staging is refreshed from prod frequently. I'd be leery of bulk-running SQL queries on production when staging would likely give you the same answer.
- The protections feel disproportionate. A quick shell script looping over sites and calling `drush sql:query` takes a few minutes to write and would bypass all protections here.
- The command feels caught in the middle. The custom Drush commands we have do very specific things and are essentially targeted script runners. This command is neither that nor a general tool. If the guardrails exist because you want non-developers or less SQL-savvy folks to use this safely, then it shouldn't be a query runner at all and should probably be a narrowly scoped command that returns exactly the data needed (e.g., a dedicated command to report on Asset Injector usage across sites). If the audience is developers who can write their own queries, the protections are unnecessary and a simple loop script does the job. A pluggable SQL runner sits in an awkward middle ground that doesn't fit either use case well.
- If a general cross-site runner is the goal, a command that runs any Drush command across all sites would be far more valuable and reusable and would make this file unnecessary.

PR #2268, opened 2026-06-04
HSD8-1902: Use active config_ignore storage on local and remove site sync hooks

joegl:
# Summary
Sets `$settings['config_ignore_storage'] = 'active'` in local settings files and removes the Drush pre/post site sync hooks that manually dropped and restored the local `config_ignore` split. This makes the local config_split's ignore rules apply on exports as well as imports, and simplifies the site sync workflow.

## Backstory
[HSD8-1902](https://stanfordits.atlassian.net/browse/HSD8-1902)

A recent `config_ignore` version change altered the default behavior so that ignore rules are always read from sync storage (`config/default`) during exports, regardless of what config_split patches are active. This caused the local split's `config_ignore` overrides to be silently bypassed on export, meaning role permission changes and other locally-unignored config would not be captured by `drush config-export` as intended.

Setting `config_ignore_storage` to `active` tells `config_ignore` to use whatever is in active configuration (including local split patches) for both imports and exports. As a side effect, this also eliminates the need for the pre/post site sync hooks in `HsHooksDrushCommands.php`, which existed solely to drop and restore the local `config_ignore` split file so that production ignore rules would be used during the first import of a site sync.

The three `core.entity_view_display` config files were updated incidentally during testing. They had a stale `items_per_page: none` value that was corrected to `0` on export.

## Need Review By (Date)
No specific date.

## Urgency
low

## Steps to Test
1. Check out the branch and run `composer install`.
1. Copy `docroot/sites/settings/default.local.settings.php` to your own `local.settings.php` if you have not already done so, to pick up the new setting.
1. Sync a site to local: `drush drupal:sync --site=<SITENAME>`. Confirm the sync completes without errors and that custom site configuration is preserved (no unexpected config deletions or overrides).
1. Make a role permission change on the local site via the UI.
1. Run `drush config-export -y` and confirm the permission change is captured in the export diff.

[HSD8-1902]: https://stanfordits.atlassian.net/browse/HSD8-1902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/default/core.entity_view_display.node.hs_event.default.yml
```diff
@@ -456,7 +456,7 @@ third_party_settings:
               provider: views
               context_mapping: {  }
               views_label: ''
-              items_per_page: none
+              items_per_page: 0
```
imonroe: views_block.items_per_page is typed as string (not integer) so that "none" will work

Changing to 0 changes the behavior. 0 !== 'none' is true, so setItemsPerPage(0) will be called, which in Drupal Views means unlimited items. Each of these view blocks on hs_event, hs_event_series, and hs_research node displays will now load all results without pagination, instead of respecting whatever the view itself configured.  

Is that change intentional?

imonroe (commented):
If the change from "none" to `0` was intentional, this is g2g, but if not, I think we should revert those changes.

config/default/core.entity_view_display.node.hs_event.default.yml
joegl: I believe so but I will have to double check. These were configs that were not getting exported because of the config_ignore rules (that get "fixed" in this PR) and I assumed that meant they came from schema changes that were more correct now.

cienvaras (approved):
@joegl LGTM!

> it would require developers to update their local.settings.php file from the default.local.settings.php file to receive the change. 

That's ok. I'll let the rest of the team know as soon as this get's merged.

config/default/core.entity_view_display.node.hs_event.default.yml
joegl: Good catch. It looks like this was a D11 schema change where it previously accepted string values, and "none" was acceptable. However it got changed and when Drupal encountered "none" it casted it to an INT which produced "0". The new schema requires a min of "1" for the INT value though. Also, "0" producing unlimited results is not part of the `view_block` schema for layouts. The correct value here is just "null".

PR #2270, opened 2026-06-05
SHS-6658: "Button" section from Vertical Button Cards causes an empty button

mariannuar:
# Summary
Fix wrong color when external link and misalignmen

## Backstory
[Comment](https://fourkitchens.clickup.com/t/36718269/SHS-6658?comment=90130273680269) from [SHS-6658](https://app.clickup.com/t/36718269/SHS-6658)

## Need Review By (Date)
06/09

## Urgency
medium

## Steps to Test
1. Visit the following pages on [Colorful](https://hs-colorful-9slo9wfkdyotqrjnaa9sh5cvbqyzwhmh.tugboatqa.com/default-views/people) and [Traditional](https://hs-traditional-9slo9wfkdyotqrjnaa9sh5cvbqyzwhmh.tugboatqa.com/default-views/people) that replicate the first issue in [this comment](https://fourkitchens.clickup.com/t/36718269/SHS-6658?comment=90130273680269)
2. Confirm that the cards with external links on the button look as expected, and the links will have a high-contrast color pairing between the text and button background
3. The issue happening here was because the field that is in the "Button" region was having this customization in its configuration:

Screenshot 2026-06-05 at 11 10 02 AM

5. So, on this page, the misalignment issue was happening because we have a nested `hs-button` class inside the wrapper that already has a `hs-button` class. We recommend avoid having nested classes.
6. The issue with the commas or pipe (|) is because of this configuration:

Screenshot 2026-06-05 at 11 17 43 AM

7. The recommendation is that if these kinds of fields are going to be used in the "Button" region from "Vertical Button Cards", let's not configure the field like that. Disable the "Display all values in the same row" toggle. Also, remove/hide the labels (e.g., "Email", "Website", etc.).
8. Without that, it should look like this 

Screenshot 2026-06-05 at 11 24 50 AM

9. We also recommend to avoid adding this kind of fields (with a lot of links) in the "Button" section, but if it's needed, it should look good.

---
  - https://app.asana.com/0/0/1215247753822356


docroot/modules/humsci/hs_layouts/patterns/vertical-button-card/vertical-button-card.html.twig
```diff
@@ -25,7 +25,7 @@
         {% endif %}
 
         {% if button %}
-          <div class="hs-button">
+          <div class="hb-card__button hs-button">
```
cienvaras: Let's remove the extra class (and related selector changes). It's only necessary to override styles added by extra button classes, but we can instruct the H&S team to not add extra classes on the links when used in the button region of this component.

docroot/themes/humsci/humsci_basic/src/scss/partials/components/_vertical-button-card.scss
```diff
@@ -71,20 +80,30 @@
...
         border-width: 0;
         border-top-width: 2px;
         width: 100%;
       }
+
+      a[href*="//"].hs-external-link,
+      .hs-external-link a[href*="//"] {
+        margin-right: 0;
+
+        &::after {
```
cienvaras: Can we keep the icons, setting the right color? They're there for a reason, better not to hide them.

docroot/themes/humsci/humsci_basic/src/scss/partials/components/_vertical-button-card.scss
```diff
@@ -174,11 +186,23 @@
...
+        background-color: color('primary');
+        border-color: color('primary');
+      }
+
+      a[href*="//"].hs-external-link,
+      .hs-external-link a[href*="//"] {
+        margin-right: 0;
+        color: color('white', 'global');
+
+        &::after {
```
cienvaras: Same as above, let's try to keep the icons.

cienvaras (changes requested):
@mariannuar Looks good! Take a look at the inline comments, but in summary:

- Don't use the extra `hb-card__button`
- Don't worry about styles added by extra button classes, those can be removed in the view
- Try to keep the external link icons, with the right color

Also, in case multiple buttons are added to the "button" region, let's add a small space between elements. I think the best way to do this is to make `hs-button` `flex`, with `column` direction and add a small `gap` value. We can't control space inside multivalued fields (like the links in the psychology example), but that's ok, better to recommend not to add this kind of things in this region.

docroot/themes/humsci/humsci_basic/src/scss/partials/components/_vertical-button-card.scss
```diff
@@ -20,18 +20,33 @@
...
-
-    &:hover,
-    &:focus,
-    &.is-focused {
-      background-color: color('primary-dark');
-      border-color: color('primary-dark');
+  .hs-button {
+    display: flex;
+    flex-direction: column;
+    gap: hb-calculate-rems(4px);
```
cienvaras: Can we do this only in colorful? It has some undesired side effects in traditional

Image

cienvaras (changes requested):
@mariannuar Only one minor thing to review

cienvaras (approved):
@mariannuar LGTM, thanks!

@ahughes3 Ready for you

Jnwibowo7321 (approved):
LGTM!

joegl (approved):

PR #2274, opened 2026-06-10
HSD8-1689: Capture hs_training layout builder display config; add single_content_sync support and deploy hook

imonroe:
# Summary
Here we capture the front-end layout builder configs for `hs_training` content and some associated configuration updates from the `stage` server. Also adds `hs_training` to the `single_content_sync` allow list, a deploy hook to re-import the display config on environments where `config_ignore` previously blocked it, and several minor config fixes.

## Backstory
[HSD8-1689](https://stanfordits.atlassian.net/browse/HSD8-1689)

We got the content type working, but we needed the layout builder work set up by Albert.  He did so on the `stage` environment, and so we got the export.

Because `core.entity_view_display.node.hs_*` is config_ignored, the updated display config won't import via `drush config:import` on staging (the deploy hook has already run there). A new deploy hook (`hs_admin_deploy_10005`) force-imports just that config through the config factory so events and cache invalidation fire correctly.

`single_content_sync` is also config_ignored, so a database update (`su_humsci_profile_update_11002`) adds `hs_training` to the allowed node types programmatically to keep it consistent with the other content types.

## Need Review By (Date)
ASAP if we're going to get it released by June 17

## Urgency
high

## Steps to Test
1. Check tugboat to make sure the `hs_training` content looks correct on the front end with the updated layout builder display.
2. Verify `hs_training` appears in the allowed node types at `/admin/config/content/single-content-sync`.
3. Confirm `drush deploy` runs `hs_admin_deploy_10005` and `su_humsci_profile_update_11002` cleanly with no errors.

[HSD8-1689]: https://stanfordits.atlassian.net/browse/HSD8-1689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/default/core.base_field_override.node.hs_training.status.yml
```diff
@@ -1,3 +1,4 @@
+uuid: 45997620-772f-4835-9118-faa81b79baf6
```
joegl, line 1: Because this existed without a UUID on first deploy, this will most likely fail to import on staging (or cause some sort of config import warning). What it might even do is delete the active config on the site and re-import. However, since this hasn't gone out to production yet that's not a real concern since it'll be created with this UUID once it's deployed. I don't think there's anything we really want to do here then.

joegl (commented):
I'm not sure exactly how we want to handle the `config/default/core.entity_view_display.node.hs_training.default.yml` changes captured here.

Because there is a `config_ignore` rule to ignore `core.entity_view_display.node.hs_*` these changes will not get imported. We do already have a deploy hook to import this configuration on deploy, and while that will resolve correctly on production and import this newer version of the config, the configuration on staging won't re-run the deploy hook, remain stale, and not import these changes across the staging environment.

We could add a second redundant hook to force these on staging, not do anything and let staging remain stale, or we could run some sort of script or command across all sites on staging after it gets deployed.

@ahughes3 I'm curious what your thoughts are. I assume for testing you'd want the display updated across all sites.

@imonroe What do you think is the best option? I'm leaning towards a redundant deploy hook. We could also just increment the number in the existing deploy hook and move it to the bottom of the deploy hook file so it re-runs on staging. 

Link to relevant deploy hook: https://github.com/SU-HSDO/suhumsci/blob/d778e11d8efd0be30a3ef1690aae04b840bc9241/docroot/modules/humsci/hs_admin/hs_admin.deploy.php#L76

PR #2343, opened 2026-08-08
HSD8-1984: Change views always_live_preview to false

ahughes3:
# Summary
See: https://stanfordits.atlassian.net/browse/HSD8-1984

The auto preview on the view settings page can sometimes be a performance hindrance. Change the default to "disabled" in Views instead of "enabled."


joegl (approved):
Makes sense to me.

PR #2344, opened 2026-08-11
HSD8-1907: Install and initial configuration of Algolia search

imonroe:
# Summary
_[briefly summarize the changes here. TL;DR - what's this PR for?]_

## Backstory
_[help a reviewer understand *why* this PR exists]_

## Need Review By (Date)
_[When does this need to be reviewed by? '10/30', 'asap', etc.]_

## Urgency

## Steps to Test
1. _[First testing step]_
2. ...

---
  - https://app.asana.com/0/0/1215175895761125


ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

ahughes3 (approved):
connection worked for me!

joegl:
@ahughes3 @imonroe I have not provided a code review for this one yet. I will make it a priority but not sure I'll get to it before I am out on Friday.

joegl:
@imonroe Could you fill out the pull request summary?

ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

docs/Config.md
```diff
@@ -12,6 +12,21 @@ The platform uses a combination of contributed and custom modules to manage conf
...
 
+### Ignoring a Single Key of New Configuration
+
+A pattern can target one key of a configuration object using colon syntax, such as `views.view.hs_*:status`. For configuration that already exists on a site, the site's value is kept and the value from `config/default` is discarded, which is what you would expect.
+
+New configuration behaves differently. When the object does not yet exist in a site's active storage, `config_ignore` removes the ignored key from the imported data instead of preserving it. Drupal then applies its own default for that key. For a configuration entity, `status` defaults to enabled, so shipping a new entity with `status: false` and a `:status` ignore pattern creates it **enabled** on every site.
+
+Plan for this whenever you add a per-key pattern for configuration that does not exist yet:
+
+- Ship the configuration in a module's `config/install` as well as in `config/default`, so new sites create it with the intended values before the first import. The two copies must be byte identical, including the `uuid`, or the import will fail on a UUID mismatch.
```
joegl: Is this step really necessary with the deploy hook?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
+# Algolia Search
+
+HSDP provides site search through Search API backed by a database index. Algolia is a hosted search service that adds relevance tuning, synonyms, facets, promoted results, and search analytics. It is available as an opt-in backend on a per-site basis.
+
+Algolia does not replace the database index. Every site keeps its database-backed `/search` page whether or not Algolia is enabled.
```
joegl: I kind of get what this is saying but it's a bit confusing. It's saying that the Drupal default search index database exists separately from the Algolia index right?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `search_api.server.hs_algolia`, a Search API server using the `search_api_algolia` backend
+- `search_api.index.hs_algolia`, a second index that runs alongside `default_index`
+- The `hs_algolia` module, which adapts the contributed module to this platform
+
+All three ship disabled. A site starts indexing only after the H&S web team adds Algolia credentials to that site's secrets file and enables the server and index.
+
+### Why This Is a Second Index
+
+`hs_algolia` does not replace `default_index`, and the `/search` page and header search box continue to query the database. Three reasons:
```
joegl: What is "the" database here?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `default_index` is the only index that can serve private content. It carries the `node_grants` field and the `content_access` processor, which filter results by the current user's access at query time. Algolia has no equivalent.
+- The two indexes have different shapes. `default_index` stores one tokenized blob of rendered HTML with a boosted title. `hs_algolia` stores flat, facetable attributes because Algolia does its own tokenizing, stemming, and relevance ranking.
+- The Algolia front end queries Algolia directly from the browser rather than through Search API views.
+
+The cost is that a site with Algolia enabled indexes its content twice. The database index writes on save and is local; Algolia is batched through cron.
```
joegl: Is this always the case? Can we turn the default index off or could that be built in at some point?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+Every site imports the same configuration from `config/default`, so the values that must differ per site are excluded from import by `config_ignore`. See [Configuration Management](Config.md) for the general conventions.
+
+| Value | Where it lives |
+|---|---|
+| `search_api.server.hs_algolia:status` | Site database, excluded by `config_ignore` |
+| `search_api.index.hs_algolia:status` | Site database, excluded by `config_ignore` |
+| `search_api.index.hs_algolia:options.algolia_index_name` | Site database, excluded by `config_ignore` |
+| Application ID and Admin API key | Per-site `secrets.settings.php`, never in this repository |
+
+> **Important:** A per-key `config_ignore` pattern strips the key entirely when the configuration does not yet exist in a site's active storage, rather than preserving it. Because Drupal defaults a config entity's status to enabled, both entities would be created enabled on their first import. Two safeguards prevent this: the `hs_algolia` module creates both entities disabled from its `config/install` on new sites, and a deploy hook in `hs_admin` forces them back off on existing sites immediately after import. Do not remove either one.
```
joegl: If we want to allow changing only one or two values in a big form but don't want to allow other values to be changed, shouldn't we create a separate form or config page to set these values? Then we aren't ignoring a single configuration key and opening up an entire form.

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+### Per-Site Configuration
+
+Every site imports the same configuration from `config/default`, so the values that must differ per site are excluded from import by `config_ignore`. See [Configuration Management](Config.md) for the general conventions.
+
+| Value | Where it lives |
+|---|---|
+| `search_api.server.hs_algolia:status` | Site database, excluded by `config_ignore` |
+| `search_api.index.hs_algolia:status` | Site database, excluded by `config_ignore` |
+| `search_api.index.hs_algolia:options.algolia_index_name` | Site database, excluded by `config_ignore` |
+| Application ID and Admin API key | Per-site `secrets.settings.php`, never in this repository |
```
joegl: Every site will need to modify the `secrets.settings.php` to add their credentials? Why can't we use the key and encrypt module to allow these to be configurable on each individual site?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+1. Run the initial index. Content is indexed through cron in batches, so populate a new index directly rather than waiting.
+
+   ```bash
+   drush @<SITE_NAME>.<ENV> search-api:index hs_algolia
+   ```
+
+1. Confirm the records appear in the Algolia dashboard, and that no unpublished or Private Page content is among them.
+
+> **Warning:** Enable a site only after a deploy has finished. The deploy hook that forces the Algolia configuration off runs once per site, and enabling a site partway through a deploy would be undone.
```
joegl: A lot of this documentation seems to apply to the first deployment of the code and config, which will only ever happen once. Have you considered what happens when a new site is installed on the platform and how it behaves during a site install?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Local Development Setup
+
+Local credentials go in `keys/secrets.settings.php`. Confirm `keys` is listed in `.gitignore` before you put anything in it.
```
joegl: This file gets pulled from the staging server whenever the sws:keys drush command is run. It's not where local settings should go. Any local settings belong in: `docroot/sites/settings/local.settings.php`

You can even stub out what should be in a new `local.settings.php` file in: `docroot/sites/settings/default.local.settings.php`

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+- Training and Project content, which have no search indexing view display yet
+- Node grants, which are never sent to a third-party service
+
+> **Note:** Because private content is excluded, a future search across sites will not surface private content. Supporting that safely requires per-user filtered queries using signed Algolia keys, which is a separate piece of work.
+
+## Enabling Algolia for a Site
+
+You need an Algolia application with its Application ID, Admin API key, and Search-only API key before you start.
+
+1. Add the credentials to the site's secrets file on Acquia at `<AH_FILES_ROOT>/<SITE_NAME>/secrets.settings.php`. Create the file if it does not exist.
```
joegl: Why was this architecture chosen to have unique site secrets.settings.php files?

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+```php
+$settings['hs_algolia_trim_html'] = TRUE;
+```
+
+This drops the tail of long pages, so content near the bottom stops matching. Use it for local and sandbox work. On a production site, either use a plan with a larger record limit or split oversized records with the `algolia_item_splitter` processor.
+
+## Platform Customizations
+
+The contributed `search_api_algolia` module handles indexing only. It does not provide a search results page. The `hs_algolia` module adds the behavior this platform needs.
```
joegl: This seems to imply the `hs_algolia` module provides a search results page. I think the last sentence here too is unnecessary.

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+- **Deletions happen immediately.** The contributed module records deletions in a database table and expects a separate scheduled task to run a drush command that clears them. This platform does not run that task, so records for deleted content would remain in Algolia indefinitely. `hs_algolia` deletes the record during request shutdown instead.
+- **Unpublishing removes content from search.** Search API stops tracking an unpublished node but leaves its Algolia record in place until the next full reindex. `hs_algolia` clears the record as soon as a published node is unpublished.
+- **URLs use the site's canonical domain.** Cron builds URLs from the current request, which on Acquia is an internal hostname. `hs_algolia` rewrites them to the domain configured in the site's 301 redirect settings.
+- **Taxonomy values are always arrays.** A reference field holding one term arrives as a string. Sending a consistent shape keeps Algolia facets and the front end simple.
+- **Tracking fields are removed** from each record, and the title is moved to the front so records are readable in the Algolia dashboard.
```
joegl: This is all really important information. Good to have here.

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Future Work
+
+- A front-end search experience. The results page, filters, and facets query Algolia directly from the browser using the search-only key, typically through Algolia's InstantSearch library. This is new work rather than a change to the existing search view.
+- Search across multiple sites, where two or more sites index into one shared Algolia application. Records are already keyed on node UUID rather than node ID, so content from different sites cannot overwrite one another, and each record carries the site name for attribution and filtering. What remains undecided is whether participation is opt-in per site, whether a shared index gets its own Algolia application, and how a site joins an existing group later.
+- Indexing Training and Project content, which first requires a search indexing view display for each and a deploy hook to roll it out.
```
joegl: This doesn't belong in this documentation.

docs/AlgoliaSearch.md
```diff
@@ -0,0 +1,137 @@
+# Algolia Search
```
joegl: This document is really verbose and I think includes a lot of unnecessary and redundant information. It could probably be trimmed down pretty substantially. Did you run it past the DocumentationStandards.md doc here?

joegl: This documentation should also be linked to from the README.md in the documentation section.

docroot/modules/humsci/hs_algolia/src/SearchApiAlgoliaHelper.php
```diff
@@ -0,0 +1,110 @@
...
+
+    foreach (ContentEntity::getIndexesForEntity($entity) as $index) {
+      $index_name = $index->getOption('algolia_index_name');
+      $object_id_field = $index->getOption('object_id_field');
+
+      // Indexes belonging to another backend have no Algolia index name.
+      if (!$index_name || !$object_id_field || !$entity->hasField($object_id_field)) {
+        continue;
+      }
+
```
joegl: Should this check if the index is enabled or not? If an index exists but is disabled (and this module is enabled) it's going to keep operating here.

joegl: Additionally the contrib module also checks if the index is set to be read only before deleting. Should this check that too?

`if ($index->get('read_only')) {`

docroot/modules/humsci/hs_algolia/src/SearchApiAlgoliaHelper.php
```diff
@@ -0,0 +1,110 @@
...
+ * The contrib helper only records the deletion in the
+ * search_api_algolia_deleted_items table. Clearing that queue requires a
+ * separate scheduled task running the module's drush command, which this
+ * platform does not run. Records for deleted or unpublished content would
+ * therefore stay in Algolia indefinitely. Deleting during shutdown keeps the
+ * request fast while guaranteeing the record is gone.
+ *
+ * @see \Drupal\search_api_algolia\SearchApiAlgoliaHelper::scheduleForDeletion()
+ */
+class SearchApiAlgoliaHelper extends ContribSearchApiAlgoliaHelper {
```
joegl: Extending and overriding a class from a contrib module can become a maintenance nightmare sometimes. Is there no other way to achieve what we need or take advantage of the default deletion table the Algolia module uses? Couldn't we just implement our own `hook_cron()` job to process the batch table periodically?  Then we don't need any of these services overrides. We are able to set up a unique Cron job in Acquia and schedule it to run as frequently across all these sites as we want (maybe something like 30 minutes)? Then we are:

- Leveraging behavior that already exists and not extending and overriding it
- Deleting objects in a bulk, batch way using the Algolia API bulk delete API instead of single unoptimized requests
- This matches the behavior of indexing new content (which is not done on demand but a batch process)

Take a look at the custom `sws:multisite:cron:scheduler` command here. We could create one for Algolia to run the Cron job across all sites in a Scheduled Job within Acquia at a completely custom interval.

Obviously the one trade off is that instead of instant it will be at or less than the scheduled Cron time, but I think that's probably acceptable. We can check with Albert if needed.

Also, I understand this implementation may have been taken from core/Stanford Sites. With H&S we have the ability to do something more custom, still at scale, but at a much smaller scale than that platform.

docroot/modules/humsci/hs_algolia/src/SearchApiAlgoliaHelper.php
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function entityDelete(EntityInterface $entity) {
```
joegl: What happens when nodes get deleted in bulk? Will this timeout sending the individual requests and is there a way to easily add them to a batch and execute that?

docroot/modules/humsci/hs_algolia/hs_algolia.info.yml
```diff
@@ -0,0 +1,10 @@
+name: 'H&S Algolia'
+type: module
+description: 'Customizations for indexing H&S content into Algolia via Search API'
+core_version_requirement: ^10.3 || ^11
+package: 'Humanities & Sciences'
+dependencies:
+  - 'drupal:node'
+  - 'search_api:search_api'
+  - 'search_api_algolia:search_api_algolia'
```
joegl: Probably should add a dependency on domain_301_redirect here since we call those settings.

docroot/modules/humsci/hs_algolia/hs_algolia.module
```diff
@@ -0,0 +1,121 @@
+<?php
```
joegl: Whole file looks great 👍

docroot/modules/humsci/hs_admin/hs_admin.deploy.php
```diff
@@ -605,3 +605,38 @@ function hs_admin_deploy_10013(): string {
...
+ * config_ignore strips the ignored key from the imported data instead of
+ * preserving it, and Drupal's
+ * default for a config entity status is TRUE. Without this hook every existing
+ * site would enable the Algolia server and index the first time it imported
+ * this configuration. Fresh sites are unaffected because hs_algolia creates
+ * both entities from its config/install before the first import.
+ *
+ * The raw configuration is written rather than disabling the config entities,
+ * because Index::postSave() and Server::postSave() tear down tracker state and
+ * call a backend that has no credentials yet.
```
joegl: This is quite a lot of information and verbosity for a deploy command. Can we trim it at all?

docroot/modules/humsci/hs_algolia/src/SearchApiAlgoliaHelper.php
```diff
@@ -0,0 +1,110 @@
...
+ * Deletes Algolia records immediately instead of queueing them.
+ *
+ * The contrib helper only records the deletion in the
+ * search_api_algolia_deleted_items table. Clearing that queue requires a
+ * separate scheduled task running the module's drush command, which this
+ * platform does not run. Records for deleted or unpublished content would
+ * therefore stay in Algolia indefinitely. Deleting during shutdown keeps the
+ * request fast while guaranteeing the record is gone.
+ *
+ * @see \Drupal\search_api_algolia\SearchApiAlgoliaHelper::scheduleForDeletion()
```
joegl: This comment is really specific to the primary function of the file, but it's technically a helper service that extends the base helper service. This comment probably applies more inline to the actual methods doing these things.

config/default/search_api.index.hs_algolia.yml
```diff
@@ -0,0 +1,277 @@
...
+    title: false
+    alt: false
+    tags: {  }
+  language_with_fallback: {  }
+  rendered_item: {  }
+tracker_settings:
+  default:
+    indexing_order: fifo
+options:
+  cron_limit: 50
```
joegl: This seems remarkably low for the Cron batch size. I suppose if we do set up manual custom jobs to run the indexing and deletion (like described in another comment) then 50 might be fine. However, the default Cron job only triggers every 4 (maybe even 6) hours. Processing only 50 at that time interval seems a bit low.

Or maybe we set up a scheduled job to run a re-index on a site daily with `drush search-api:index hs_algolia`? Or would that be too burdensome across the platform?

config/default/search_api.index.hs_algolia.yml
```diff
@@ -0,0 +1,277 @@
...
+    type: string
+    configuration:
+      value: '[node:content-type:name]'
+  url:
+    label: 'Page URL'
+    property_path: custom_value
+    type: string
+    configuration:
+      value: '[node:url:absolute]'
+datasource_settings:
```
joegl: We may want to eventually (or now) add the `datasource_settings` to `config_ignore`. Different sites may want different content indexed, and some sites have custom content types that only exist on their site.

joegl (changes requested):
Left some comments, questions, suggestions. This is a big implementation and want to get this right. 

One additional thought too:

- On the server configuration page, do we want to hide the credential fields (even just visually)? We probably don't want anyone using or seeing these or getting confused about how the credentials are stored (this page: `/admin/config/search/search-api/server/hs_algolia/edit`). This could probably be achieved pretty easily with a form alter hook.

docs/AlgoliaSearch.md
imonroe: Yes, correct.

docs/AlgoliaSearch.md
imonroe: The database search index, as opposed to the algolia index

docs/AlgoliaSearch.md
imonroe: It is certainly possible to do so, yes.

docs/AlgoliaSearch.md
imonroe: Yes.  When a new site is provisioned, the initial state of the algolia index will be turned off.  The tricky bit is sites which already exist, where we want the module to exist, but we want it turned off.

docroot/modules/humsci/hs_algolia/src/SearchApiAlgoliaHelper.php
imonroe: So, I think the `cron` approach is a good one.  I'm going to refactor to handle the deletes that way.

config/default/search_api.index.hs_algolia.yml
imonroe: Raised to 200. Worth noting the setting does not work the way the name suggests: Search API cron indexes one batch of `cron_limit` items, then keeps taking batches until search_api.settings:cron_worker_runtime (15s) runs out, so a run already indexes a few hundred items at 50. Raising cron_limit mostly means fewer Algolia round trips per run. The lever is cron_worker_runtime, but it's global and would change default_index behavior on every site, so I left it alone. Initial population still goes through `drush search-api:index hs_algolia`

docs/AlgoliaSearch.md
imonroe: updated documentation to clarify

ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

docs/Config.md
imonroe: Yes, they cover disjoint cases. Drupal marks all existing deploy hooks as already-run when a module is first installed, so on a newly provisioned site hs_admin_deploy_10014() never fires. config/install handles new sites; the deploy hook handles existing ones.

docs/AlgoliaSearch.md
imonroe: Edited down, and linked in the readme.

ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

joegl:
@imonroe I noticed quite a few changes made here. Is this ready again for review?

docroot/modules/humsci/hs_algolia/hs_algolia.info.yml
imonroe: fixed

docs/AlgoliaSearch.md
imonroe: Refactored to use a form and to store the secrets in config-ignored keys

docs/AlgoliaSearch.md
imonroe: fixed

docs/AlgoliaSearch.md
imonroe: refactored to change the approach

docs/AlgoliaSearch.md
imonroe: removed

docroot/modules/humsci/hs_admin/hs_admin.deploy.php
imonroe: edited and rewritten

imonroe:
> @imonroe I noticed quite a few changes made here. Is this ready again for review?

Shortly.  I'm dotting my I's and crossing my T's.  I'll re-request when I'm finished.

imonroe:
Ok, I've done a substantial re-write of the whole PR, and I've incorporated most of the feedback from the first round.  I think that this is ready for a re-review as a whole.

joegl (commented):
This is looking great. I'm wondering if @ahughes3 wants to re-review on Tugboat with the changes made? A few other points:

1. Some sites use content_access on specific node types that are being indexed here to hide them from the public. As far as I understand, Algolia would still index these nodes. Is there a way we could prevent these from being indexed? Perhaps in a hook we could check for content access node grants?
2. We'll want to set up the Cron job ASAP to run these. @imonroe Could you take a look at the SWS drush commands Cron commands we have? If we write a custom one to run this Cron job across all sites it'd make it much easier to put into the Acquia scheduled jobs.

imonroe:
I added a check for anonymous access before indexing, and I added a drush command to run any cron jobs, modeled after the version in the sws drush commands.

ahughes3:

Links:
- https://pr2344-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-colorful-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://hs-traditional-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://english-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://history-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com
- https://west-ipzheuuktwdmx5qeu4wd30jilvoiekm9.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174

Visual Diffs:
- https://dashboard.tugboatqa.com/6a7b59e03457b1eeda835174/visualdiffs

docroot/modules/humsci/hs_admin/hs_admin.deploy.php
```diff
@@ -605,3 +605,31 @@ function hs_admin_deploy_10013(): string {
...
+  foreach (['search_api.server.hs_algolia', 'search_api.index.hs_algolia'] as $name) {
+    // Write raw config rather than saving the entity. Index::postSave() and
+    // Server::postSave() would call a backend that has no credentials yet.
+    $config = \Drupal::configFactory()->getEditable($name);
+    if ($config->isNew() || !$config->get('status')) {
+      continue;
+    }
+
+    $config->set('status', FALSE)->save();
+    $disabled[] = $name;
```
joegl:
```suggestion
    $config->set('status', FALSE)->save();
    $disabled[] = $name;
    
    // Disable tracking for the index.
    if ($name === 'search_api.index.hs_algolia') {
      $index = \Drupal::entityTypeManager()
        ->getStorage('search_api_index')
        ->loadUnchanged('hs_algolia');
      \Drupal::service('search_api.index_task_manager')->stopTracking($index);
    }
```

Enabling an index turns tracking on for index (which happens here during config import). The typical disable process (`Index::postSave()`) then turns tracking off, but we're not going through that process. Recommend turn tracking off here for disabled indexes.

docroot/modules/humsci/hs_algolia/hs_algolia.module
```diff
@@ -0,0 +1,263 @@
...
+
+      // Cron builds URLs from the current request, which on Acquia is the
+      // internal hostname rather than the site's public domain.
+      if ($canonical_host && $current_host && $canonical_host !== $current_host && is_string($value)) {
+        $value = str_replace($current_host, $canonical_host, $value);
+      }
+    }
+
+    hs_algolia_trim_record($object);
+  }
```
joegl:
```suggestion
  }

  // A record being sent to Algolia is live, so cancel any deletion queued for
  // it. Otherwise a later cron run removes it again, for example when a node
  // is unpublished and then republished before the queue drains.
  $object_ids = array_filter(array_column($objects, 'objectID'));
  $index_name = $index->getOption('algolia_index_name');
  if ($object_ids && $index_name) {
    \Drupal::database()->delete('search_api_algolia_deleted_items')
      ->condition('index_id', $index_name)
      ->condition('object_id', $object_ids, 'IN')
      ->execute();
  }
```

This is a theoretical issue and fix, suggested by Claude. Essentially, a node can be queued for deletion when being unpublished, however if it's re-published before the deletion Cron runs, then the deletion Cron will unindex the item and it will disappear until the node is edited again and added back to process.

> If the node is republished and re-indexed before hs_algolia_cron drains the queue (cron order, a failed drain, or Index now), the drain deletes the live record. Search API still counts it as indexed, so it stays missing until the next edit. Suggest cancelling queued deletions for any record being sent.

I think it's worth looking into.

docroot/modules/humsci/hs_algolia/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,175 @@
...
+      '#description' => $this->t('Index published content into Algolia. Disabling removes every record from the Algolia index.'),
+      '#default_value' => $index?->status() ?? FALSE,
+    ];
+
+    $form['application_id'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Application ID'),
+      '#description' => $this->t('From the API Keys page of the Algolia application.'),
+      '#default_value' => $config->get('application_id'),
+      '#maxlength' => 128,
```
joegl:
```suggestion
      '#maxlength' => 128,
      '#disabled' => $index?->status() ?? FALSE,
```

Changing the Application ID or index name while Algolia is enabled points the index at a new, empty Algolia index, and nothing triggers a reindex. The old index also keeps its public records. And because settings save before the disable, changing the name and unchecking Enable in one save clears the new index, not the old one. Suggest disabling these two fields while enabled, so a change has to go through disable (clears) and re-enable (re-tracks).

docroot/modules/humsci/hs_algolia/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,175 @@
...
+      '#empty_option' => $this->t('- Select -'),
+      '#default_value' => $config->get('api_key'),
+    ];
+
+    $form['index_name'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Algolia index name'),
+      '#description' => $this->t('Name of the index inside the Algolia application. It is created on first use. Use the site name, for example <code>archaeology</code>.'),
+      '#default_value' => $config->get('index_name'),
+      '#maxlength' => 128,
```
joegl:
```suggestion
      '#maxlength' => 128,
      '#disabled' => $index?->status() ?? FALSE,
```

See above comment.

docroot/modules/humsci/hs_algolia/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,175 @@
...
+      if (!$index->status()) {
+        $index->enable()->save();
+      }
+      $this->messenger()->addStatus($this->t('Algolia search is enabled. Content is indexed on cron. To index everything now, use <a href=":url">Index now</a> on the index page.', [
+        ':url' => Url::fromRoute('entity.search_api_index.canonical', ['search_api_index' => self::INDEX_ID])->toString(),
+      ]));
+    }
+    else {
+      if ($index->status()) {
+        $index->disable()->save();
```
joegl:
```suggestion
        // Disabling clears the whole Algolia index, so any queued deletions
        // are moot. Drop them so cron does not carry them indefinitely.
        $this->database->delete('search_api_algolia_deleted_items')->execute();
```

The code above requires injecting the database service.

joegl (changes requested):
@imonroe This looks good to me. A couple things flagged by Claude I want your thoughts on. Otherwise let's get this in.

